PR #5, opened 2018-10-11
Personal/macarl/gatedbuild

matthewfcarlson:
Added configuration validation as well as added skipping tests as needed


UefiBuild/MuBuild/ConfigValidator.py
```diff
@@ -0,0 +1,307 @@
+## @file ConfigValidator.py
+# This module contains support for Outputting Junit xml.
```
spbrogan: comment incorrect

UefiBuild/MuBuild/MuBuild.py
```diff
@@ -70,9 +72,63 @@ def get_mu_config():
     parser.add_argument (
     '-p', '--pkg','--pkg-dir', dest = 'pkg', required = False, type=str,help = 'The package or folder you want to test/compile relative to the Mu Config'
     )
+    parser.add_argument (
+    '-check-config', '--cc', dest = 'check_config', required = False, action="store_true",help = 'The package or folder you want to test/compile relative to the Mu Config'
```
spbrogan: incorrect

UefiBuild/PluginManager.py
```diff
@@ -215,6 +215,9 @@ def GetTestName(self, packagename, environment):
     def IsTargetDependent(self):
         return False
 
+    def ValidateConfig(self,config):
```
spbrogan: what does a plugin do if error.

UefiBuild/MuBuild/MuBuild.py
```diff
@@ -191,17 +252,21 @@ def get_mu_config():
...
-        for Descriptor in pluginManager.GetPluginsOfClass(PluginManager.IMuBuildPlugin):
+        for Descriptor in pluginList:
             #Get our targets
             targets = ["DEBUG"]
             if Descriptor.Obj.IsTargetDependent() and "Targets" in mu_config:
                 targets = mu_config["Targets"]
 
-
+            #merge the repo level and package level
+            pkg_configuration = merge_config(mu_config,pkg_config,Descriptor.descriptor)
```
spbrogan: shouldn't this happen before check_package_config?  
Does it really need to happen on this inner loop or could it happen 1 loop up?

spbrogan (changes requested):

UefiBuild/PluginManager.py
matthewfcarlson: Generally it's expected that a plugin will throw an exception if it encounters an error just like the validators

matthewfcarlson:
Fixed all the issues except the validate config

UefiBuild/PluginManager.py
matthewfcarlson: Another solution would be to return false or some message about what went wrong but then it's harder to see where the exception is coming from

matthewfcarlson:
I've removed the case insensitive dictionary and now I've fixed a few other things I noticed

spbrogan (changes requested):
looks like you pulled in a submodule.  can you remove that

matthewfcarlson:
Removed the submodule

spbrogan:
another one in common

spbrogan (approved):

PR #10, opened 2018-12-13
Upstream resolved issue in DxeServicesLib for FvHandle being NULL

spbrogan:
Undo Project Mu change since upstream is now fixed.


corthon (approved):
Looks good.

PR #13, opened 2018-12-19
ConSplitter: do not perform TextOut SetMode during device add flow if the required mode is already set

dmitryantipov:


MdeModulePkg/Universal/Console/ConSplitterDxe/ConSplitter.c
```diff
@@ -30,6 +30,9 @@ WITHOUT WARRANTIES OR REPRESENTATIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED.
 
 #include "ConSplitter.h"
 
+// MU_CHANGE: do not perform set mode during device add flow if the required mode is already set
```
corthon, line 33: Is there a reason this isn't generalized to only *ever* set the mode if it doesn't already match?

dmitryantipov: This function gets called in 2 cases - internal call when adding a new device (which flow I'm modifying) and protocol call that lands here when the user of the protocol explicitly wants to SetMode. According to UEFI spec, SetMode is supposed to clear the screen if the mode is already set, so we're keeping that behavior. However, the spec doesn't specify the ConSplitter behavior on addition of new ConOut devices - and I can't think of a reason we'd want to clear the screens on all consoles when adding a new ConOut device to the ConSplitter.

MdeModulePkg/Universal/Console/ConSplitterDxe/ConSplitter.c
```diff
@@ -4873,10 +4878,13 @@ ConSplitterTextOutSetMode (
...
   for (Index = 0, ReturnStatus = EFI_SUCCESS; Index < Private->CurrentNumberOfConsoles; Index++) {
-    Status = Private->TextOutList[Index].TextOut->SetMode (
+    // MU_CHANGE: only skip if the flag is true and current mode matches the required one
+    if ((mPerformingDeviceAdd == FALSE) || (TextOutModeMap[Index] != Private->TextOutList[Index].TextOut->Mode->Mode)) {
+      Status = Private->TextOutList[Index].TextOut->SetMode (
                                                     Private->TextOutList[Index].TextOut,
                                                     TextOutModeMap[Index]
                                                     );
+    }
     if (EFI_ERROR (Status)) {
```
corthon, line 4889: Thanks to the "if" this has the potential to no longer be initialized.

corthon (changes requested):

MdeModulePkg/Universal/Console/ConSplitterDxe/ConSplitter.c
dmitryantipov: Addressed now. Thanks.

dmitryantipov:
Let's not complete this pull request yet as I'm looking for more feedback.

corthon (approved):

PR #14, opened 2019-01-16
Fix Ip6Dxe not cleaning up installed protocol

chris-oo:
Ip6Dxe in Ip6DriverBinding start can fail after installing the protocol, which leaves a dangling pointer as the installed protocol interface is freed during the cleanup path. Fix the use-after-free by removing the installed protocol on failure paths.

### DIFF
--- a/NetworkPkg/Ip6Dxe/Ip6Driver.c
+++ b/NetworkPkg/Ip6Dxe/Ip6Driver.c
@@ -564,7 +564,7 @@ Ip6DriverBindingStart (
                   NULL
                   );
   if (EFI_ERROR (Status)) {
-    goto ON_ERROR;
+    goto FREE_SERVICE; // MU_CHANGE
   }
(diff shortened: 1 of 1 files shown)


mikeytdisco (approved):

corthon (approved):

NetworkPkg/Ip6Dxe/Ip6Driver.c
```diff
@@ -564,7 +564,7 @@ Ip6DriverBindingStart (
                   NULL
                   );
   if (EFI_ERROR (Status)) {
-    goto ON_ERROR;
+    goto FREE_SERVICE;
```
corthon: Missing a MU_CHANGE comment.

corthon (changes requested):

corthon (approved):

PR #15, opened 2019-01-30
Fix typo

changeworld:
Ture -> True


corthon:
Thanks for submitting!

We're currently in the process of migrating to a new branch and have to be selective about which PRs we migrate. This is a minor change to the upstream project (TianoCore) and will likely be fixed in their source. Sorry we we're able to take it this time, but please don't be discouraged. I'm sure there's still plenty of room for improvement in the new branch.

PR #20, opened 2019-02-26
Change TPL for SnpDxe's ExitBootServices handler from NOTIFY to CALLBACK

dmitryantipov:


corthon:
Can you withdraw and rebase this against the new release/201903 branch?

PR #30, opened 2019-06-06
Standardize behavior when VariablePolicy is disabled.

corthon:
Remove the unnecessary error when registering a VariablePolicy after policy enforcement has been disabled.

Add tests for this and other errors when booting with VariablePolicy disabled.


kuqin12 (approved):

matthewfcarlson (approved):

matthewfcarlson:
Does this mean there an expectation on those who create policies to make sure that enforcement is on when creating policies? It seemed like a previous approach was to throw an error when enforcement wasn't on

spbrogan (approved):

corthon:
@matthewfcarlson,
If they truly need to know whether the policy is being enforced, yes, though the platform can always disable the policy engine after the check. I think the practice we landed on is that policy creators should only be alerted when something is wrong with their specific policy and leave the engine decisions up to the platform.

PR #32, opened 2019-07-25
Clean up and align NVMe driver with tianocore as much as possible.

spbrogan:


spbrogan:
@corthon  - what do you think about removing all the debugging comments.  I don't think they are worth carrying mu_changes

MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressHci.c
```diff
@@ -1126,9 +1126,7 @@ NvmeUnregisterShutdownNotification (
 {
   EFI_STATUS                      Status;
   EFI_RESET_NOTIFICATION_PROTOCOL *ResetNotify;
-  //MS_CHANGE - BEGIN
-  ReportStatusCode((EFI_ERROR_MAJOR | EFI_ERROR_CODE), (EFI_IO_BUS_SCSI | EFI_IOB_EC_INTERFACE_ERROR));
```
matthewfcarlson, line 1130: Would these report status code removals be good candidates for instead pivoting to emitting telemetry here?

matthewfcarlson (approved):

matthewfcarlson:
I agree- I think this debugging was probably helpful when developing/first implementing IoMMU, but not worth the MU_CHANGE and accompanying commit. When we need to debug issues, we'll probably add our own print statements that relevant to the issue at hand.

MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressHci.c
spbrogan: no.  telemetry is not in basecore yet

PR #37, opened 2019-08-22
Bug: Incorrect insertion into NVMe IO Queue based on hardcode size

spbrogan:
The size of the IO queues is correctly determined by getting the min size of the SW driver and hardware. 
The logic used to insert and roll over the queues does not use this value but instead is hard-coded to the value of the SW driver.  This causes a hang on devices which have smaller queue size than driver.


spbrogan:
Tested on a physical device and confirmed this resolves the hang.

spbrogan:
Reported to Tianocore as BZ 2118 https://bugzilla.tianocore.org/show_bug.cgi?id=2118

PR #40, opened 2019-11-20
TCBZ1969: Point ReportStatusCode router data to new buffer when reallocation occurs

kuqin12:
When buffer for RSC router is reallocated due to 7/8 of buffer is used, it is needed to update RscData buffer to point to new buffer. Otherwise this might cause random memory corruption issue.


mikeytdisco (approved):

spbrogan:
Related to Tianocore Bugzilla 1969
https://bugzilla.tianocore.org/show_bug.cgi?id=1969

spbrogan (approved):

PR #49, opened 2020-02-11
Add start and end delimiter for included ASL file

ghost:
The banner with the ASL file name at the start and end of the included ASL file helps to find out the file causing the error/warning when compiling the DSDT.iiii or SSDT.iiii

### DIFF
--- a/BaseTools/Source/Python/Trim/Trim.py
+++ b/BaseTools/Source/Python/Trim/Trim.py
@@ -247,7 +247,11 @@ def TrimPreprocessedVfr(Source, Target):
 ## Create a banner for the included ASL file
 #
 # Create a banner to indicate the start and
-# end of the included ASL file
+# end of the included ASL file. Banner looks like:-
+# 
+#   /*************************************
+#   *               @param               *
+#   *************************************/
 #
 # @param Pathname       File pathname to be included in the banner
 #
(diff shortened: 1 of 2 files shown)


BaseTools/Source/Python/Trim/Trim.py
```diff
@@ -244,6 +244,21 @@ def TrimPreprocessedVfr(Source, Target):
     except:
         EdkLogger.error("Trim", FILE_OPEN_FAILURE, ExtraData=Target)
 
+## Create a banner for the included ASL file
+#
+# Create a banner to indicate the start and
+# end of the included ASL file
+#
```
corthon, line 255: Can you add a note to this comment about what the final banner will look like (based on sample params)?

BaseTools/Source/Python/Trim/Trim.py
```diff
@@ -244,6 +244,21 @@ def TrimPreprocessedVfr(Source, Target):
     except:
         EdkLogger.error("Trim", FILE_OPEN_FAILURE, ExtraData=Target)
```
corthon, line 246: Needs "MU_CHANGE" tags.

corthon (changes requested):

corthon (approved):

PR #51, opened 2020-03-10
REBASE: Better fix for VarPol UnitTest

corthon:
Replace the superfluous pointer casts with the correct function
prototype. Needed to add EFIAPI.


spbrogan (approved):

spbrogan:
i don't think it was "superfluous".  Your function prototype didn't use the right calling convention so you would have issues with gcc and clang.  The cast probably made it worse in that you forced the compiler to ignore the incompatibility.

PR #54, opened 2020-03-23
[TCBZ2616] Update to latest extension version series

spbrogan:


spbrogan:
Link to bugzilla
https://bugzilla.tianocore.org/show_bug.cgi?id=2616

pip-requirements.txt
```diff
@@ -9,8 +9,10 @@
 # https://pypi.org/project/pip/
 # https://pip.pypa.io/en/stable/user_guide/#requirements-files
 # https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format
-#
+# https://www.python.org/dev/peps/pep-0440/#version-specifiers
 ##
 
 edk2-pytool-library==0.10.*
```
corthon, line 15: Do we want to do the same thing on Library while we're at it, to make sure that Library isn't too far off of the new Extensions base version?

corthon (changes requested):
Needs MU_CHANGE

pip-requirements.txt
spbrogan: I think i am ok leaving it free floating within the 0.10.x series.  10.x is our latest version  All docs say you should run pip install --upgrade and thus it should get upgraded.

PR #57, opened 2020-03-23
Revert "Merged PR 2171: [BGZ2586] Fixed PcdValueInit generation on linux

spbrogan:
This reverts commit cb889a4f2239c97f6858d2f4a6a7510737021a95.


spbrogan:
Don't complete this.  I think we should not revert this change.

corthon:
@spbrogan, status?

spbrogan:
I think we need tianocore to resolve.  My reason for holding was that in the edk2 they asked us to confirm if the problem still occurs after the "other" fix they made.  Regardless they are making different makefiles on linux and windows and it seems odd that this isn't needed on Windows but is on linux.  I think it can be removed on linux as well.

PR #58, opened 2020-03-23
[TCBZ2624] Don't require meta files for custom source types

corthon:
A change in GenMake.py went from this:
```make
$(FFS_OUTPUT_DIR)\$(MODULE_GUID).map :
	if exist $(OUTPUT_DIR)\<some_file>.map $(CP) $(OUTPUT_DIR)\<some_file>.map $(FFS_OUTPUT_DIR)\$(MODULE_GUID).map
```

to this:

```make
$(FFS_OUTPUT_DIR)\$(MODULE_GUID).map : $(OUTPUT_DIR)\<some_file>.map
	if exist $(OUTPUT_DIR)\<some_file>.map $(CP) $(OUTPUT_DIR)\<some_file>.map $(FFS_OUTPUT_DIR)\$(MODULE_GUID).map
```

For custom source files that do not generate maps, the change will fail. The 'if' statement on the second line should be sufficient to cover the "missing file" case without breaking the build.

### DIFF
--- a/BaseTools/Source/Python/AutoGen/GenMake.py
+++ b/BaseTools/Source/Python/AutoGen/GenMake.py
@@ -731,7 +731,7 @@ cleanlib:
                     if Dst not in self.ResultFileList:
                         self.ResultFileList.append(Dst)
                     if '%s :' %(Dst) not in self.BuildTargetList:
-                        # MU_CHANGE - Don't require a .map file for custom source types.
+                        # MU_CHANGE - [TCBZ2624] Don't require a .map file for custom source types.
                         # self.BuildTargetList.append("%s : %s" %(Dst,Src))
                         self.BuildTargetList.append("%s :" %(Dst))
                         self.BuildTargetList.append('\t' + self._CP_TEMPLATE_[self._Platform] %{'Src': Src, 'Dst': Dst})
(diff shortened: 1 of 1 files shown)


corthon:
@matthewfcarlson, FYI.

spbrogan:
we probably need edk2 basetools reviewers to look.  I have no idea if this will break some other scenario.

corthon:
For what it's worth, this is reverting to the way it originally was. This is actually a one-line revert of a massive commit that was somewhat unrelated to this.

corthon:
Actually, I went looking for the commit and found a totally different commit (what is up with my gitk right now?).

The real commit is a good one, so this may be bad.
The real commit was resolving the incremental build issue.
https://github.com/tianocore/edk2/commit/b34ed98694f053bb70d8dbe11240accf06ac23ce

matthewfcarlson:
I agree with Sean that we should get some tianocore maintainers to take a look but I'm impressed that you found the bug :)

matthewfcarlson (approved):

corthon:
So, we're pretty confident that this would cause incremental builds to break again in some cases, making builds take longer.

I'll close this PR while I look for another option. If we need this fix ASAP, we can reopen this PR and get it through as a temp fix.

corthon:
Reverts commit b34ed98694f053bb70d8dbe11240accf06ac23ce

corthon:
Will make sure that's reflected in the final commit message.

spbrogan (approved):

PR #59, opened 2020-05-09
Added support for Fmp depex for capsule generating routine

kuqin12:
Added support for Fmp depex for capsule generating routine based on existed CapsuleDependencyClass from EDKII


spbrogan:
close this as it was for review with partners.  Will integrate from upstream on next integration.  This will probably still be useful during that time as i think there is a mu change in here.  

Thank you.

PR #61, opened 2020-05-22
Remove Mu Override for Legacy8259Protocol.

spbrogan:
Trend is to allow platform and silicon layers to maintain this if needed for compat.  PcAt and the Core do not need this anymore.


spbrogan:
Also note as we try to turn on more Core CI features across the tree there are conflicts between this override and other silicon and platform packages which define this protocol as well.

PR #65, opened 2020-07-28
Update VariablePolicy to have the latest fixes and tests from the EDK2 RFC

corthon:
Update the functionality of the '#' wildcard and add testing for it.
Update the unit test format.

### DIFF
--- a/MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyLib.inf
+++ b/MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyLib.inf
@@ -36,3 +36,8 @@
   BaseMemoryLib
   MemoryAllocationLib
   SafeIntLib
+
+
+[BuildOptions]
+  MSFT:NOOPT_*_*_CC_FLAGS   = -DINTERNAL_UNIT_TEST
+  GCC:NOOPT_*_*_CC_FLAGS    = -DINTERNAL_UNIT_TEST
--- a/MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyUnitTest/UefiVariablePolicyUnitTest.c
+++ b/MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyUnitTest/UefiVariablePolicyUnitTest.c
@@ -25,7 +25,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Guid/VariableFormat.h>
 
 #include <Protocol/VariablePolicy.h>
-#include <Library/VariablePolicyLib.h>
+#include <Library/UefiVariablePolicyLib.h>
 
 #ifndef INTERNAL_UNIT_TEST
 #error Make sure to build thie with INTERNAL_UNIT_TEST enabled! Otherwise, some important tests may be skipped!
--- a/MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyUnitTest/UefiVariablePolicyUnitTest.c
+++ b/MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyUnitTest/UefiVariablePolicyUnitTest.c
@@ -385,14 +385,20 @@ WildcardPoliciesShouldMatchDigits (
   CHAR16        *CheckVar1Name = L"Wildcard1VarName12";
   CHAR16        *CheckVar2Name = L"Wildcard2VarName34";
   CHAR16        *CheckVarBName = L"WildcardBVarName56";
+  CHAR16        *CheckVarFName = L"WildcardFVarName0A";
+  CHAR16        *CheckVarZName = L"WildcardZVarName56";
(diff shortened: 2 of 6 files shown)


MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyLib.inf
```diff
@@ -36,3 +36,8 @@
   BaseMemoryLib
   MemoryAllocationLib
   SafeIntLib
+
+
+[BuildOptions]
+  MSFT:NOOPT_*_*_CC_FLAGS   = -DINTERNAL_UNIT_TEST
```
kuqin12, line 42: Should these test flags really be in this library?

corthon: It's only when building in NOOPT.
This is the best way I've found to do certain internal tests so far. I *do* think it's a place for improvement.

kuqin12 (approved):

MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyUnitTest/UefiVariablePolicyUnitTest.c
```diff
@@ -345,25 +385,34 @@ WildcardPoliciesShouldMatchDigits (
   CHAR16        *CheckVar1Name = L"Wildcard1VarName12";
   CHAR16        *CheckVar2Name = L"Wildcard2VarName34";
   CHAR16        *CheckVarBName = L"WildcardBVarName56";
+  CHAR16        *CheckVarFName = L"WildcardFVarName0A";
+  CHAR16        *CheckVarZName = L"WildcardZVarName56";
```
spbrogan: aren't there three duplicates with Name56?  

Should you include a few more like a, z, and Z.  Just to make sure you exercise all the conditionals.

corthon: Sure thing.

corthon: Though, I'll have to remember not to drop the new test cases if/when this lands in EDK2 (since the patches don't currently have them (or the docs, for that matter)).

MdeModulePkg/Universal/Variable/UefiVariablePolicy/ReadMe.md
```diff
@@ -0,0 +1,354 @@
...
+At power-on, the Variable Policy Engine is:
+
+-   Enabled -- present policy entries are evaluated on variable access
+    calls.
+-   Unlocked -- new policy entries can be registered.
+
+Policy is expected to be clear on power-on. Policy is volatile and not
+preserved across system reset.
+
+## DXE Protocol Functions
```
spbrogan: might be nice to call this DXE Protocol
Then have the structure and guid and global name

Then define the functions.  This makes skimming easier/faster.

MdeModulePkg/Universal/Variable/UefiVariablePolicy/ReadMe.md
```diff
@@ -0,0 +1,354 @@
...
+  UINT32    Version;
+  UINT16    Size;
+  UINT16    OffsetToName;
+  EFI_GUID  Namespace;
+  UINT32    MinSize;
+  UINT32    MaxSize;
+  UINT32    AttributesMustHave;
+  UINT32    AttributesCantHave;
+  UINT8     LockPolicyType;
+  UINT8     Padding[3];
```
spbrogan: More common to use Reserved rather than Padding.  
If the impact is minor i would tweak it...but if it causes more than structure def change probably not worth it.

MdeModulePkg/Universal/Variable/UefiVariablePolicy/ReadMe.md
corthon: Fine with that.

MdeModulePkg/Universal/Variable/UefiVariablePolicy/ReadMe.md
```diff
@@ -0,0 +1,354 @@
+---
+title:      UEFI Variable Policy Whitepaper
+version:    1.0 RFC
+copyright:  Copyright (c) Microsoft Corporation.
+---
+
+# UEFI Variable Policy
```
spbrogan, line 7: can you add a section on testing and provide brief description of the two tests.  Host and target.

MdeModulePkg/Library/UefiVariablePolicyLib/UefiVariablePolicyLib.inf
spbrogan: wouldn't a PCD be better?

spbrogan (approved):

PR #66, opened 2020-07-31
Code-First -- Add new unaccepted memory type describing unaccepted memory.

chris-oo:
This PR adds a new EFI_MEMORY_TYPE, EfiUnacceptedMemoryType that is used to describe unaccepted memory which is used by new virtual machine architectures like AMD SEV-SNP.

The most up to date version of the ECR can be found in the PR.


spbrogan:
> Physical address ranges not described by the memory map must be treated as unaccepted.

I think this is similar to the "Open" question but maybe this comment can be scoped.  Something like: *For platforms that support unaccepted memory any physical address ranges not described by the memory map must be treated as unaccepted.*

chris-oo:
> 
> 
> > Physical address ranges not described by the memory map must be treated as unaccepted.
> 
> I think this is similar to the "Open" question but maybe this comment can be scoped. Something like: _For platforms that support unaccepted memory any physical address ranges not described by the memory map must be treated as unaccepted._

That makes sense, I'll edit it.

spbrogan:
This doc needs to contain an license statement.
SPDX-License-Identifier: CC-BY-4.0

chris-oo:
> 
> 
> This doc needs to contain an license statement.
> SPDX-License-Identifier: CC-BY-4.0

Added.

ajfish:
I think it may make sense to add a cross reference to the specific technology. Maybe a glossary entry in Appendix R for Unaccepted Memory that describes the AMD implementation as an example?

jyao1:
Intel Trusted Domain Execution (TDX) (https://software.intel.com/content/www/us/en/develop/articles/intel-trust-domain-extensions.html) also need this feature. We will also use EfiUnacceptedMemoryType to pass the information in the guest firmware.

Sponsor: Jiewen Yao

I am a little concerned about the sentence: "For platforms that support unaccepted memory any physical address ranges not described by the memory map must be treated as unaccepted.".

UEFI memory map does not report MMIO. It will bring confusing that unreported memory is unaccepted or MMIO.

I recommend "All unaccepted memory must be reported explicitly. The unreported physical address range is treated as not-present physical address or MMIO."

jyao1:
I checked the patch. It seems we also need add HOB attribute - EFI_RESOURCE_ATTRIBUTE_UNACCEPTED.

It can tell the DXE core the initial GCD memory type - EfiGcdMemoryTypeUnaccepted.

chris-oo:
@ajfish Would linking to an AMD doc be sufficient in the Appendix, or would you like to see a brief writeup of how unaccepted memory works for SEV-SNP along with links to the appropriate documents?

@jyao1 I can add you to the sponsor list as well. And you're correct that we should modify the language to include MMIO, even though that's not described by the memory map. I'll have a follow up change to the ECR to add something along the lines of what you've proposed, thanks! I'll work with @spbrogan to have the right ECR for the PI spec change as well since we do need the new HOB attribute. 

Just as a further FYI, I've moved the ECR markdown to under source control to allow easier commenting and history changes in the PR. For future comments, it's probably easier to add specific comments to sections of the ECR that we're discussing.

ajfish:
> On Aug 18, 2020, at 3:00 PM, Chris Oo <notifications@github.com> wrote:
> 
> 
> @ajfish <https://github.com/ajfish> Would linking to an AMD doc be sufficient in the Appendix, or would you like to see a brief writeup of how unaccepted memory works for SEV-SNP along with links to the appropriate documents?
> 
> 

Seems like a Glossary entry would be fine and that s one or 2 sentences and a link. 

Thanks,

Andrew Fish

jyao1:
> I'll work with @spbrogan to have the right ECR for the PI spec change as well since we do need the new HOB attribute.

Thank you. I believe we need both new HOB attribute (EFI_RESOURCE_ATTRIBUTE_UNACCEPTED) and new GCD type (EfiGcdMemoryTypeUnaccepted) in PI spec.

jyao1:
In TDX, we have four different memory types:
1) Accepted Private memory (without S-bit in GPA) - The memory encrypted in TD, invisible to VMM.
2) Accepted Public memory (with S-bit in GPA) - The memory unencrypted in TD, visible to VMM for communication.
3) Unaccepted memory - The memory need to be accepted by TD, before use.
4) emulated MMIO - The MMIO need to be accessed by special TD-VMM-call, from TD.

We need a way to let TD guest OS known those type. The current proposal is:
1) it is reported as normal UEFI memory type (BootService, conventional) with EFI_MEMORY_CPU_CRYPTO attribute. This should be default memory we choose for TD.
2) it is reported as normal UEFI memory type (BootService) without EFI_MEMORY_CPU_CRYPTO attribute. A special TDX IOMMU driver does the conversion from private to shared, and from shared to private.
3) it is reported as the new unaccepted memory type with EFI_MEMORY_CPU_CRYPTO attribute.
4) it is reported in ASL code, instead of UEFI memory type, because we don't need RUNTIME attribute.

chris-oo:
> 
> 
> In TDX, we have four different memory types:
> 
>     1. Accepted Private memory (without S-bit in GPA) - The memory encrypted in TD, invisible to VMM.
> 
>     2. Accepted Public memory (with S-bit in GPA) - The memory unencrypted in TD, visible to VMM for communication.
> 
>     3. Unaccepted memory - The memory need to be accepted by TD, before use.
> 
>     4. emulated MMIO - The MMIO need to be accessed by special TD-VMM-call, from TD.
> 
> 
> We need a way to let TD guest OS known those type. The current proposal is:
> 
>     1. it is reported as normal UEFI memory type (BootService, conventional) with EFI_MEMORY_CPU_CRYPTO attribute. This should be default memory we choose for TD.
> 
>     2. it is reported as normal UEFI memory type (BootService) without EFI_MEMORY_CPU_CRYPTO attribute. A special TDX IOMMU driver does the conversion from private to shared, and from shared to private.
> 
>     3. it is reported as the new unaccepted memory type with EFI_MEMORY_CPU_CRYPTO attribute.
> 
>     4. it is reported in ASL code, instead of UEFI memory type, because we don't need RUNTIME attribute.

@jyao1  I think in this case we should report a new memory attribute that represents if a region of memory is shared or not. The existing ` EFI_MEMORY_CPU_CRYPTO` seems to me to express a different semantic than what we really want. What do you think about adding a new memory attribute, something like `EFI_MEMORY_SHARED_VISIBILITY` that we could use to note ranges in the EFI memory map that are shared. This also allows different isolation architectures to report shared visibility on any type of memory, like conventional, unaccepted, etc. 

We would also need to clarify that all ranges otherwise are assumed to be private.

Unaccepted Memory UEFI ECR.md
```diff
@@ -0,0 +1,158 @@
...
+| EFI_INVALID_PARAMETER  | MemoryType is EfiPersistentMemory or EfiUnacceptedMemory. ***Text Added*** |
+| EFI_INVALID_PARAMETER  | Memory is NULL.                                                            |
+| EFI_NOT_FOUND          | The requested pages could not be found.                                    |
+
+#### Appendix R - Glossary<br>
+...<br>
+***Text Added***<br>
+**Unaccepted Memory** <br>
+
+Some Virtual Machine platforms, such as AMD SEV-SNP, introduce the concept of memory acceptance, requiring memory to be accepted before it can be used by the guest. This protects against a class of attacks by the virtual machine platform.<br>
```
chris-oo: @ajfish Added this little glossary blurb.

Unaccepted Memory UEFI ECR.md
```diff
@@ -0,0 +1,158 @@
...
+| EfiReservedMemoryType   | Not usable.                                                                                                                                                                                                                                                          |
+| ...                     | ...                                                                                                                                                                                                                                                                  |
+| EfiUnacceptedMemoryType ***Text Added*** | A memory region that represents unaccepted memory, that must be accepted by the guest before it can be used. Unless otherwise noted, all other EFI memory types are accepted. For platforms that support unaccepted memory, all unaccepted valid memory will be reported as unaccepted in the memory map. Unreported physical address ranges must be treated as not-present memory. ***Text Added*** |
```
chris-oo: @jyao1 I've changed the description to describe memory that the guest should expect is usable, but unaccepted, and added a final sentence about not reported ranges. I don't think we want to require platforms to have the concept MMIO, but describing it as "not-present memory" allows the platform to do whatever it wants.

jyao1: OK. Looks good.

May I know your plan for new HOB attribute (EFI_RESOURCE_ATTRIBUTE_UNACCEPTED) and new GCD type (EfiGcdMemoryTypeUnaccepted)?

Do you need any help on that? If so, I can write some initial content then we can collaborate to improve/align the word.

jyao1:
I think in this case we should report a new memory attribute that represents if a region of memory is shared or not. The existing EFI_MEMORY_CPU_CRYPTO seems to me to express a different semantic than what we really want. What do you think about adding a new memory attribute, something like EFI_MEMORY_SHARED_VISIBILITY that we could use to note ranges in the EFI memory map that are shared. This also allows different isolation architectures to report shared visibility on any type of memory, like conventional, unaccepted, etc.

@chris-oo I am OK to add new attribute. Would you please clarify the current usage of EFI_MEMORY_CPU_CRYPTO in guest? I am not aware of that.

chris-oo:
> I think in this case we should report a new memory attribute that represents if a region of memory is shared or not. The existing EFI_MEMORY_CPU_CRYPTO seems to me to express a different semantic than what we really want. What do you think about adding a new memory attribute, something like EFI_MEMORY_SHARED_VISIBILITY that we could use to note ranges in the EFI memory map that are shared. This also allows different isolation architectures to report shared visibility on any type of memory, like conventional, unaccepted, etc.
> 
> @chris-oo I am OK to add new attribute. Would you please clarify the current usage of EFI_MEMORY_CPU_CRYPTO in guest? I am not aware of that.

From reading the specification the verbage around `EFI_MEMORY_CPU_CRYPTO` describes more of the capability of a memory region to be protected, not the current state of if it's protected or not. 

However, after getting some feedback internally, what's the reason for UEFI not being able to revoke visibility of all memory ranges during ExitBootServices? By leaving guest memory in different states of visibility, it's leaving a potential mess that the guest OS would need to handle, as now we're saying its legal for UEFI to have any range be host visible or not. Instead, it's much simpler to say that UEFI will revoke shared visibility on all ranges described by the memory map, and the guest OS will instead modify visibility as needed.

jyao1:
> > I think in this case we should report a new memory attribute that represents if a region of memory is shared or not. The existing EFI_MEMORY_CPU_CRYPTO seems to me to express a different semantic than what we really want. What do you think about adding a new memory attribute, something like EFI_MEMORY_SHARED_VISIBILITY that we could use to note ranges in the EFI memory map that are shared. This also allows different isolation architectures to report shared visibility on any type of memory, like conventional, unaccepted, etc.
> > @chris-oo I am OK to add new attribute. Would you please clarify the current usage of EFI_MEMORY_CPU_CRYPTO in guest? I am not aware of that.
> 
> From reading the specification the verbage around `EFI_MEMORY_CPU_CRYPTO` describes more of the capability of a memory region to be protected, not the current state of if it's protected or not.
> 
> However, after getting some feedback internally, what's the reason for UEFI not being able to revoke visibility of all memory ranges during ExitBootServices? By leaving guest memory in different states of visibility, it's leaving a potential mess that the guest OS would need to handle, as now we're saying its legal for UEFI to have any range be host visible or not. Instead, it's much simpler to say that UEFI will revoke shared visibility on all ranges described by the memory map, and the guest OS will instead modify visibility as needed.

@chris-oo There is shared memory allocated as DMA common buffer for virtual IO devices, especially for the asynchronized transfer. I am not sure if we FreeMemory at ExitBootService, which may cause memory map key change. Anyway, this memory is already there when OS loaded is invoked after ReadyToBoot.

ajfish:
> On Sep 10, 2020, at 4:04 PM, jyao1 <notifications@github.com> wrote:
> 
> 
> I think in this case we should report a new memory attribute that represents if a region of memory is shared or not. The existing EFI_MEMORY_CPU_CRYPTO seems to me to express a different semantic than what we really want. What do you think about adding a new memory attribute, something like EFI_MEMORY_SHARED_VISIBILITY that we could use to note ranges in the EFI memory map that are shared. This also allows different isolation architectures to report shared visibility on any type of memory, like conventional, unaccepted, etc.
> @chris-oo <https://github.com/chris-oo> I am OK to add new attribute. Would you please clarify the current usage of EFI_MEMORY_CPU_CRYPTO in guest? I am not aware of that.
> 
> From reading the specification the verbage around EFI_MEMORY_CPU_CRYPTO describes more of the capability of a memory region to be protected, not the current state of if it's protected or not.
> 
> However, after getting some feedback internally, what's the reason for UEFI not being able to revoke visibility of all memory ranges during ExitBootServices? By leaving guest memory in different states of visibility, it's leaving a potential mess that the guest OS would need to handle, as now we're saying its legal for UEFI to have any range be host visible or not. Instead, it's much simpler to say that UEFI will revoke shared visibility on all ranges described by the memory map, and the guest OS will instead modify visibility as needed.
> 
> @chris-oo <https://github.com/chris-oo> There is shared memory allocated as DMA common buffer for virtual IO devices, especially for the asynchronized transfer. I am not sure if we FreeMemory at ExitBootService, which may cause memory map key change. Anyway, this memory is already there when OS loaded is invoked after ReadyToBoot.
> 
There is no timer event activity going on while the drivers are processing the EVT_SIGNAL_EXIT_BOOT_SERVICES events, but DMA could be in flight. For well formed EVT_SIGNAL_EXIT_BOOT_SERVICES events they generally do simple things like turn off DMA. 

In terms of changing the memory map I don’t think that is needed. If you think about the memory map already is a contract and as soon as ExitBootServices() ownership of big chunks of the memory map transition from EFI control to OS control. We just need a well defined contract on how that freeing back happens. Today it is mostly that DMA into those freed back memory regions will stop, and debuggers will unhook after ExitBootServices() return. 

The common problem is the order of the EVT_SIGNAL_EXIT_BOOT_SERVICES is not defined. I think the edk2 implementation process them in order, but that is an implementation choice, and not architectural. I guess worst case we could define a protocol that allows a driver to change the memory state in its EVT_SIGNAL_EXIT_BOOT_SERVICES event vs. the CPU Driver that would do it generically. So basically a DMA buffer could be allocated in a way that makes its the responsibility of that drivers  EVT_SIGNAL_EXIT_BOOT_SERVICES to convert the memory state back. Think stop all DMA, call protocol member to to convert memory state. The contract could get a little wonky as it could be the PCI Root Bridge Driver that opts the DMA memory out, and the actual drivers EVT_SIGNAL_EXIT_BOOT_SERVICES that does the conversation. 

Sorry I don’t know the specifics really well, just trying to give some options. 

Thanks,

Andrew Fish

jyao1:
Thanks @ajfish 

So, the flow will be:
1) OS calls GetMemoryMap(), the DMA buffer is there and marked as BootServicesData.
2) OS calls ExitBootServices()
3) All EBS handler is called.
4) IOMMU driver will convert all shared DMA buffer to private buffer, and update the page table to clear S-bit. 

That means: After EBS, all memory should be private memory (ideally, because DMA buffer is freed).
Before EBS, there might be shared memory (as DMA buffer).

If OS/Loader need to know that information on shared/private before EBS, we need some indicator.
If OS/Loader don't need to know, then we can hide this information and no EFI_MEMORY_CPU_CRYPTO is needed. (IOMMU driver may expose optional private info for debug purpose.)

spbrogan:
@chris-oo  - I would like to open a new ECR with the PIWG for the hob changes.  Can you commit that doc in the next update.

chris-oo:
Sorry about the the late reply, been busy with some other things. 

@jyao1 I don't think the OS Loader should need to know memory is shared. Rather, we should stop all DMAs and revert their visibility to private as you described above, and say that all memory is private after EBS returns. Does that sound okay?

jyao1:
> Sorry about the the late reply, been busy with some other things.
> 
> @jyao1 I don't think the OS Loader should need to know memory is shared. Rather, we should stop all DMAs and revert their visibility to private as you described above, and say that all memory is private after EBS returns. Does that sound okay?

Yes. I agree. Then no EFI_MEMORY_CPU_CRYPTO is needed.

Now, I am waiting for PIWG ECR for GCD and Hob change. Do you please to draft it ?

chris-oo:
Yes. I will have a markdown using the same ECR template done tomorrow.

Unaccepted Memory UEFI ECR.md
chris-oo: FYI I have the ECR included in the PR. Please let me know if anything is missing, thanks!

corthon:
Since we're moving to 2008 as the default branch, it will be necessary to cherry-pick this forward after it's completed. (Optionally, you can retarget at 2008 now.) Attn: @chris-oo

jyao1:
@chris-oo - I just thought one more question on Unaccepted memory:
In UEFI 2.8, 2.3.4 x64 Platforms

```
Paging mode is enabled and any memory space defined by the UEFI memory map is identity
mapped (virtual address equals physical address), although the attributes of certain regions
may not have all read, write, and execute attributes or be unmarked for purposes of platform
protection. The mappings to other regions are undefined and may vary from implementation
to implementation.
```

What is the expectation for the unaccepted memory? Do we need build 1:1 mapping paging for it? Or it is NOT needed.

My proposal is: Since UEFI will not use it and OS cannot use it directly, there is NO need to create 1:1 mapping for the unaccepted memory. It may save the memory space required in virtual BIOS.
If OS wants to use accept it, the OS need create 1:1 mapping paging before use it.

Do you think it is OK?

spbrogan:
I agree.  While UEFI is active UEFI owns the page table and memory allocations.  Given there is no API in UEFI to accept memory (except by requesting allocation) then there is no reason to map it if it is not allocated in UEFI.  Once the OS has control (post exit boot services) it can accept, map, allocate, etc.

corthon (approved):

jyao1:
@chris-oo  and  @spbrogan

I just realize that we need another ECR for ACPI specification - 15 System Address Map Interfaces. Table 15-374 Address Range Types

I recommend we add  "8  AddressRangeUnaccepted"

Can we do that in this P-R as well?

chris-oo:
Regarding your earlier comment, I think it's fine to leave up to the individual platform for what it wants to do on Unaccepted memory regarding pagetables. Since there is no way currently to directly use unaccepted memory, it should be left up to the platform whether it wants to identity map those regions as well.

What exactly is the point of ACPI Chapter 15 System Address Map Interfaces? From my cursory reading it seems to be for support of hot-add ranges via Memory devices? Is that correct? Would this be in addition to the UEFI memory map?

chris-oo:
Additionally, @ajfish @jyao1 do you have any remaining comments on the UEFI ECR? I'd like to get sign off and close out on that one this week.

jyao1:
> Regarding your earlier comment, I think it's fine to leave up to the individual platform for what it wants to do on Unaccepted memory regarding pagetables. Since there is no way currently to directly use unaccepted memory, it should be left up to the platform whether it wants to identity map those regions as well.

Yes. agree. Can we add some sentence to the ECR?

> 
> What exactly is the point of ACPI Chapter 15 System Address Map Interfaces? From my cursory reading it seems to be for support of hot-add ranges via Memory devices? Is that correct? Would this be in addition to the UEFI memory map?

This is for legacy E820 table - to report the memory map in a non-UEFI system. Currently, many non-UEFI Linux systems are using this way to report the BIOS memory map. That is why it is described in ACPI specification.

jyao1:
> Additionally, @ajfish @jyao1 do you have any remaining comments on the UEFI ECR? I'd like to get sign off and close out on that one this week.

My only remaining comment is to add a sentence to describe that the unaccepted memory does not need identical paging in BIOS phase. I am asked by our OS people, so I prefer to make it clear.

chris-oo:
Okay I can do that. Let me think a bit about what I want to say there, and I'll push an update to the PR.

chris-oo:
@jyao1 I've added a short extension addition that explicitly calls out unaccepted memory as undefined, aka platform dependent. 

I'll have another ECR draft for the ACPI spec change soon as well.

jyao1:
> @jyao1 I've added a short extension addition that explicitly calls out unaccepted memory as undefined, aka platform dependent.
> 
> I'll have another ECR draft for the ACPI spec change soon as well.

thank you! look forward to the ACPI ECR.

MichaelARothman:
Since this concept could conceivably be used as an analogy to test/untest memory and have a usage model on bare metal, I'd propose some very minor tweaks to the language of the ECR submission into USWG as follows:

Original:
| EfiUnacceptedMemoryType ***Text Added*** | A memory region that represents unaccepted memory, that must be accepted by the guest before it can be used. Unless otherwise noted, all other EFI memory types are accepted. For platforms that support unaccepted memory, all unaccepted valid memory will be reported as unaccepted in the memory map. Unreported physical address ranges must be treated as not-present memory. ***Text Added*** |

Changed:
| EfiUnacceptedMemoryType ***Text Added*** | A memory region that represents unaccepted memory, that must be accepted by the boot target before it can be used. Unless otherwise noted, all other EFI memory types are accepted. For platforms that support unaccepted memory, all unaccepted valid memory will be reported as unaccepted in the memory map. Unreported physical address ranges must be treated as not-present memory. ***Text Added*** |

Original:
| EfiUnacceptedMemoryType ***Text Added*** | A memory region that represents unaccepted memory, that must be accepted by the guest before it can be used. Unless otherwise noted, all other EFI memory types are accepted. For platforms that support unaccepted memory, all unaccepted valid memory will be reported as unaccepted in the memory map. Unreported physical address ranges must be treated as not-present memory. ***Text Added*** |

Changed:
| EfiUnacceptedMemoryType ***Text Added*** | A memory region that represents unaccepted memory, that must be accepted by the boot target before it can be used. Unless otherwise noted, all other EFI memory types are accepted. For platforms that support unaccepted memory, all unaccepted valid memory will be reported as unaccepted in the memory map. Unreported physical address ranges must be treated as not-present memory. ***Text Added*** |

chris-oo:
I've updated the PR to use "boot target" instead of "guest" as you've suggested.

chris-oo:
@jyao1 Sorry about the delay, I thought I posted this before the new year but checking again I never did. Would you like to start a draft of the ACPI ECR changes you'd like to see? I'm more than happy to help you refine and review them.

jyao1:
> @jyao1 Sorry about the delay, I thought I posted this before the new year but checking again I never did. Would you like to start a draft of the ACPI ECR changes you'd like to see? I'm more than happy to help you refine and review them.

Yes. I will draft it and send for review.

jyao1:
Create ACPI ECR pull request at https://github.com/chris-oo/mu_basecore/pull/1.

Submit ACPI ECR - https://mantis.uefi.org/mantis/view.php?id=2196

corthon:
@chris-oo, is this PR still active? Can we close it?

spbrogan:
This was accepted into UEFI specification.  
Closing the PR

PR #74, opened 2020-10-17
Its late 2020 time to move default to VS2019

spbrogan:


corthon (dismissed):

corthon:
@spbrogan, this has been approved but never merged. It has now missed the 2102 integration. Should we retarget at 2102 and skip 2008?

corthon:
This should be retargeted at release/202102, and backported if necessary.

PR #80, opened 2021-02-10
Update nasm dependency to 2.15.5

spbrogan:
In this patch series update nasm dependency to 2.15.5 as there are new assembly files coming soon that will require this version.
Also clean up the iasl nuget publishing config/doc since it is no longer used.  For iasl see the project here https://dev.azure.com/projectmu/acpica


spbrogan:
will do in two commits to avoid mixing things

PR #83, opened 2021-06-14
Add Device State Platform Modes 4-7

joschock:
Add additional Device State Platform Modes 4-7.

### DIFF
--- a/MdeModulePkg/Include/Library/DeviceStateLib.h
+++ b/MdeModulePkg/Include/Library/DeviceStateLib.h
@@ -22,10 +22,14 @@ MU_CHANGE: new file
 #define DEVICE_STATE_UNIT_TEST_MODE            (1 << (5))
 #define DEVICE_STATE_MEM_PROTECTIONS_OFF       (1 << (6))
 
-#define DEVICE_STATE_PLATFORM_MODE_0           (1 << (24))
-#define DEVICE_STATE_PLATFORM_MODE_1           (1 << (25))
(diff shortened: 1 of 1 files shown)


mikeytdisco (approved):

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -22,6 +22,11 @@ MU_CHANGE: new file
 #define DEVICE_STATE_UNIT_TEST_MODE            (1 << (5))
 #define DEVICE_STATE_MEM_PROTECTIONS_OFF       (1 << (6))
 
+#define DEVICE_STATE_PLATFORM_MODE_4           (1 << (20))
```
spbrogan: John - As far as you are aware is any entity using the raw value?  Should we just move 0-3 down in bit field and then add 4-7 on top?  Since this hasn't been upstreamed to edk2 we have more flexibility.

joschock: AFAIK, no one is using the raw value; I'd be happy to move it around so the bits are in order if you prefer - I just wasn't sure how baked the existing definitions were.

spbrogan: Everything we have a macro based and should handle it fine....Transition for a device in the field should be transparent as well since these are not cached but detected on each boot, right?

joschock: Updated definitions.

spbrogan (approved):

PR #86, opened 2021-07-06
Add Initial Support for Dynamic Variable Flash Info in Standalone MM

makubacki:
Adds a new GUID that is used to identify a HOB that passes variable
flash information to UEFI variable drivers in HOB consumption phases
such as DXE, Traditional MM, and Standalone MM.

This information was previously passed directly with PCDs such
as EfiMdeModulePkgTokenSpaceGuid.PcdFlashNvStorageVariableBase
and gEfiMdeModulePkgTokenSpaceGuid.PcdFlashNvStorageVariableSize.
However, the Standalone MM variable driver instance does not have
direct access to the PCD database. Therefore, this HOB will first
be considered as the source for variable flash information and
if platforms do not produce the HOB, reading the information from
the PCDs directly will be a backup to provide backward compatibility.

As a consequence, these changes are backward compatible with existing
platforms. Platform code is expected to produce the variable flash
information HOB if the platform needs dynamic variable flash support in
a Standalone MM environment.

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableNonVolatile.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableNonVolatile.c
@@ -134,6 +134,7 @@ InitRealNonVolatileVariableStore (
 {
   EFI_FIRMWARE_VOLUME_HEADER            *FvHeader;
   VARIABLE_STORE_HEADER                 *VariableStore;
+  VARIABLE_FLASH_INFO                   *VariableFlashInfo;  // MU_CHANGE TCBZ3479 - Add Variable Flash Information HOB
   UINT32                                VariableStoreLength;
   EFI_HOB_GUID_TYPE                     *GuidHob;
   EFI_PHYSICAL_ADDRESS                  NvStorageBase;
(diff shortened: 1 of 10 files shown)


joschock (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableNonVolatile.c
```diff
@@ -150,19 +151,31 @@ InitRealNonVolatileVariableStore (
...
   //
-  NvStorageSize = PcdGet32 (PcdFlashNvStorageVariableSize);
   NvStorageData = AllocateRuntimeZeroPool (NvStorageSize);
   if (NvStorageData == NULL) {
     return EFI_OUT_OF_RESOURCES;
   }
 
-  NvStorageBase = NV_STORAGE_VARIABLE_BASE;
-  ASSERT (NvStorageBase != 0);
-
```
TaylorBeebe, line 165: Are these MU_CHANGEs?

TaylorBeebe (approved):

TaylorBeebe (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableNonVolatile.c
makubacki: They are being removed. For me, MU_CHANGE comments would be a bit more confusing but I can add them if they're deemed helpful here.

kuqin12:
Is this a proposal that we can essentially consume dynamic Pcd for NV storage information again?

spbrogan:
both commit titles needs to include TCBZ3479 so that we can track removal/rebase

spbrogan (approved):

makubacki:
> Is this a proposal that we can essentially consume dynamic Pcd for NV storage information again?

It allows the NV storage information on SPI flash to be communicated to the variable driver in all scenarios including Standalone MM. PCDs will continue to work the way they have. But if the HOB introduced here is present, it will be used instead of the PCDs.

makubacki:
> both commit titles needs to include TCBZ3479 so that we can track removal/rebase

Commit titles updated.

PR #87, opened 2021-07-17
Route raw SMC calls to the StandaloneMMPkg

rohansenUSA:
**This PR enables routing of SMC calls without an EFI_MM_COMMUNICATE_HEADER to the StandalonMMPkg**

SMC calls without an EFI_MM_COMMUNICATE_HEADER will have an Event ID different from ARM_SMC_ID_MM_COMMUNICATE_AARCH64. 
These calls will be forwarded to handlers that has registered with the HandlerType of gArmRawSmcCallGuid. The handler will be receiving the Event ID in the CommBuffer.


corthon (changes requested):
Should be closed. Will be workshopped internally and resubmitted.

PR #92, opened 2021-09-23
Remove save-state context after unit test is complete

mapaig:
The current unit test framework leaves behind a save-state cache file after completing unit tests. Subsequent test attempts can cause failures due to consuming old save-state context, unless the file has been manually removed.

This change adds save-state cache file cleanup after unit tests are completed to prevent consumption of older unit test context.

### DIFF
--- a/UnitTestFrameworkPkg/Library/UnitTestLib/UnitTestLib.c
+++ b/UnitTestFrameworkPkg/Library/UnitTestLib/UnitTestLib.c
@@ -149,6 +149,18 @@ FreeUnitTestFramework (
   IN UNIT_TEST_FRAMEWORK_HANDLE  FrameworkHandle
   )
 {
+  EFI_STATUS Status;
+
+  //
+  // If there is a persisted context, delete it now
+  //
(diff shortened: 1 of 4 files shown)


spbrogan:
This is actually by design but I agree it also causes problems especially to those that are new to the framework or using it in ways that don't require reboots/automation.   

Many of UEFI/shell based (on device) tests actually need a reboot.  To support this in automation you make a startup.nsh with all of your tests listed.  Then each time it reboots the shell launches and runs the startup script.  Each test that has already run just outputs its report because the cache file shows that the test is done.  Once it gets to the test which did the reboot it will resume testing.  Without the cache file automation wouldn't work.

Happy to hear if you have ideas?

spbrogan (changes requested):
Any thoughts on the comments left

mapaig:
I think this can be closed. I was seeing an issue with automated SmmPagingProtections test related to the cache file, but I'm not seeing that anymore and it doesn't look like this solution fits with some automated use cases anyway.

PR #93, opened 2021-09-28
[TCBZ3662]Fix a missing parameter in DEBUG print in XhciInitializeDeviceSlot()

joschock:
Fix a debug print in XhciSched that specified a parameter in the format string but did not supply the actual parameter.


makubacki:
It would be good to have a TCBZ and change tag to help track it.

joschock:
> It would be good to have a TCBZ and change tag to help track it.

Done - thanks.

makubacki (approved):

spbrogan:
What about putting out the status for the failure case?  Otherwise looks like a good fix and should help debugging if this problem occurs.

joschock:
> What about putting out the status for the failure case? Otherwise looks like a good fix and should help debugging if this problem occurs.

Added a status print.

spbrogan (approved):

PR #94, opened 2021-12-08
Adding logging errors to console for SpellCheck and MarkDownLintCheck

bkarstens:
Added lines to log errors to console for SpellCheck (if AuditOnly is False) and MarkDownLintCheck. This change brings these plugins in line with the other CI plugins where errors can be quickly seen in the console instead of digging into the build log.

### DIFF
--- a/.pytool/Plugin/MarkdownLintCheck/MarkdownLintCheck.py
+++ b/.pytool/Plugin/MarkdownLintCheck/MarkdownLintCheck.py
@@ -144,6 +144,7 @@ class MarkdownLintCheck(ICiBuildPlugin):
         results = self._check_markdown(path_to_check, config_file_path, Ignores)
         for r in results:
             tc.LogStdError(r.strip())
+            logging.error(r.strip())
 
         # add result to test case
         overall_status = len(results)
--- a/.pytool/Plugin/MarkdownLintCheck/MarkdownLintCheck.py
+++ b/.pytool/Plugin/MarkdownLintCheck/MarkdownLintCheck.py
@@ -66,6 +66,9 @@ class MarkdownLintCheck(ICiBuildPlugin):
 
     def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM, PLMHelper, tc, output_stream=None):
         Errors = []
+        audit_only = True
+        if "AuditOnly" in pkgconfig and not pkgconfig["AuditOnly"]:
+            audit_only = False
 
         abs_pkg_path = Edk2pathObj.GetAbsolutePathOnThisSystemFromEdk2RelativePath(
             packagename)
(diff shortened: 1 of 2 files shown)


bkarstens:
Maybe this should actually go in Tianocore 😐

spbrogan:
i have looked at this but worried it isn't consistant with other plugins.  As i know the build plugin doesn't' show the failure unless you look in the log.  That is kind of the idea of the console log.  Anyway, I'll do a little more research but i wanted to ack the PR at least.  Thanks.

bkarstens:
> i have looked at this but worried it isn't consistant with other plugins. As i know the build plugin doesn't' show the failure unless you look in the log. That is kind of the idea of the console log. Anyway, I'll do a little more research but i wanted to ack the PR at least. Thanks.

@spbrogan - When I run stuart_ci_build, I see the the other ci plugins in the console. They seem to already have lines to log errors to the console:

- https://github.com/microsoft/mu_basecore/blob/31238d9688cf55d0db96b3b557f15587087d02a9/.pytool/Plugin/CharEncodingCheck/CharEncodingCheck.py#L114
- https://github.com/microsoft/mu_basecore/blob/31238d9688cf55d0db96b3b557f15587087d02a9/.pytool/Plugin/GuidCheck/GuidCheck.py#L219
- https://github.com/microsoft/mu_basecore/blob/31238d9688cf55d0db96b3b557f15587087d02a9/.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py#L123
- https://github.com/microsoft/mu_basecore/blob/31238d9688cf55d0db96b3b557f15587087d02a9/.pytool/Plugin/DependencyCheck/DependencyCheck.py#L111
- https://github.com/microsoft/mu_basecore/blob/31238d9688cf55d0db96b3b557f15587087d02a9/.pytool/Plugin/LibraryClassCheck/LibraryClassCheck.py#L139

.pytool/Plugin/MarkdownLintCheck/MarkdownLintCheck.py
```diff
@@ -144,6 +144,7 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
         results = self._check_markdown(path_to_check, config_file_path, Ignores)
         for r in results:
             tc.LogStdError(r.strip())
+            logging.error(r.strip())
```
spbrogan: So it looks like you have a new pattern that says if audit_only if false then report the errors to the screen at elevated level.  If audit_only then don't.  

I would be good with that pattern.  

What do you think?

spbrogan: sorry...the reason these tests have an AUDIT_ONLY flag is because they are noisy and don't have 100% opt in.  So we didn't want to mess up the log with stuff that is in AUDIT_ONLY

bkarstens: I've updated the markdownlintcheck to also be gated on `not audit_only`.

spbrogan (approved):

PR #101, opened 2022-02-22
Reset USB port during enumeration if GetPortStatus returns device error

akulkar4:
During USB device enumeration, issuing a hot reset on a port is skipped if there
is a reset change status already detected on the port. This can happen when
enumerating devices after a host controller soft reset (which drives a hot reset
down the ports).

However, in certain cases an attached device may not be responsive even if the
reset change and connection status bits are set. For e.g., according to xHCI
spec section 4.19.5.1 the port reset change bits can be set when a hot reset
driven on the port transitions to a warm reset and completes with errors. For
such instances it is worthwhile to force a hot reset during enumeration to try
and recover unresponsive devices.

During enumeration check whether querying port status returns EFI_DEVICE_ERROR
and try a port reset if there is a device attached to the port.

### DIFF
--- a/MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
+++ b/MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
@@ -458,7 +458,7 @@ XhcGetRootHubPortStatus (
   // For those devices behind hub, we get its attach/detach event by hooking Get_Port_Status request at control transfer for those hub.
   //
   ParentRouteChart.Dword = 0;
-  Status = XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus);  // MU_CHANGE
+  Status                 = XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus); // MU_CHANGE
 
 ON_EXIT:
   gBS->RestoreTPL (OldTpl);
--- a/MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
+++ b/MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
@@ -458,7 +458,7 @@ XhcGetRootHubPortStatus (
   // For those devices behind hub, we get its attach/detach event by hooking Get_Port_Status request at control transfer for those hub.
   //
   ParentRouteChart.Dword = 0;
-  XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus);
+  Status = XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus);  // MU_CHANGE
 
 ON_EXIT:
   gBS->RestoreTPL (OldTpl);
--- a/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
+++ b/MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
@@ -913,7 +913,8 @@ UsbEnumeratePort (
   //
   Status = HubApi->GetPortStatus (HubIf, Port, &PortState);
 
-  if (EFI_ERROR (Status)) {
+  // MU_CHANGE - try a port reset if get port status returns device error
+  if (EFI_ERROR (Status) && (Status != EFI_DEVICE_ERROR)) {
     DEBUG ((DEBUG_ERROR, "UsbEnumeratePort: failed to get state of port %d\n", Port));
     return Status;
   }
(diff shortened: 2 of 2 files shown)


MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
```diff
@@ -458,7 +458,7 @@ XhcGetRootHubPortStatus (
   // For those devices behind hub, we get its attach/detach event by hooking Get_Port_Status request at control transfer for those hub.
   //
   ParentRouteChart.Dword = 0;
-  XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus);
+  Status                 = XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus); // MU_CHANGE
```
corthon, line 461: Why so much whitespace?

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
```diff
@@ -995,7 +995,9 @@ UsbEnumeratePort (
     // Now, new device connected, enumerate and configure the device
     //
     DEBUG ((DEBUG_INFO, "UsbEnumeratePort: new device connected at port %d\n", Port));
-    if (USB_BIT_IS_SET (PortState.PortChangeStatus, USB_PORT_STAT_C_RESET)) {
+    if (USB_BIT_IS_SET (PortState.PortChangeStatus, USB_PORT_STAT_C_RESET) &&
+        (Status != EFI_DEVICE_ERROR))     // MU_CHANGE
+    {
```
corthon, line 1006: This doesn't look right. It looks like we'd be forcing the first branch, and never taking the second branch. The only thing that sets Status seems to be the GetPortStatus at the top of the function, and if Status were EFI_DEVICE_ERROR we would have already returned. Therefore, Status == EFI_SUCCESS and this will always return true, right?

corthon (changes requested):

MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
akulkar4: This was because the uncrustify check was failing if it was unaligned with the `=` on line 460: 
```
Status = XhcPollPortStatusChange (Xhc, ParentRouteChart, PortNumber, PortStatus); // MU_CHANGE
```

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
akulkar4: Ah, that's true my initial revision had this on line 916, which was omitted while rebasing to 202111, will update it. 
```C
if (EFI_ERROR (Status)) && (Status != EFI_DEVICE_ERROR))
```

MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
corthon: Ah, of course. It was [visually] masked by the red "removed" line.

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
akulkar4: Fixed in the latest. We'll now not return immediately on `EFI_DEVICE_ERROR` but instead try a reset if there is a device connected. If not, Status (received from `GetPortStatus`) is returned at the end of the function anyway so `EFI_DEVICE_ERROR` will still be returned but it will happen later.

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
```diff
@@ -913,7 +913,8 @@ UsbEnumeratePort (
   //
   Status = HubApi->GetPortStatus (HubIf, Port, &PortState);
 
-  if (EFI_ERROR (Status)) {
+  // MU_CHANGE - try a port reset if get port status returns device error
+  if (EFI_ERROR (Status) && (Status != EFI_DEVICE_ERROR)) {
```
corthon, line 922: If we're continuing on EFI_DEVICE_ERROR, should we initialize `PortStatus` to some predictable value? Right now it's used in the tests below, but we have no guarantee that it has been initialized.

corthon (changes requested):

corthon:
Many apologies for the latency on the feedback. Updates to Github PRs go to a different email address which was marking them as unimportant. I've marked them so they land in my inbox.

MdeModulePkg/Bus/Usb/UsbBusDxe/UsbEnumer.c
akulkar4: Yes, zeroing it out initially sounds like a good idea. `GetPortStatus` implementations for Xhci, Uhci and Ehci do it, but it would be good to not rely on the host controller side implementation in case it comes back uninitialized on the error path.

corthon (approved):

PR #103, opened 2022-03-10
Add support for in-place updates of .inf files to OverrideValidation.py

joschock:
This adds a feature whereby OverrideValidation can be used to regenerate existing override tags in an .inf file "in-place". This basically allows batch-updates of override tags within an INF file without having to update each tag manually. 

Example usage:
```
>OverrideValidation.py -w c:\src -r c:\src\FooPkg\OverridingModule.inf
Updating:
#Override : 00000002 | BarPkg/OverridenModule.inf | 143b08782a2abc620d1eb57461c6e290 | 2022-03-10T23-09-45 | 6f8c53a3fcd79b202c708e7aa58256cafbf24bc4

>git diff
diff --git a/FooPkg/OverridingModule.inf b/FooPkg/OverridingModule.inf    
index 2d4ca47299..90da207a39 100644
--- a/FooPkg/OverridingModule.inf  
+++ b/FooPkg/OverridingModule.inf  
@@ -8,7 +8,7 @@
 # 
 # 
-#Override : 00000002 | BarPkg/OverridenModule.inf | 4f7eed98e3c084eecdff5fa2e1e57db1 | 2021-11-23T21-41-21 | 44b40c0358489da6c444e7cfb2be26e56b7c16a1
+#Override : 00000002 | BarPkg/OverridenModule.inf | 143b08782a2abc620d1eb57461c6e290 | 2022-03-10T23-09-45 | 6f8c53a3fcd79b202c708e7aa58256cafbf24bc4
 #

### CODE REVIEW

File: BaseTools/Plugin/OverrideValidation/OverrideValidation.py [hunk 1]
```diff
@@ -574,6 +575,14 @@ def path_parse():
         '-t', '--targetpath', dest = 'TargetPath', type=str,
         help = '''Specify the absolute path to your target module/file/folder by passing t Path/To/Target or --targetpath Path/To/Target.'''
         )
+    group.add_argument (
+        '-r', '--regenpath', dest = 'RegenPath', type=str,
+        help = '''Specify the absolute path to an inf with existing overrides to regen by passing r Path/To/Target or --regenpath Path/To/Target.'''
+        )
+    parser.add_argument (
+        '-p', '--prefixes', dest = 'RegenPrefix', nargs="*", default=None,
```
kuqin12: Can we please rename this argument to be package path? (assuming that is what the purpose of this argument?)

kuqin12:
Could you please add the PR description for usage into `BaseTools\Plugin\OverrideValidation\ReadMe.md` as well? Thanks!

BaseTools/Plugin/OverrideValidation/OverrideValidation.py
```diff
@@ -628,20 +657,61 @@ def path_parse():
...
+
+        RegenInfOutData = ""
+        for line in RegenInfData.splitlines (True):
+            match = v1_regex.match(line)
+            if match is None:
+                match = v2_regex.match(line)
+
+            if match is not None:
+                rel_path = match.group(3)
+                abs_path = resolvePath(path_prefixes, rel_path)
```
kuqin12: This function seems to be the same as GetAbsolutePathOnThisSystemFromEdk2RelativePath from `edk2toollib.uefi.edk2.path_utilities import Edk2Path`, might want to use the library function instead of a local implementation.

BaseTools/Plugin/OverrideValidation/OverrideValidation.py
```diff
@@ -596,22 +605,42 @@ def path_parse():
...
+        # For a drive root, this will rip off the os.sep
+        if not os.path.normcase(Paths.RegenPath).startswith(os.path.normcase(Paths.WorkSpace.rstrip(os.sep)) + os.sep):
+            raise RuntimeError("Module is not within specified Workspace.")
 
     return Paths
 
+# resolve relative paths to absolute path over the given list of prefixes.
+# prefixes: a list of path prefixes to check. if more than one prefix matches, those closer to the front of the list are preferred
+# relpath: a relative path to resolve against the prefixes.
+def resolvePath(prefixes, relpath):
```
kuqin12: Same as the previous comment: if we could use the library method, we might be able to drop this implementation.

kuqin12 (changes requested):

BaseTools/Plugin/OverrideValidation/OverrideValidation.py
```diff
@@ -628,20 +657,61 @@ def path_parse():
...
+
+                mod_hash = ModuleHashCal(abs_path)
+                # only update the line if the hash has changed - this ensures the timestamp tracks actual changes rather than last time it was run.
+                if (mod_hash != match.group(4)):
+                    VERSION_INDEX = Paths.Version - 1
+
+                    if VERSION_INDEX == 0:
+                        line = '#%s : %08d | %s | %s | %s\n' % (match.group(1), FORMAT_VERSION_1[0], rel_path, mod_hash, datetime.utcnow().strftime("%Y-%m-%dT%H-%M-%S"))
+                    elif VERSION_INDEX == 1:
+                        git_hash = ModuleGitHash(abs_path)
```
kuqin12: Could this abs_path potentially come from a track tag, that does not exist in the current repo configuration? Maybe we should check it?

joschock: well, it has to exist in order for the abs_path to resolve. So if you want to use the tool in this mode, all the paths referenced in the target .inf need to exist. I could modify the code to print a warning and skip if the path doesn't exist, is that more desirable?

joschock:
> Could you please add the PR description for usage into `BaseTools\Plugin\OverrideValidation\ReadMe.md` as well? Thanks!

Done.

BaseTools/Plugin/OverrideValidation/OverrideValidation.py
joschock: Done.

BaseTools/Plugin/OverrideValidation/OverrideValidation.py
joschock: Done.

BaseTools/Plugin/OverrideValidation/OverrideValidation.py
joschock: done.

kuqin12 (approved):

PR #106, opened 2022-03-25
Removed comments to enable acpiview shell command

JoeLopez333:
With the previous MS Changes, the UEFI shell command "Acpiview" will attempt to dereference an uninitialized pointer, potentially causing an exception. These code changes restore acpiview functionality.


corthon:
@spbrogan , 
1) I didn't recall why we needed to turn this off in the first place. It was in response to VS2017, I believe, but don't remember what was failing. I *do* know that the ACPI view code is pretty crap and has all kinds of memory problems that we've found but haven't wanted to have the fight with upstream.
2) Do we want this to hit 2202 *first* and then backport?

spbrogan (approved):

PR #123, opened 2022-10-04
Add line ending check plugin [Rebase & FF]

makubacki:
## Description

Adds a simple CI plugin to check file line endings. This is meant to:

1. Supplement the Uncrustify Check plugin which fails if line endings
   are CRLF but does not obviously indicate that error since it is
   non-visible whitespace.

2. Discover line endings issues in files not checked by Uncrustify
   (which only operates against .c and .h files right now).

Pre-existing line ending issues (LF instead of CRLF are fixed in the first commit).

- [ ] Breaking change?
  **Possibly** - This plugin will break the CI build for any package that does not have CRLF
  line endings.

## How This Was Tested

1. Run against all packages in mu_basecore and other platform packages.
2. Verified ignore file list works as expected.
3. Verified line endings are interpreted as expected.
4. Verified plugin returns success and failure in expected cases.

## Integration Instructions

Check plugin results when integrating this mu_basecore change.


.pytool/Plugin/LineEndingCheck/LineEndingCheck.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if abs_pkg_path is None:
+            tc.SetSkipped()
+            tc.LogStdError(f"Package folder not found {abs_pkg_path}")
+            return 0
+
+        all_files = [n for n in glob.glob(os.path.join(abs_pkg_path, '*.*'),
+                     recursive=True)]
+
+        if "IgnoreFiles" in package_config:
```
spbrogan: why not leverage the more capable gitignore syntax?

.pytool/Plugin/LineEndingCheck/Readme.md
```diff
@@ -0,0 +1,19 @@
+# Line Ending Check Plugin
+
+This CiBuildPlugin scans all the files in a package to verify that the line endings are CRLF.
+
+## Configuration
+
+The plugin can be configured to ignore certain files.
+
+``` yaml
+"CharEncodingCheck": {
```
spbrogan: this looks incorrect.  Copy and paste error?

.pytool/Plugin/LineEndingCheck/LineEndingCheck.py
makubacki: Updated to use gitignore syntax

.pytool/Plugin/LineEndingCheck/Readme.md
makubacki: Yes. Forgot to search the readme. Updated.

makubacki:
@spbrogan, all feedback is addressed.

.pytool/Plugin/LineEndingCheck/Readme.md
```diff
@@ -0,0 +1,19 @@
...
+"LineEndingCheck": {
+    "IgnoreFiles": []
+}
+```
+
+### IgnoreFiles
+
+An **optional** list of Unix style filename patterns used to exclude files from being checked.
+
+See [fnmatch](https://docs.python.org/3/library/fnmatch.html#module-fnmatch) for more details.
```
spbrogan: is fnmatch still relevant?

.pytool/Plugin/LineEndingCheck/Readme.md
```diff
@@ -0,0 +1,19 @@
...
+
+``` yaml
+"LineEndingCheck": {
+    "IgnoreFiles": []
+}
+```
+
+### IgnoreFiles
+
+An **optional** list of Unix style filename patterns used to exclude files from being checked.
```
spbrogan: Is this still accurate?  Are we saying that .gitignore syntax is the same as unix style filename pattern?  Doesn't it also work with directories?  Should we mention, they must be relative from package root?

.pytool/Plugin/LineEndingCheck/Readme.md
```diff
@@ -0,0 +1,19 @@
+# Line Ending Check Plugin
```
spbrogan, line 1: Should there also be a comment that this plugin is really only relevant if the workspace is checked out on Linux or with autocrlf=false

.pytool/Plugin/LineEndingCheck/Readme.md
makubacki: No. I removed this section.

.pytool/Plugin/LineEndingCheck/Readme.md
makubacki: Added guidance in https://github.com/microsoft/mu_basecore/pull/123/commits/d69740acdb0747e1e37dd1c867a271212e16b73c

.pytool/Plugin/LineEndingCheck/Readme.md
makubacki: Yes, the git ignore syntax is more robust. Along with the fnmatch link, I forgot to update this section. It's updated now.

https://github.com/microsoft/mu_basecore/pull/123/commits/7211bef603a41285dd60cb6de852f7cfb1b04e8c

.pytool/Plugin/LineEndingCheck/Readme.md
makubacki: https://github.com/microsoft/mu_basecore/pull/123/commits/7211bef603a41285dd60cb6de852f7cfb1b04e8c

spbrogan (approved):

PR #126, opened 2022-10-13
Fix buffer overflow when merging guard pages in MergeMemoryMap

cfernald:
## Description

Checks that the next map entry is valid before dereferencing to merge the guard pages. If the final entry is at the end of a page with no valid page following it, then this can cause an access violation.

## How This Was Tested

Tested on Q35 platform boot.

## Integration Instructions

N/A


MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c
```diff
@@ -464,7 +464,13 @@ MergeMemoryMap (
     NextMemoryMapEntry = NEXT_MEMORY_DESCRIPTOR (MemoryMapEntry, DescriptorSize);
```
makubacki, line 464: Not entirely related to your exact change but walking through the surrounding area...

The API says, `DescriptorSize` is the size in bytes of an individual `EFI_MEMORY_DESCRIPTOR` and that is passed as the "current `EFI_MEMORY_DESCRIPTOR` size" to `NEXT_MEMORY_DESCRIPTOR ()`.

Why does the `CopyMem ()` call use `sizeof (EFI_MEMORY_DESCRIPTOR)`?

makubacki (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c
cfernald: Yeah, definitely seems like an over-sight in this original change. In reality it's somewhat moot as the internal size here comes from the same definition, but I will go ahead and change this to make the routine consistent

TaylorBeebe: Because each element is of type EFI_MEMORY_DESCRIPTOR (even though the actual memory used for each descriptor is greater), doesn't that mean the portion containing the actual descriptor will be consistently stored at the start of each entry? If that's the case, then the CopyMem is still correct regardless of what DescriptorSize is.

TaylorBeebe (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryAttributesTable.c
cfernald: I think it depends on the reality vs the interface of the function. The interface looks designed to allow the caller to have a large definition of the structure then the function. I assumes this is similar to the external EFI interfaces that do this to allow the caller to iterate even if they don't know the size of the newer sturcture. For an internal function like this its pretty much moot though. In the end it probably makes no difference one way or the other.

makubacki: I'm not sure it practically matters here either.

But, if the interface defines the `EFI_MEMORY_DESCRIPTOR` size to be set via the `DescriptorSize` parameter and the function uses `EFI_MEMORY_DESCRIPTOR` instead of `DescriptorSize`, that's an implementation bug to fail to conform to the interface. I just couldn't tell without much more investigation if that's the case or something else is going on.

My criteria to change is:

1. Is it correct?
2. Is it consistent?
3. Is it confusing?

For me, it seems to arguably fail (1) and fail (2)/(3). (2) because the call one line below to `NEXT_MEMORY_DESCRIPTOR ()` references to the size of the same variable as `DescriptorSize`. And (3) because of (2) and the effort required to subsequently figure out what's going on.

If that's the case, I think we should save future people trouble and change it or at least leave a comment saying why it is done this way.

cfernald: agreed, fixing this.

PR #134, opened 2022-10-19
Merge Image Record Lists Before Breaking up the Memory Map

TaylorBeebe:
## Description

Now that we allow for protected and non-protected image lists, we need to merge them together (if necessary) before passing them into the memory map breakup logic. This will enable images to be loaded and non-protected without affecting our more general NX protection policy and any future mitigations we may add.

## Breaking change?

No

## How This Was Tested

Forcing a split between protected and nonprotected images and booting to shell to examine the outcome of the memory map breakup logic

## Integration Instructions

N/A


MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1786,28 +1896,88 @@ GetMemoryMapWithPopulatedAccessAttributes (
...
 
   NumberOfDescriptors   = *MemoryMapSize / *DescriptorSize;
   NumberOfBitmapEntries = (NumberOfDescriptors % 8) == 0 ? NumberOfDescriptors : (((NumberOfDescriptors / 8) * 8) + 8);
 
   Bitmap = AllocateZeroPool (NumberOfBitmapEntries / 8);
 
+  if (Bitmap == NULL) {
+    FreePool (*MemoryMap);
+    *MemoryMapSize  = 0;
+    *DescriptorSize = 0;
```
apop5: I think there are some memory leaks in the error conditions.

ArrayOfListEntryPointers

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1655,9 +1655,43 @@ SyncBitmapAndSetAccessAttributesInMemoryMap (
...
+                                        MemoryMap buffer.
+  @param[in, out]   MemoryMap           A pointer to the current memory map.
+  @param[in]        DescriptorSize      Size, in bytes, of an individual EFI_MEMORY_DESCRIPTOR.
+  @param[in]        Bitmap              Pointer to the beginning of the bitmap to be updated
+**/
+STATIC
+VOID
+SetAccessAttributesInMemoryMap (
+  IN CONST UINTN             *MemoryMapSize,
+  IN  EFI_MEMORY_DESCRIPTOR  *MemoryMap,
```
os-d: nit: Should this be `IN OUT`? Or comment changed to `IN` only?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1655,9 +1655,43 @@ SyncBitmapAndSetAccessAttributesInMemoryMap (
...
+  IN CONST UINTN             *DescriptorSize,
+  IN CONST UINT8             *Bitmap
+  )
+{
+  EFI_MEMORY_DESCRIPTOR  *MemoryMapEntry;
+  EFI_MEMORY_DESCRIPTOR  *MemoryMapEnd;
+  UINTN                  Index = 0;
+
+  MemoryMapEntry = MemoryMap;
+  MemoryMapEnd   = (EFI_MEMORY_DESCRIPTOR *)((UINT8 *)MemoryMap + *MemoryMapSize);
```
os-d, line 2022: Do we need to check MemoryMapSize isn't null before dereferencing or we are confident that the callers populated it correctly?

os-d, line 2022: Do we need to check MemoryMapSize isn't null before dereferencing or we are confident that the callers populated it correctly?

kenlautner (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
os-d: I see this is only called in one place below, where it is checked as input, so seems we are confident (CodeQL still might complain).

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Added an OutOfResourcesCleanup section with gotos whenever an allocation fails and also free ArrayOfListEntryPointers in the non-error case.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Updated to IN OUT :)

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Yes, added NULL checks for all input paramters

TaylorBeebe: Added NULL checks for all input parameters

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1626,22 +1627,20 @@ FilterMemoryMapAttributes (
 }
 
 /**
-  Access attributes in the memory map based on the memory protection policy and
-  mark visited regions in the bitmap.
+  Set every bit in the bitmap which corrosponds to a memory map descriptor with nonzero attributes.
```
makubacki:
```suggestion
  Set every bit in the bitmap which corresponds to a memory map descriptor with nonzero attributes.

```

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1667,6 +1705,79 @@ SyncBitmapAndSetAccessAttributesInMemoryMap (
...
+  IN CONST  UINTN                  *DescriptorSize,
+  IN        LIST_ENTRY             *NonProtectedImageList
+  )
+{
+  LIST_ENTRY               *NonProtectedImageRecordLink = NULL;
+  IMAGE_PROPERTIES_RECORD  *NonProtectedImageRecord     = NULL;
+  EFI_MEMORY_DESCRIPTOR    *MemoryMapEntry              = NULL;
+  EFI_MEMORY_DESCRIPTOR    *MemoryMapEnd                = NULL;
+
+  if ((MemoryMapSize == 0) || (MemoryMap == NULL) ||
```
makubacki: This is technically okay, but did you mean to check the value to be zero (in addition to the pointer being non-NULL)?

The edk2 style guideline prefers null checks to be against `NULL` (instead of `0`) if just checking the pointer.

makubacki (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Was a mistake - fixed

PR #137, opened 2022-10-21
Add Memory Protection Special Region Reporting Interface

TaylorBeebe:
## Description

When we initialize memory protections, we need to enable platforms to report regions which should have special protections not based on their EFI memory type similarly to what we do with loaded images. One example where this is needed is the interrupt entry code buffer which is allocated as EFiBootServicesCode in early DXE. Many platforms will choose to enable NX protections on code memory types to ensure all memory is either read/write OR execute at any time, but when this buffer is allocated, we do not yet have the ability to set attributes on it because the CPU architectural protocol has not been installed. So, we need to report this region so when memory protections are being initialized a special action can be taken to ensure correct attributes are applied to the buffer.

This PR adds the infrastructure to enable these special actions when initializing memory protections. Special memory regions are reported from PEI by creating a guided HOB, and special regions are fetched and added in DXE via protocol.

## Breaking change?
No

## How This Was Tested

Building HOB entries to ensure they are being collected in DXE,and reporting regions in early DXE.

## Integration Instructions

N/A


MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,284 @@ SetUefiImageMemoryAttributes (
...
+
+  GuidHob = GetFirstGuidHob (&gMemoryProtectionSpecialRegionHobGuid);
+
+  while (GuidHob != NULL) {
+    DEBUG ((DEBUG_INFO, "%a - 1\n", __FUNCTION__));
+    HobSpecialRegion = (MEMORY_PROTECTION_SPECIAL_REGION *)GET_GUID_HOB_DATA (GuidHob);
+    NewSpecialRegion = AllocateCopyPool (sizeof (MEMORY_PROTECTION_SPECIAL_REGION_LIST_ENTRY), HobSpecialRegion);
+
+    if (NewSpecialRegion == NULL) {
+      return EFI_OUT_OF_RESOURCES;
```
apop5, line 362: could you add a debug message here to report the  case where there are multiple hobs, and only the last one fails to get memory allocated..

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,284 @@ SetUefiImageMemoryAttributes (
...
+EFIAPI
+AddSpecialRegion (
+  IN EFI_PHYSICAL_ADDRESS  Start,
+  IN EFI_PHYSICAL_ADDRESS  Length,
+  IN UINT64                Attributes
+  )
+{
+  MEMORY_PROTECTION_SPECIAL_REGION_LIST_ENTRY  *SpecialRegionEntry = NULL;
+  EFI_PHYSICAL_ADDRESS                         AlignedStart        = 0;
+  EFI_PHYSICAL_ADDRESS                         AlignedLength       = 0;
```
apop5: If this is a length, even though EFI_PHYSICAL_ADDRESS resolves to a UINT64, it would be better to call this a UINT64.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,284 @@ SetUefiImageMemoryAttributes (
...
+
+    GuidHob = GetNextGuidHob (&gMemoryProtectionSpecialRegionHobGuid, GET_NEXT_HOB (GuidHob));
+  }
+
+  return EFI_SUCCESS;
+}
+
+/**
+  Create a sorted array of MEMORY_PROTECTION_SPECIAL_REGION structs describing
+  all memory protection special regions.
```
apop5: can you add a comment about the allocated memory needing to be freed by caller?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -397,6 +682,47 @@ DumpBitmap (
...
+  for ( ; Index < Count; Index++) {
+    DEBUG ((
+      DEBUG_INFO,
+      "Memory Protection Special Region: 0x%llx - 0x%llx. Attributes: 0x%llx\n",
+      SpecialRegions[Index].Start,
+      SpecialRegions[Index].Start + SpecialRegions[Index].Length,
+      SpecialRegions[Index].EfiAttributes
+      ));
+  }
+
```
apop5, line 717: is there a memory leak here with GetSpecialRegions allocating memory and no free?

MdeModulePkg/Include/Guid/MemoryProtectionSpecialRegion.h
```diff
@@ -0,0 +1,26 @@
+/** @file -- MemoryProtectionSpecialRegion.h
+
+  The required struct definition and GUID for reporting special memory protection
+  regions via the HOB.
+
+  Copyright (c) Microsoft Corporation. All rights reserved.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef __MEMORY_PROTECTION_SPECIAL_REGION_GUID_H__
```
apop5: can the __MEMORY_PROTECTION_SPECIAL_REGION_GUID_H__ get change to follow edk2 coding?
https://edk2-docs.gitbook.io/edk-ii-c-coding-standards-specification/v/release-2.20/5_source_files/53_include_files

MEMORY_PROTECTION_SPECIAL_REGION_GUID_H__

MdeModulePkg/Include/Protocol/MemoryProtectionSpecialRegion.h
```diff
@@ -0,0 +1,42 @@
+/** @file -- MemoryProtectionSpecialRegion.h
+
+  The memory protection special region protocol enables fetching all special regions
+  and adding new ones.
+
+  Copyright (c) Microsoft Corporation. All rights reserved.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef __MEMORY_PROTECTION_SPECIAL_REGION_PROTOCOL_H__
```
apop5: can the __MEMORY_PROTECTION_SPECIAL_REGION_PROTOCOL_H__ get change to follow edk2 coding?
https://edk2-docs.gitbook.io/edk-ii-c-coding-standards-specification/v/release-2.20/5_source_files/53_include_files

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Added a debug message

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: The length field is now a UINT64

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Added a comment in the function header stating that the caller must free the allocated array.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: I think this is related to your request for a comment in the header stating that the allocated array should be freed by the caller. Is that right?

MdeModulePkg/Include/Guid/MemoryProtectionSpecialRegion.h
TaylorBeebe: Updated the ifdef, and renamed the file to conform to the edk2 requirement of no duplicate file names

MdeModulePkg/Include/Protocol/MemoryProtectionSpecialRegion.h
TaylorBeebe: Updated the ifdef, and renamed the file to conform to the edk2 requirement of no duplicate file names

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
apop5: yes.

DumpMemoryProtectionSpecialRegions()
calls GetSpecialRegions () then goes through th results and print info

GetSpecialRegions() allocates the buffer it returns:

SpecialRegions = AllocatePool (sizeof (MEMORY_PROTECTION_SPECIAL_REGION) * mSpecialMemoryRegionsPrivate.Count);

DumpMemoryProtectionSpecialRegions() isn't freeing that buffer

apop5: DumpMemoryProtectionSpecialRegions seems to be an internal debug function, would it be better to just go through the special region list directly and not need to call allocate?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,277 @@ SetUefiImageMemoryAttributes (
...
+  MEMORY_PROTECTION_SPECIAL_REGION_LIST_ENTRY  *SpecialRegionEntry;
+  LIST_ENTRY                                   *SpecialRegionEntryLink;
+  UINTN                                        Index = 0;
+
+  if ((SpecialRegions == NULL) || (*SpecialRegions != NULL) || (Count == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  if (mSpecialMemoryRegionsPrivate.Count == 0) {
+    return EFI_SUCCESS;
```
os-d, line 409: Should we set Count to 0 here?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,277 @@ SetUefiImageMemoryAttributes (
...
+
+  SortMemoryProtectionSpecialRegionList (&mSpecialMemoryRegionsPrivate.SpecialRegionList);
+
+  *SpecialRegions = AllocatePool (sizeof (MEMORY_PROTECTION_SPECIAL_REGION) * mSpecialMemoryRegionsPrivate.Count);
+
+  if (*SpecialRegions == NULL) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  *Count = mSpecialMemoryRegionsPrivate.Count;
```
os-d: nit: This assignment seems unnecessary since `*Count` is not used in between here and line 438 where it is set again.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,284 @@ SetUefiImageMemoryAttributes (
...
+GetSpecialRegions (
+  OUT MEMORY_PROTECTION_SPECIAL_REGION  **SpecialRegions,
+  OUT UINTN                             *Count
+  )
+{
+  MEMORY_PROTECTION_SPECIAL_REGION_LIST_ENTRY  *SpecialRegionEntry;
+  LIST_ENTRY                                   *SpecialRegionEntryLink;
+  UINTN                                        Index = 0;
+
+  if ((SpecialRegions == NULL) || (*SpecialRegions != NULL)) {
```
apop5: Count should be null checked too.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -255,6 +262,277 @@ SetUefiImageMemoryAttributes (
...
+                           );
+
+    CopyMem (
+      &((*SpecialRegions)[Index++]),
+      &SpecialRegionEntry->SpecialRegion,
+      sizeof (MEMORY_PROTECTION_SPECIAL_REGION)
+      );
+  }
+
+  *Count = (Index == mSpecialMemoryRegionsPrivate.Count) ? mSpecialMemoryRegionsPrivate.Count : Index;
```
os-d: In this case, should `mSpecialMemoryRegionsPrivate.Count` be given the actual count?

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: *Count set to 0 in this case

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: removed

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Private count updated if the loop through the list yielded a different count

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Ah I misunderstood where this comment was in the code - this has been fixed :)

apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Added

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: I went ahead and updated this dump to just use the global list instead of the GetSpecialRegions() interface to save an allocation. Thanks for the suggestion!

makubacki (approved):

PR #145, opened 2022-10-25
Add the Interrupt Entry Code to the Special Region List

TaylorBeebe:
## Description

The interrupt entry code should have RO applied, but if attributes are applied based on the memory type then NX would be applied. Adding the range to the special region list will result in RO being applied during memory protection initialization.

## Breaking change?
No

## How This Was Tested

Booting to Windows

## Integration Instructions

N/A

### DIFF
--- a/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeException.c
+++ b/UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeException.c
@@ -183,8 +183,8 @@ InitializeCpuInterruptHandlers (
                   );
   ASSERT (!EFI_ERROR (Status) && InterruptEntryCode != NULL);
 
-  ASSERT_EFI_ERROR (gBS->LocateProtocol (&gMemoryProtectionSpecialRegionProtocolGuid, NULL, (VOID **)&SpecialRegionProtocol));
-
+  Status = gBS->LocateProtocol (&gMemoryProtectionSpecialRegionProtocolGuid, NULL, (VOID **)&SpecialRegionProtocol);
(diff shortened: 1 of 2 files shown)


kenlautner (approved):

apop5 (approved):

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeException.c
```diff
@@ -179,9 +181,23 @@ InitializeCpuInterruptHandlers (
                   EFI_SIZE_TO_PAGES (TemplateMap.ExceptionStubHeaderSize * CPU_INTERRUPT_NUM),
                   (EFI_PHYSICAL_ADDRESS *)(UINTN)&InterruptEntryCode
                   );
-  // MU_CHANGE END
   ASSERT (!EFI_ERROR (Status) && InterruptEntryCode != NULL);
 
+  ASSERT_EFI_ERROR (gBS->LocateProtocol (&gMemoryProtectionSpecialRegionProtocolGuid, NULL, (VOID **)&SpecialRegionProtocol));
```
makubacki: You'll have to move this code outside the assert.

Might not execute if debug asserts are disabled.

makubacki (changes requested):
I'll revisit after the assert code is updated. The rest looks good.

UefiCpuPkg/Library/CpuExceptionHandlerLib/DxeException.c
TaylorBeebe: Ah yes my mistake

makubacki (approved):

PR #146, opened 2022-10-26
Update Special Region Logic to Handle Broken Cases

TaylorBeebe:
## Description

This PR fixes some cases which could prevent an accurate split of the special regions in the memory map:

1. We need to ensure the insertion address hasn't reached the end of the memory map, not that the buffer size is larger than the address itself.
2. The special region list needs to be sorted. It was sorted automatically when calling GetSpecialRegions() and would have needed to also be sorted when calling or before calling SeparateSpecialRegionsInMemoryMap() because the list is passed in directly. To simplify things, we do an ordered insert whenever we add a special region so the list is always sorted and removed the old sorting function.
3. When walking through the memory map to see if some records need to be split, if the memory map entry covers MORE than the special region range, we need to trim the map entry to not include that special region and insert the special region descriptor at the end of the memory map in case a subsequent special region contains some section of the remainder of that map entry range. Because the special region list is sorted, we don't need to do this in the case that the start of the map entry described region is before the start of the special region.

## Breaking change?
No

## How This Was Tested

Testing different special region configurations

## Integration Instructions

N/A


MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -265,74 +265,290 @@ SetUefiImageMemoryAttributes (
...
-    SpecialRegionEntryLink     = SpecialRegionEntryLink->ForwardLink;
-    NextSpecialRegionEntryLink = SpecialRegionEntryLink->ForwardLink;
+    CodeSectionLink = CodeSectionLink->ForwardLink;
   }
+
+  CodeSectionToInsertLink->BackLink              = CodeSectionLink->BackLink;
+  CodeSectionToInsertLink->ForwardLink           = CodeSectionLink;
+  CodeSectionToInsertLink->BackLink->ForwardLink = CodeSectionToInsertLink;
+  CodeSectionToInsertLink->ForwardLink->BackLink = CodeSectionToInsertLink;
+  return EFI_SUCCESS;
```
os-d: This appears to be the same function three times for the different types. Does it make sense to make a generic function to call? It gets a little bit uglier with having to pass the type in, but if the types are similar enough in the parts that are getting rearranged a simple abstract struct could be used?

If it's too difficult or I'm missing there are differences in these ordered insert functions, then we can disregard.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -265,74 +265,290 @@ SetUefiImageMemoryAttributes (
...
+
+/**
+  Merges every LIST_ENTRY within ArrayOfListEntriesToBeMerged into ImagePropertiesRecordList
+
+  @param[in]  ListToMergeInto                 Pointer to the head of a list of IMAGE_PROPERTIES_RECORD entries
+                                              into which the input ListToBeMerged will be merged
+  @param[in]  ListToBeMerged                  Pointer to the head of a list of IMAGE_PROPERTIES_RECORD entries
+                                              which will be merged into ListToMergeInto
+  @param[in]  ListToBeMergedCount             Number of IMAGE_PROPERTIES_RECORD entries in ListToBeMerged
+  @param[out] ArrayOfListEntriesToBeMerged    Pointer to an allocated array of LIST_ENTRY* which were merged
```
os-d: nit: `Pointer to a caller allocated array`

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -265,74 +265,290 @@ SetUefiImageMemoryAttributes (
...
+    Status = OrderedInsertImageRecordListEntry (ListToMergeInto, ArrayOfMergedElements[Index++]);
+    if (EFI_ERROR (Status)) {
+      break;
+    }
+  }
+
+  // If we did not merge all elements of the list, unmerge them and free the input array
+  if (!IsListEmpty (ListToBeMerged)) {
+    OrderedInsertImagePropertiesRecordArray (ListToBeMerged, ArrayOfMergedElements, Index - 1);
+    FreePool (*ArrayOfMergedElements);
```
os-d: Do we want to free ArrayOfMergedElements in the failure case? It seems odd to me to free a caller allocated buffer in this case. I.e. from the caller perspective, I would assume I allocate my buffer, pass it in, check status, if status failed, free all my memory I allocated.

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: I agree, but this PR should be scoped to just the special range updates (the only reason this is showing up as a diff is because I needed to move this logic to avoid forward declaration). I'll create a PR after this goes in to change the interface.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Will update in a subsequent PR

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Hmmmm.... I think maybe if the function only compares UINT64 values, then I can pass in the offset of the value being compared, the offset of that expected signature, and the expected signature itself to create an insert function.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -265,74 +265,290 @@ SetUefiImageMemoryAttributes (
...
+
+  @retval EFI_SUCCESS                    MEMORY_PROTECTION_SPECIAL_REGION_LIST_ENTRY inserted into the list
+  @retval EFI_INVALID_PARAMETER          CodeSectionList or CodeSectionToInsertLink were NULL
 **/
 STATIC
-VOID
-SortMemoryProtectionSpecialRegionList (
-  IN LIST_ENTRY  *SpecialRegionList
+EFI_STATUS
+OrderedInsertSpecialRegionListEntry (
```
apop5: does this need to handle the case of the SpecialRegionToInsertLink having overlap with the existing list?  Right now it will just create a new entry even if there is overlap.

TaylorBeebe: For the linked list function itself. I don't think it needs to handle the scenario you describe.

For the logic that's actually adding the region entry, it might be a good redundancy measure but it's not immediately obvious what the best course of action is when there is overlap. Maybe adding a remove() function to the protocol is a good idea so the caller of add() can remove the overlapping entry before calling add() again, or maybe the best way of resolving it would be to just have whatever is being inserted "overwrite" the existing special entry region which might necessitate one special region list entry becoming three.

apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Yeah it works - I'll consolidate them

TaylorBeebe: Please review: https://github.com/microsoft/mu_basecore/pull/148/files

TaylorBeebe:
Rebased on the most recent 202202 so there are fewer diffs in this PR. Requesting a re-review, thanks :)

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2023,30 +1969,38 @@ SeparateSpecialRegionsInMemoryMap (
...
+          // SpecialRegionStart is now guaranteed to be equal to MapEntryStart. Populate a new descriptor
+          // for the region covered by the special region. This entry needs to go to
+          // the end of the memory map in case a subsequent special region will cover some portion
+          // of the remaining map entry region
           POPULATE_MEMORY_DESCRIPTOR_ENTRY (
             MapEntryInsert,
-            SpecialRegionEnd,
-            EFI_SIZE_TO_PAGES (MapEntryEnd - SpecialRegionEnd),
+            SpecialRegionStart,
+            EFI_SIZE_TO_PAGES (SpecialRegionEnd - SpecialRegionStart),
```
os-d, line 1979: I'm slightly confused by the comment vs code here. The comment says this goes to the end of the memory map, in the code it goes to the end of the special region. Is that guaranteed to be the end of the memory map? Or am I misinterpreting?

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: It is a bit confusing. In this context, "end of the memory map" means "the next available blank EFI_MEMORY_DESCRIPTOR at the end of the memory map."

Explained another way: If I find that the memory range described by the special region ends before the end of the memory range described by the memory map descriptor I'm currently inspecting, then I want to create a new memory map descriptor that describes the special region and put that descriptor at the end of the map. Then, I trim the memory map entry I'm currently inspecting to just the remaining region uncovered by that special region in case a subsequent special region is also covered by that remaining region.

This is one of those things that's easier with a visual representation, but my ASCII art is subpar :D

apop5 (approved):

PR #148, opened 2022-10-26
Generalize List Functions in MemoryProtectionSupport

TaylorBeebe:
## Description

Instead of making multiple ordered insert functions for each type containing the input list entry pointer, inserts can be done generically by the type being compared and the offsets of the relevant signature and value being compared. This PR updates the OrderedInsert, MergeList, and OrderedInsertArray functions to be agnostic to the struct type containing the list entry provided the struct contains at least the value which is being compared.

This PR also updates the interface of what was MergeImagePropertiesRecordLists() and is now MergeListsUint64Comparison() to be passed a triple pointer so the array can be allocated within the function instead of requiring the allocation to occur before the function call.

## Breaking change?
No

## How This Was Tested

Comparing the function output against the previous behavior

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -265,6 +265,229 @@ SetUefiImageMemoryAttributes (
 extern LIST_ENTRY  mGcdMemorySpaceMap;
 
 // ---------------------------------------
+//     LINKED LIST SUPPORT FUNCTIONS
+// ---------------------------------------
+
+/**
+  Inserts the input EntryToInsert into List by comparing UINT64 values at LIST_ENTRY + ComparisonOffset. If the input
+  Signature is non-zero, a signature check based on each LIST_ENTRY + SignatureOffset will be performed.
+
+  @param[in] List                       Pointer to the head of the list into which EntryToInsert will be inserted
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -264,6 +264,229 @@ SetUefiImageMemoryAttributes (
...
+                                              ListToBeMerged will be merged
+  @param[in]  ListToBeMerged                  Pointer to the head of a list which will be merged
+                                              into ListToMergeInto
+  @param[in]  ListToBeMergedCount             Number of LIST_ENTRY* in ListToBeMerged
+  @param[out] ArrayOfMergedElements           Pointer to an unallocated array of LIST_ENTRY*. The array will be
+                                              allocated if the function returns success and contain every
+                                              LIST_ENTRY* merged into ListToMergeInto
+  @param[in] ComparisonOffset                 Offset of the field to compare each list entry against relative to the
+                                              list entry pointer
+  @param[in] ComparisonOffset                 Offset of the signature to compare each list entry against relative to the
```
os-d: nit: SignatureOffset

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c

kenlautner (approved):

PR #149, opened 2022-10-27
Check Page Table on Protection Policy Application

TaylorBeebe:
## Description

This update checks the page table before calling SetMemoryAttributes() to avoid a TLB and cache flush if it's unnecessary. A subtle and unresolved bug can cause the pages being allocated/freed to not have consistent attributes which can cause unpredictable issues. Instead of calling SetMemoryAttributes() on every allocate and free, check the page table to see if the attributes across the region are consistent and the call to SetMemoryAttributes() is necessary. This update will reap similar performance benefits to the same call we added in CoreInternalFreePages() because the number of times where we actually need to call SetMemoryAttributes() is small and branch prediction will expedite the most common result.

## Breaking change?

No

## How This Was Tested

Booting to shell

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -1482,8 +1482,10 @@ ApplyMemoryProtectionPolicy (
   IN  UINT64                Length
   )
 {
-  // UINT64  OldAttributes;
+  UINT64  OldAttributes;
   UINT64  NewAttributes;
+
+  EFI_STATUS  Status; // MU_CHANGE
 
   //
   // The policy configured in Dxe NX Protection Policy // MU_CHANGE
@@ -1541,7 +1543,7 @@ ApplyMemoryProtectionPolicy (
   //
   NewAttributes = GetPermissionAttributeForMemoryType (NewType);
 
-  // MU_CHANGE START: TODO: There is a potential bug where attributes are not properly set
+  // MU_CHANGE START: There is a potential bug where attributes are not properly set
   //                  for all pages during a call to AllocatePages(). This may be due to a bug somewhere
   //                  during the free page process.
   // if (OldType != EfiMaxMemoryType) {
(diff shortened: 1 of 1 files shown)


os-d (approved):

kenlautner (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1554,6 +1554,18 @@ ApplyMemoryProtectionPolicy (
   //   // newly added region of a type that does not require protection
   //   return EFI_SUCCESS;
   // }
+
+  // To catch the edge case where the attributes are not consistent across the range, get the
+  // attributes from the page table to see if they are consistent. If they are not consistent,
+  // GetMemoryAttributes() will return an error.
+  if (MemoryAttributeProtocol != NULL) {
+    if (!EFI_ERROR (MemoryAttributeProtocol->GetMemoryAttributes (MemoryAttributeProtocol, Memory, Length, &OldAttributes)) &&
```
apop5, line 1562: Nit: can you move the call to MemoryAttributeProtocol out of the if statement?

You are saving a EFI_STATUS local variable, but readability is better having it outside.

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1554,6 +1554,18 @@ ApplyMemoryProtectionPolicy (
   //   // newly added region of a type that does not require protection
   //   return EFI_SUCCESS;
   // }
+
+  // To catch the edge case where the attributes are not consistent across the range, get the
+  // attributes from the page table to see if they are consistent. If they are not consistent,
+  // GetMemoryAttributes() will return an error.
+  if (MemoryAttributeProtocol != NULL) {
```
apop5, line 1561: nit:
MemoryAttributeProtocol 

This is a global variable defined in ? Should it be mMemoryAttributeProtocol ?

TaylorBeebe: It's declared in DxeMain.h, and I think that means it should have the 'g' prefix convention. I can post a separate PR to change all references to gMemoryAttributeProtocol if that sounds right.

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
TaylorBeebe: For upstream deviations I usually try to optimize for a smaller footprint, and I still think this is quite readable. Thoughts?

apop5 (approved):

PR #151, opened 2022-10-31
Handle Special Region Overlap

TaylorBeebe:
## Description

This update handles the case where a caller tries to add a special region which is partially or completely already present in the special region list. The overlapping special region will be allowed into the list if it does not conflict with the existing special regions. A special region conflicts if it overlaps with an existing interval and its attributes are not a subset of the attributes of the existing interval.

## Breaking change?
No

## How This Was Tested

Testing various interval combinations

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -60,7 +60,7 @@ EFI_MEMORY_ATTRIBUTE_PROTOCOL  *MemoryAttributeProtocol = NULL;
 #define CHECK_SUBSUMPTION(AStart, AEnd, BStart, BEnd) \
   ((AStart < BStart) && (AEnd > BEnd))
 
-// TRUE if A is a subset of B
+// TRUE if A is a bitwise subset of B
 #define CHECK_SUBSET(A, B)  ((A | B) == B)
 
 /**
@@ -555,8 +555,10 @@ MergeOverlappingSpecialRegions (
         // If the attributes are the same between both entries, just expand BackEntry and delete ForwardEntry
         if (BackEntry->SpecialRegion.EfiAttributes == ForwardEntry->SpecialRegion.EfiAttributes) {
           BackEntry->SpecialRegion.Length = ForwardEnd - BackStart;
+          ForwardLink                     = ForwardLink->ForwardLink;
           RemoveEntryList (ForwardLink);
           FreePool (ForwardEntry);
+          continue;
         }
         // If BackEntry subsumes ForwardEntry, we need to create a new list entry to split BackEntry
         else if (CHECK_SUBSUMPTION (BackStart, BackEnd, ForwardStart, ForwardEnd)) {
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -51,6 +51,18 @@ EFI_MEMORY_ATTRIBUTE_PROTOCOL  *MemoryAttributeProtocol = NULL;
...
 #define SPECIAL_REGION_PATTERN  7732426 // 7-(S) 7-(P) 3-(E) 2-(C) 4-(I) 2-(A) 6-(L)
 
+// TRUE if A and B have overlapping intervals
+#define CHECK_OVERLAP(AStart, AEnd, BStart, BEnd) \
+  ((AStart <= BStart && AEnd > BStart) || \
+  (BStart <= AStart && BEnd > AStart))
+
+// TRUE if A interval subsumes B interval
+#define CHECK_SUBSUMPTION(AStart, AEnd, BStart, BEnd) \
+  ((AStart < BStart) && (AEnd > BEnd))
```
apop5, line 61: double check my logic here, but doens't Aend and Bend need to be >= instead of just >?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -51,6 +51,18 @@ EFI_MEMORY_ATTRIBUTE_PROTOCOL  *MemoryAttributeProtocol = NULL;
...
+
+// TRUE if A is a subset of B
```
apop5: Nit: Can you add a comment that this macro is comparing EFI_MEMORY_* bitwise attributes?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -502,15 +672,30 @@ CollectSpecialRegionHobs (
...
+        "%a - Special region 0x%llx - 0x%llx conflicts with another special region!\n",
+        __FUNCTION__,
+        HobSpecialRegion->Start,
+        HobSpecialRegion->Start + HobSpecialRegion->Length
+        ));
+      ASSERT (FALSE);
+      GuidHob = GetNextGuidHob (&gMemoryProtectionSpecialRegionHobGuid, GET_NEXT_HOB (GuidHob));
+      continue;
+    }
+
```
apop5: nit style: I think the continue can be removed and everything can just be an else statement?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Technically true, but if AEnd == BEnd, we don't need to create a new special region entry to cover BEnd->AEnd which is how this macro is used.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: I'll put bitwise subset because this is just the generic definition of a bitwise subset

os-d (approved):

apop5 (approved):

apop5 (approved):

PR #152, opened 2022-11-01
Activate Improved Memory Protection Init Functionality

TaylorBeebe:
## Description

This update activates all the added memory protection logic to ensure when the CPU architecture protocol is installed, we can create a clear understanding of the state of memory and maximize paging protections based on the memory protection policy for the boot.

## Breaking change?
Yes

## How This Was Tested

Booting to OS on Q35 using various memory protection configuration presets.

## Integration Instructions

Any platform which requires special permissions for a particular memory region will need to add a memory protection special region by either building a memory protection special region HOB in PEI or utilizing the memory protection special region protocol in DXE before the CPU architecture protocol is installed.

Here are some cases where you might want to create a special region:

1. You're using FSP API mode and need to thunk back into a 32-bit image which may have NX applied based on its memory type
2. You allocated memory in DXE before the CPU architecture protocol was installed and need that memory to have particular attributes not in line with the attributes which would be applied based on its memory type.
3. There is an executing region of memory not loaded using the PE loader which cannot have its access attributes updated without disabling paging.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -2518,7 +2518,7 @@ GetMemoryMapWithPopulatedAccessAttributes (
   EFI_STATUS  Status;
   UINTN       AdditionalRecordCount, NumMemoryMapDescriptors, NumBitmapEntries, \
               NumMemorySpaceMapDescriptors, MemorySpaceMapDescriptorSize, MapKey, \
-              ExpandedMemoryMapSize, MemorySpaceMapSize, BitmapIndex;
+              ExpandedMemoryMapSize, BitmapIndex;
   UINT32                           DescriptorVersion;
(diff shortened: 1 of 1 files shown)


os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2506,8 +2504,8 @@ RemoveAttributesOfNonProtectedImageRanges (
                                           EFI_MEMORY_DESCRIPTOR
 
   @retval         EFI_SUCCESS             The memory map was returned in the MemoryMap buffer
-  @retval         EFI_INVALID_PARAMETER   One of the input parameters has an invalid value
   @retval         EFI_OUT_OF_RESOURCES    Failed to allocate memory
+  @retval         Other                   One of the supportion functions failed
```
kenlautner: Typo.  I assume you mean supporting.

kenlautner (approved):

PR #157, opened 2022-11-02
Add Stack Cookie Failure to ExceptionPersistenceLib

TaylorBeebe:
## Description

Add ExceptionPersistStackCookie to the ExceptionPersistenceLib enum to enable checking for a stack cookie check failure interrupt between boots.

## Breaking change?
No

## How This Was Tested

Enabling Stack Cookies

## Integration Instructions

N/A


Flickdm (approved):

makubacki:
Can you provide more information in the commit description about the context around int 2C?

TaylorBeebe:
> Can you provide more information in the commit description about the context around int 2C?

I think @spbrogan can provide context on this. The decision to use 2C was made years back by him and Jeremiah.

makubacki:
> > Can you provide more information in the commit description about the context around int 2C?
> 
> I think @spbrogan can provide context on this. The decision to use 2C was made years back by him and Jeremiah.

Okay, thanks. It's one of those things that's better to capture in the codebase while we still have the info.

makubacki:
The "I2C" terminology is a bit vague and confusing with the I2C bus. Did you intentionally rule out calling it something like `ExceptionPersistStackCookie`?

TaylorBeebe:
> > > Can you provide more information in the commit description about the context around int 2C?
> > 
> > 
> > I think @spbrogan can provide context on this. The decision to use 2C was made years back by him and Jeremiah.
> 
> Okay, thanks. It's one of those things that's better to capture in the codebase while we still have the info.

I spoke with Sean about this. The reasoning behind 2C was because it's typically used to invoke the debugger in Windows and we would have handled stack cookies by setting the debug flag. Now that we can hook the stack cookie failure into our exception persistence library, using 2C is no longer necessary. I'll change this to some other unused interrupt value.

TaylorBeebe:
> The "I2C" terminology is a bit vague and confusing with the I2C bus. Did you intentionally rule out calling it something like `ExceptionPersistStackCookie`?

Updated to ExceptionPersistStackCookie

makubacki:
> > > > Can you provide more information in the commit description about the context around int 2C?
> > > 
> > > 
> > > I think @spbrogan can provide context on this. The decision to use 2C was made years back by him and Jeremiah.
> > 
> > 
> > Okay, thanks. It's one of those things that's better to capture in the codebase while we still have the info.
> 
> I spoke with Sean about this. The reasoning behind 2C was because it's typically used to invoke the debugger in Windows and we would have handled stack cookies by setting the debug flag. Now that we can hook the stack cookie failure into our exception persistence library, using 2C is no longer necessary. I'll change this to some other unused interrupt value.

Thanks for following up

makubacki (approved):

PR #159, opened 2022-11-02
Update __security_check_cookie_Failure to Emit 0x40 Interrupt

TaylorBeebe:
## Description

Stack Cookie failures will cause a 0x40 interrupt. The intel software development manual states interrupts 0-32 are predefined and our use of interrupt 64 is an arbitrary choice.

## Breaking change?

No

## How This Was Tested

Enabling stack cookies on Q35

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Library/BaseBinSecurityLibRng/X64/security_check_cookie.asm
+++ b/MdePkg/Library/BaseBinSecurityLibRng/X64/security_check_cookie.asm
@@ -17,7 +17,7 @@ __security_check_cookie PROC PUBLIC
     ret
 
 __security_check_cookie_Failure:
-    int     3
+    int     2ch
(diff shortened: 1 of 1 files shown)


Flickdm (approved):

makubacki:
The commit message still describes the interrupt as 0x2C.

Can you also add the rationale for 0x40 to the commit message? There's nothing wrong with the value. Just so people in the future know whether it was arbitrary or based on some criteria.

makubacki (changes requested):
Would like to see a couple quick commit message updates.

TaylorBeebe:
> Would like to see a couple quick commit message updates.

updated

makubacki (approved):

PR #165, opened 2022-11-10
Update Stack Cookie Vector to Fixed PCD

TaylorBeebe:
## Description

To enable more easily setting the stack cookie failure vector, update the check to reference a fixed at build PCD in MdePkg. The interrupt vector value default is now 0x42 instead of 0x40 to avoid conflicting with the HPET interrupt vector.

## Breaking change?

No

## How This Was Tested

Triggering the interrupt on Q35

## Integration Instructions

Platforms can update the value by setting the PcdStackCookieExceptionVector PCD in their platform DSC file.

### DIFF
--- a/MdePkg/Library/BaseBinSecurityLibRng/BaseBinSecurityLibRng.inf
+++ b/MdePkg/Library/BaseBinSecurityLibRng/BaseBinSecurityLibRng.inf
@@ -32,4 +32,7 @@
   DebugLib
   BaseLib
   RngLib
+  PcdLib
 
+[FixedPcd]
+  gEfiMdePkgTokenSpaceGuid.PcdStackCookieExceptionVector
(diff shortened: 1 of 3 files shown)


makubacki (approved):

makubacki:
It's probably worth calling out in the description (or integration section) that you moved the int to 0x42 from 0x40. The integration section could mention that platforms can update the PCD in their platform build files (obvious I know, but some people don't understand the edk build environment and just want to know how to solve their problem).

TaylorBeebe:
> It's probably worth calling out in the description (or integration section) that you moved the int to 0x42 from 0x40. The integration section could mention that platforms can update the PCD in their platform build files (obvious I know, but some people don't understand the edk build environment and just want to know how to solve their problem).

Updated

kenlautner (approved):

PR #166, opened 2022-11-10
Personal/klautner/pei memory buckets 202202

kenlautner:
## Description
Added memory buckets to PEI for all memory types that persist into runtime.  This is reduce memory fragmentation for memory that is passed into the OS.  Currently only EfiRuntimeServicesData is allocated in PEI.

- [x] Breaking change?
  Yes.  A platform using MU_BASECORE will need to add the MemoryBucketLib and MemoryBucketHelperLib libaries to their DSC files or they won't be able to build with this change.

## How This Was Tested

I was able to build the firmware for TGL and ADL systems and saw that the different EfiRuntimeServicesData allocations were consolidated in memory.

## Integration Instructions

Add MemoryBucketLib and MemoryBucketHelperLib to your platform DSC file.


kenlautner:
Do people think it makes sense to have the memory bucket sizes set with PCDs?

kenlautner:
Closing as backporting this change is unnecessary.

PR #167, opened 2022-11-10
Add support for PEI memory bucket allocation

kenlautner:
## Description

Added memory buckets to PEI for all memory types that persist into runtime. This is to reduce memory fragmentation for memory that is passed into the OS. Currently only EfiRuntimeServicesData is allocated in PEI.

MemoryBucketLib is added to keep track of PEI memory allocations of relevant types (currently only memory that persists into runtime) and consolidate their allocations to reduce fragmentation.

MemoryBucketLib keeps track of all the relevant information about the PEI memory buckets.  It considers the newly added PeiMemoryBucket PCDs to decide what memory allocations to keep track of and possibly not be used at all.   If all the PCDs remain at their default value of 0 PEI memory buckets will not be used and PEI memory allocation will function the same as it has before this change.  Otherwise PEI memory buckets will be created for the memory types with non-zero PCD values. 

These buckets are created in post-mem PEI and there is currently no support for pre-mem buckets.

- [x] Breaking change?
  Yes. A platform using MU_BASECORE will need to add the MemoryBucketLib and MemoryBucketHelperLib libaries to their DSC files or they won't be able to build with this change.

## How This Was Tested

I was able to build the firmware for TGL and ADL systems and saw that the different EfiRuntimeServicesData allocations were consolidated in memory.

## Integration Instructions

Add MemoryBucketLib and MemoryBucketHelperLib to your platform DSC file.  Additionally change the values for the added PCDs for relevant memory types.


makubacki:
Please give me an opportunity to review this before it is submitted. Should be able to by Friday.

makubacki:
> Please give me an opportunity to review this before it is submitted. Should be able to by Friday.

I see the branch is passing build. I'll take a look tomorrow.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,332 @@
+/**@file
+Library the defines Runtime memory buckets for use during PEI
```
makubacki:
```suggestion
Library that defines Runtime memory buckets for use during PEI
```

makubacki:
The PR needs more high-level information either in the PR description or a file in the repo to describe the changes.

- What is the overall design? In other words, how are these changes supposed to work at a high-level?
- Why would a platform want to use this? Do they always need it? Or under certain circumstances?
- Why is MemoryBucketHelperLib split out from MemoryBucketLib?

I know some of this can be gleaned from the code changes or is clear in your mind but adding this will help confirm the changes in the code match the design expectations and help platforms integrate the change since this is a change primarily meant for platform integration (and required due to the library dependency introduced).

makubacki:
Some of the changes are missing `MU_CHANGE` tags.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,35 @@
+/** @file
+Guid for Pei storing and retreiving the PEI memory buckets.
```
makubacki:
```suggestion
Guid for Pei storing and retrieving the PEI memory buckets.
```

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,35 @@
...
+
+#ifndef _MEMORY_BUCKET_INFORMATION_H_
+#define _MEMORY_BUCKET_INFORMATION_H_
+
+#define MEMORY_BUCKET_INFORMATION_GUID \
+{ 0x36138737, 0xb6db, 0x4ed6, { 0x9b, 0x4b, 0xf9, 0x61, 0x28, 0xe7, 0x19, 0x3c } }
+
+extern EFI_GUID  gMemoryBucketInformationGuid;
+
+#define EFI_MEMORY_TYPE_STATISTICS_GUID \
```
makubacki: Is this from the EFI spec?

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,35 @@
...
+extern EFI_GUID  gEfiMemoryTypeStatisticsGuid;
+
+extern UINTN  NumberOfBuckets;
+
+typedef struct {
+  EFI_PHYSICAL_ADDRESS    BaseAddress;
+  EFI_PHYSICAL_ADDRESS    MaximumAddress;
+  UINT64                  CurrentNumberOfPages;
+  UINT64                  NumberOfPages;
+  UINTN                   InformationIndex;
```
makubacki: I recommend trying to avoid using `UINTN` in structures since the field size can be intrepreted differently across build boundaries.

For example, a struct generated in 32-bit PEI being read in 64-bit DXE.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
makubacki: If it is something locally invented just drop the `EFI_` prefix.

MdeModulePkg/Include/Library/MemoryBucketHelperLib.h
```diff
@@ -0,0 +1,198 @@
...
+  This function initialized the PEI memory buckets.  This should be called
+  when the first memory type being tracked in these buckets is allocated.
+
+  @param[in] StartingAddress   The starting address of the memory buckets.
+                               All of them are contiguous.
+**/
+VOID
+EFIAPI
+InitializeMemoryBuckets (
+  EFI_PHYSICAL_ADDRESS  StartingAddress
```
makubacki: Please add the `IN`, `OUT`, etc modifiers. This is especially useful for functions in header files.

MdeModulePkg/Include/Library/MemoryBucketLib.h
```diff
@@ -0,0 +1,130 @@
...
+  @param[in] MemoryType         The type of memory allocated by this HOB.
+
+**/
+VOID
+InternalBuildRuntimeMemoryAllocationHob (
+  IN EFI_PHYSICAL_ADDRESS  BaseAddress,
+  IN UINT64                Length,
+  IN EFI_MEMORY_TYPE       MemoryType
+  );
+
```
makubacki: Generally, internal functions should not be in a library external interface.

Can you give a bit more info about why some of these functions are here?

MdeModulePkg/Include/Library/MemoryBucketLib.h
```diff
@@ -0,0 +1,130 @@
+/** @file
```
makubacki, line 1: It's a little unclear who is expected to call each function in this file.

For each function, are they expected to be called by (1) PeiMain (2) the platform (3) others/both?

Thinking from the perspective of a platform integrating this change, what do they need to do with these functions?

makubacki:
I'll look at the C files after we can clarify the library interface and integration model mentioned in other comments.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: Fixed.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
kenlautner: Fixed.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
kenlautner: Changed to not include EFI.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
kenlautner: Updated to use UINT32.

MdeModulePkg/Include/Library/MemoryBucketHelperLib.h
kenlautner: Updated all new functions to include IN/OUT.

MdeModulePkg/Include/Library/MemoryBucketLib.h
kenlautner: These functions are taken directly from PEI memory services where these are used for internal hob memory allocations.  I end up needing them for the same reason.  I can move these functions to MemoryBucketHelperLib to not make them forward facing.  Another option is that I can consolidate MemoryBucketLib with MemoryServices if that makes more sense.  I didn't want to change one of the core services too much but it might be a better path.

MdeModulePkg/Include/Library/MemoryBucketLib.h
kenlautner: Currently everything is only expected to be called by PEI memory services unless a platform wants to touch the memory allocation itself.  As I mentioned above I could consolidate MemoryBucketLib with MemoryServices to reduce confusion.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,35 @@
+/** @file
+Guid for Pei storing and retrieving the PEI memory buckets.
+**/
```
cfernald, line 9: copyright and license

MdeModulePkg/Core/Pei/Memory/MemoryServices.c
```diff
@@ -650,6 +658,15 @@ PeiAllocatePages (
     FreeMemoryBottom = &(Hob.HandoffInformationTable->EfiFreeMemoryBottom);
   }
 
+  // MU_CHANGE START
+  // Check to make sure we aren't allocating memory in runtime buckets
+  SyncMemoryBuckets ();
+  if (CheckIfInRuntimeBoundary (*FreeMemoryTop)) {
```
cfernald: Wouldn't this also require the number of pages to properly check for overlap?

MdeModulePkg/Include/Library/MemoryBucketHelperLib.h
```diff
@@ -0,0 +1,198 @@
...
+                          structure.
+**/
+VOID
+EFIAPI
+UpdateRuntimeMemoryStats (
+  IN UINTN            Pages,
+  IN EFI_MEMORY_TYPE  MemoryType
+  );
+
+VOID
```
cfernald: Missing function comment

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.inf
```diff
@@ -0,0 +1,30 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = MemoryBucketHelperLib
+  FILE_GUID                      = 916b6cbe-52f3-439b-8f4c-b97453e36cc7
+  MODULE_TYPE                    = PEIM
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = MemoryBucketHelperLib|DXE_CORE DXE_DRIVER DXE_RUNTIME_DRIVER DXE_SMM_DRIVER SMM_CORE PEIM SEC PEI_CORE UEFI_APPLICATION UEFI_DRIVER
```
cfernald: Unless I misunderstand the intent. This will only support PEI_CORE, right?

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.c
```diff
@@ -0,0 +1,353 @@
+/**@file
```
cfernald: What is the benefit of having the separate helper library? It seems just having separate file in MemoryBucketLib and a private header function in the MemoryBucketLib directory for the internal function would make integration easier and be enforced what is internal vs external. Sorry if I missed something explaining this.

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.c
```diff
@@ -0,0 +1,353 @@
...
+
+EFI_PHYSICAL_ADDRESS  CurrentBucketTops[4] = {
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesCode
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesData
+  MAX_ALLOC_ADDRESS, // EfiACPIReclaimMemory
+  MAX_ALLOC_ADDRESS  // EfiACPIMemoryNVS
+};
+
+// Total number of memory buckets
+UINTN  NumberOfBuckets = 4;
```
cfernald: might be better to acquire programmatically to avoid someone forgetting to update this.

e.g.
```
NumberOfBuckets = sizeof(MemoryTypes)/sizeof(MemoryTypes[0]);
```

Additionally, this could probably be a macro or `const`.

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.c
```diff
@@ -0,0 +1,353 @@
...
+
+#include <PiPei.h>
+
+#include <Uefi.h>
+#include <Library/DebugLib.h>
+#include <Library/MemoryBucketHelperLib.h>
+#include <Library/BaseMemoryLib.h>
+
+// Memory types being kept in buckets in PEI.  Currently only runtime types.
+EFI_MEMORY_TYPE  MemoryTypes[4] = {
```
cfernald: consider making this `const`

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,35 @@
...
+{ 0x36138737, 0xb6db, 0x4ed6, { 0x9b, 0x4b, 0xf9, 0x61, 0x28, 0xe7, 0x19, 0x3c } }
+
+extern EFI_GUID  gMemoryBucketInformationGuid;
+
+#define MEMORY_TYPE_STATISTICS_GUID \
+  { 0x6146C0D6, 0x8E30, 0x4DC2, { 0xA9, 0xCB, 0x5D, 0x85, 0x10, 0xC4, 0x8B, 0x39 }}
+
+extern EFI_GUID  gMemoryTypeStatisticsGuid;
+
+extern UINTN  NumberOfBuckets;
```
cfernald: I don't currently see this being used anywhere outside of the helper lib. Can this be removed? If not consider a comment here explaining its intended use.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,35 @@
...
+  UINT64                  NumberOfPages;
+  UINT32                  InformationIndex;
+  BOOLEAN                 Special;
+  BOOLEAN                 Runtime;
+} EFI_MEMORY_TYPE_STATISTICS;
+
+typedef struct {
+  EFI_MEMORY_TYPE_STATISTICS    RuntimeBuckets[4];
+  EFI_PHYSICAL_ADDRESS          CurrentTopInBucket[4];
+} PEI_MEMORY_BUCKET_INFORMATION;
```
cfernald, line 28: similar to as above, I only see this  being used in the memory bucker libraries, perhaps the helper lib header would be more appropriate.

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.c
```diff
@@ -0,0 +1,353 @@
...
+};
+
+// Total number of memory buckets
+UINTN  NumberOfBuckets = 4;
+
+// Information storage for Hob
+PEI_MEMORY_BUCKET_INFORMATION  RuntimeBucketHob;
+
+// Number of pages that can fit into each memory bucket.
+UINT32  RuntimeMemLength = 50;
```
cfernald: Should this be a PCD?

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.c
```diff
@@ -0,0 +1,353 @@
...
+  This function initialized the PEI memory buckets.  This should be called
+  when the first memory type being tracked in these buckets is allocated.
+
+  @param[in] StartingAddress   The starting address of the memory buckets.
+                               All of them are contiguous.
+**/
+VOID
+EFIAPI
+InitializeMemoryBuckets (
+  EFI_PHYSICAL_ADDRESS  StartingAddress
```
cfernald: Here and elsewhere in this file, missing the IN/OUT tags

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.c
```diff
@@ -0,0 +1,353 @@
...
+  MemoryType.
+
+  @param[in] MemoryType   The type of memory we are interested in.
+
+  @retval    Returns an address that points to the end of memory in a
+             memory bucket that is specified by MemoryType
+**/
+EFI_PHYSICAL_ADDRESS
+EFIAPI
+GetCurrentBucketEnd (
```
cfernald: a bit nitpicky, but I find the mix of end and top a bit confusing. Perhaps bottom would be more consistent.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,332 @@
...
+/**
+  Function that returns the address associated with the end of the
+  memory bucket structure.
+
+  @retval    The memory address below the memory bucket structure.
+
+**/
+EFI_PHYSICAL_ADDRESS
+EFIAPI
+GetEndOfBucketsAddress (
```
cfernald: Consider "Bottom" here too for consistency

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,332 @@
...
+    //
+    return EFI_NOT_AVAILABLE_YET;
+  }
+
+  if (RUNTIME_PAGE_ALLOCATION_GRANULARITY > DEFAULT_PAGE_ALLOCATION_GRANULARITY) {
+    Granularity = RUNTIME_PAGE_ALLOCATION_GRANULARITY;
+  }
+
+  MemBucketHob = GetFirstGuidHob (&gMemoryBucketInformationGuid);
+  SyncMemoryBuckets ();
```
cfernald: consider a comment explaining this multi-stage initialization in detail to improve readability

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,332 @@
...
+  } else if (!PrivateData->PeiMemoryInstalled && PrivateData->SwitchStackSignal) {
+    //
+    // When PeiInstallMemory is called but temporary memory has *not* been moved to permanent memory,
+    // the AllocatePage will depend on the field of PEI_CORE_INSTANCE structure.
+    //
+    InitializeMemoryBuckets ((EFI_PHYSICAL_ADDRESS)(PrivateData->FreePhysicalMemoryTop));
+    FreeMemoryTop    = GetCurrentBucketTop (MemoryType);
+    FreeMemoryBottom = GetCurrentBucketEnd (MemoryType);
+  } else {
+    InitializeMemoryBuckets ((EFI_PHYSICAL_ADDRESS)(Hob.HandoffInformationTable->EfiFreeMemoryTop));
```
cfernald: Is the use of global here before memory init problematic?

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,332 @@
...
+  RemainingPages = (UINTN)(FreeMemoryTop - FreeMemoryBottom) >> EFI_PAGE_SHIFT;
+  //
+  // The number of remaining pages needs to be greater than or equal to that of the request pages.
+  //
+  Pages = ALIGN_VALUE (Pages, EFI_SIZE_TO_PAGES (Granularity));
+  if (RemainingPages < Pages) {
+    //
+    // Try to find free memory by searching memory allocation HOBs.
+    //
+    // Status = FindFreeMemoryFromMemoryAllocationHob (MemoryType, Pages, Granularity, Memory);
```
cfernald: Is this intentionally left commented?

MdeModulePkg/Core/Pei/Memory/MemoryServices.c
kenlautner: This check is to make sure pei memory allocations that aren't included in the pei memory buckets aren't being allocated within the predefined memory bucket space.  The only time we might care about the pages being allocated is if we're allocating memory before the memory bucket range that might run into it.  This case is impossible because the memory allocation process in PEI is contiguous, and the memory buckets are allocated on the first runtime memory allocation.  So if the starting address is within the range of the pei memory buckets we just need to move it forward outside of that range and we don't have to worry about it.

MdeModulePkg/Library/MemoryBucketHelperLib/MemoryBucketHelperLib.inf
kenlautner: You're correct.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,42 @@
...
+
+typedef struct {
+  EFI_PHYSICAL_ADDRESS    BaseAddress;
+  EFI_PHYSICAL_ADDRESS    MaximumAddress;
+  UINT64                  CurrentNumberOfPages;
+  UINT64                  NumberOfPages;
+  UINT32                  InformationIndex;
+  BOOLEAN                 Special;
+  BOOLEAN                 Runtime;
+} EFI_MEMORY_TYPE_STATISTICS;
```
makubacki: `EFI_MEMORY_TYPE_STATISTICS` is not specific to memory buckets so it should not be in a memory bucket specific header file. If you need the type in this file, please put its definition in a separate header file and include that file here.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,42 @@
...
+
+typedef struct {
+  EFI_MEMORY_TYPE_STATISTICS    RuntimeBuckets[4];
```
makubacki: Use macros for constant numeric values.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifndef _MEMORY_BUCKET_INFORMATION_H_
+#define _MEMORY_BUCKET_INFORMATION_H_
+
+#define MEMORY_BUCKET_INFORMATION_GUID \
+{ 0x36138737, 0xb6db, 0x4ed6, { 0x9b, 0x4b, 0xf9, 0x61, 0x28, 0xe7, 0x19, 0x3c } }
+
+extern EFI_GUID  gMemoryBucketInformationGuid;
+
+#define MEMORY_TYPE_STATISTICS_GUID \
```
makubacki: Same as other comment, move this to a dedicated header file for memory type statistics.

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
makubacki: After adding the new header file, any files dependent on memory type statistics should include that header (not depend on including it through this header file).

MdeModulePkg/Core/Pei/Memory/MemoryServices.c
```diff
@@ -650,6 +652,30 @@ PeiAllocatePages (
...
+    if (IsRuntimeMemoryInitialized ()) {
+      *FreeMemoryTop    = GetCurrentBucketTop (MemoryType);
+      *FreeMemoryBottom = GetCurrentBucketBottom (MemoryType);
+    } else {
+      InitializeMemoryBuckets (*FreeMemoryTop);
+      if (IsRuntimeType (MemoryType)) {
+        *FreeMemoryTop    = GetCurrentBucketTop (MemoryType);
+        *FreeMemoryBottom = GetCurrentBucketBottom (MemoryType);
+      }
+    }
```
makubacki: I think you can clean up/reduce logic here.

```suggestion
  // Check if we're using the memory buckets
  if (IsRuntimeType (MemoryType)) {
    if (!IsRuntimeMemoryInitialized ()) {
      InitializeMemoryBuckets (*FreeMemoryTop);
    }

    *FreeMemoryTop    = GetCurrentBucketTop (MemoryType);
    *FreeMemoryBottom = GetCurrentBucketBottom (MemoryType);
```

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+
+  RuntimeMemoryStats[Index].CurrentNumberOfPages = (UINT64)RuntimeMemoryStats[Index].CurrentNumberOfPages + (UINT64)Pages;
+
+  RuntimeBucketHob.RuntimeBuckets[Index]     = RuntimeMemoryStats[Index];
+  RuntimeBucketHob.CurrentTopInBucket[Index] = CurrentBucketTops[Index];
+}
+
+/**
+  This function checks to see if there is an attempt to allocate memory in the runtime
+  memory buckets that we have defined.
```
makubacki: The current description makes it sound like this is checking the actual memory allocation after the fact. But this function is really checking an address likely before the allocation.

```suggestion
  This function checks to see if the given address lies within the currently defined
  runtime memory bucket range.
```

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+  { 0, MAX_ALLOC_ADDRESS, 0, 0, EfiACPIMemoryNVS,       TRUE, FALSE }    // EfiACPIMemoryNVS
+};
+
+EFI_PHYSICAL_ADDRESS  CurrentBucketTops[4] = {
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesCode
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesData
+  MAX_ALLOC_ADDRESS, // EfiACPIReclaimMemory
+  MAX_ALLOC_ADDRESS  // EfiACPIMemoryNVS
+};
+
```
makubacki, line 47: These global variables need to be prefixed with "m" since they will be module globals.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+#include <PiPei.h>
+
+#include <Library/DebugLib.h>
+#include <Library/MemoryBucketLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/HobLib.h>
+
+// Memory types being kept in buckets in PEI.  Currently only runtime types.
+EFI_MEMORY_TYPE  MemoryTypes[4] = {
```
makubacki: Like the comment in MemoryBucketInformation.h, please use a macro to keep "4" in sync across usage from a single definition.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
```diff
@@ -0,0 +1,36 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = MemoryBucketLib
+  FILE_GUID                      = 916b6cbe-52f3-439b-8f4c-b97453e36cc7
+  MODULE_TYPE                    = PEIM
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = MemoryBucketLib|DXE_CORE DXE_DRIVER DXE_RUNTIME_DRIVER DXE_SMM_DRIVER SMM_CORE PEIM SEC PEI_CORE UEFI_APPLICATION UEFI_DRIVER
```
makubacki: You probably need two separate libraries. PEI has a lot of functionality specific to PEI (producing HOB info / checking info for allocations). DXE is consumer focused. So, either library interfaces are exposed to DXE that don't make sense and/or a lot of irrelevant functionality is being linked to DXE.

In either case, if something is only intended in PEI, place it in a PEI (even `PEI_CORE`) specific library interface/instance. Otherwise, the DXE type that makes sense.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
```diff
@@ -0,0 +1,36 @@
+## @file
+# Library the defines Runtime memory buckets for use during PEI
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = MemoryBucketLib
```
makubacki, line 10: Please prefix the library instance with the phase type, e.g. `PeiMemoryBucketLib`.

makubacki: This also affects the filenames.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
```diff
@@ -0,0 +1,36 @@
+## @file
+# Library the defines Runtime memory buckets for use during PEI
```
makubacki: This library doesn't define the buckets, right? It provides logic to manage the buckets in PEI.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+
+EFI_PHYSICAL_ADDRESS  CurrentBucketTops[4] = {
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesCode
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesData
+  MAX_ALLOC_ADDRESS, // EfiACPIReclaimMemory
+  MAX_ALLOC_ADDRESS  // EfiACPIMemoryNVS
+};
+
+// Total number of memory buckets
+UINTN  NumberOfBuckets = 4;
```
makubacki: Is this meant to be a constant? In any case, please use the macro for the base definition, that can be used in statically sizeed structures like the arrays. If this needs to change, initialize it from the macro, otherwise, delete it.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+  VOID
+  )
+{
+  RuntimeMemoryStats[0].NumberOfPages = FixedPcdGet8 (PcdPeiMemoryBucketRuntimeCode);
+  RuntimeMemoryStats[1].NumberOfPages = FixedPcdGet8 (PcdPeiMemoryBucketRuntimeData);
+  RuntimeMemoryStats[2].NumberOfPages = FixedPcdGet8 (PcdPeiMemoryBucketAcpiReclaimMemory);
+  RuntimeMemoryStats[3].NumberOfPages = FixedPcdGet8 (PcdPeiMemoryBucketAcpiMemoryNvs);
+
+  // Disable memory buckets if the PCDs are unaltered.
+  if ((RuntimeMemoryStats[0].NumberOfPages + RuntimeMemoryStats[1].NumberOfPages +
```
makubacki: Condition would be easier to read of addition was done separately, stored into a local variable, and variable compared here.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+/**
+  TODO
+**/
+VOID
+EFIAPI
+InitializeMemoryBucketSizes (
+  VOID
+  )
+{
+  RuntimeMemoryStats[0].NumberOfPages = FixedPcdGet8 (PcdPeiMemoryBucketRuntimeCode);
```
makubacki: An enum would be nice here. Haven't checked usage of index to bucket type elsewhere yet.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+    MemoryBucketsDisabled = TRUE;
+  }
+}
+
+/**
+  This function initialized the PEI memory buckets.  This should be called
+  when the first memory type being tracked in these buckets is allocated.
+
+  NOTE: The memory buckets can be created in Pre-mem PEI if runtime pages are
+  allocated in pre-mem PEI.
```
makubacki, line 104: Have you accounted for the pre-mem to post-mem memory transition in bucket addresses? If you're tracking addresses in your own HOB/area, you need to manually rebase your addresses across the transition.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+// Information storage for Hob
+PEI_MEMORY_BUCKET_INFORMATION  RuntimeBucketHob;
+
+// TRUE if PEI memory buckets are disabled
+BOOLEAN  MemoryBucketsDisabled = FALSE;
+
+// TRUE if we have initialized the runtime memory buckets.
+BOOLEAN  RuntimeMemInitialized = FALSE;
+
+/**
```
makubacki, line 57: Resolve todo.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+
+  // Disable memory buckets if the PCDs are unaltered.
+  if ((RuntimeMemoryStats[0].NumberOfPages + RuntimeMemoryStats[1].NumberOfPages +
+       RuntimeMemoryStats[2].NumberOfPages + RuntimeMemoryStats[3].NumberOfPages) == 0)
+  {
+    MemoryBucketsDisabled = TRUE;
+  }
+}
+
+/**
```
makubacki, line 87: Can you add an ASCII picture to this function description showing where you're allocating the buckets (and any data structures) relative to the PEI mem map?

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesCode
+  MAX_ALLOC_ADDRESS, // EfiRuntimeServicesData
+  MAX_ALLOC_ADDRESS, // EfiACPIReclaimMemory
+  MAX_ALLOC_ADDRESS  // EfiACPIMemoryNVS
+};
+
+// Total number of memory buckets
+UINTN  NumberOfBuckets = 4;
+
+// Information storage for Hob
```
makubacki, line 48: Are any of these globals potentially getting written in pre-mem? Need to avoid that for PEI core code.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
makubacki: It looks like you're using `MemoryTypeToIndex()` to do the conversion. But it's not being used in places like this leaving a gap between what it returns and what is being used here.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,447 @@
...
+      Index = 1;
+      break;
+    case EfiACPIReclaimMemory:
+      Index = 2;
+      break;
+    case EfiACPIMemoryNVS:
+      Index = 3;
+      break;
+    default:
+      DEBUG ((DEBUG_ERROR, "We got an incorrect MemoryType\n"));
```
makubacki: Recommend adding function name. Will make a lot more sense in the context of the full debug log.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
```diff
@@ -0,0 +1,36 @@
+## @file
```
makubacki, line 1: Can you please add at least a brief readme to explain memory buckets and important interfaces being added. This library directory seems like an okay place to do that.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
makubacki: There's also logic present like HOB producing logic that can't be used in DXE.

MdeModulePkg/Core/Pei/Memory/MemoryServices.c
kenlautner: You're right about the general logic.  There is a reason why I include an additional IsRuntimeType check which is to deny memory bucket operations if after initializing we can find that the platform doesn't want to use memory buckets.  That should be its own separate function though, so I'll make that adjustment.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: You're suggesting to use another enum to represent the 0-3 values for easier readability?

MdeModulePkg/Include/Guid/MemoryBucketInformation.h
kenlautner: Updated.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
kenlautner: Switched to just PEI_CORE.  This shouldn't be used by anything else.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
kenlautner: Updated.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: Added.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: I added some logic for this but would appreciate feedback.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1272,6 +1281,13 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportAlternativeQueueSize|FALSE|BOOLEAN|0x40000151
   # MU_CHANGE [END]
 
+  # MU_CHANGE [BEGIN] - Add structure PCD for PEI memory buckets
```
makubacki: These look like normal (non structure) PCDs.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -658,7 +667,7 @@
   # MU_CHANGE
   # This protocol provides access to Internal Event services that do not need TPL_APPLICATION
   gInternalEventServicesProtocolGuid = {0x7ecd162a, 0xc664, 0x11ec, { 0x9d, 0x64, 0x02, 0x42, 0xac, 0x12, 0x00, 0x02 } }
-
+
```
makubacki: I noticed some whitespace come into the file in a few places. Can you please try to trim trailing whitespace in all files touched?

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
```diff
@@ -0,0 +1,36 @@
+## @file
+# Library the creates and manages Runtime memory buckets for use during PEI
```
makubacki:
```suggestion
# Library that creates and manages Runtime memory buckets for use during PEI
```

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,530 @@
+/**@file
+Library for defining PEI memory buckets. This keeps track of the
+different buckets and the data for the associated HOB.
+For internal use only.
```
makubacki: Internal is a bit ambiguous here (i.e. "internal" to a company, package, driver, etc.).

Can you please add some more context?

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,530 @@
...
+EFIAPI
+SetMemoryBucketsFromHob (
+  VOID  *MemoryBuckets
+  )
+{
+  PEI_MEMORY_BUCKET_INFORMATION  *TempStats;
+  UINTN                          Index;
+
+  if (MemoryBuckets == NULL) {
+    DEBUG ((DEBUG_ERROR, "NO PEI RUNTIME MEMORY BUCKETS YET\n"));
```
makubacki, line 477: If this is an unexpected condition, an `ASSERT` would help bring more attention to it.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -0,0 +1,530 @@
...
+    return;
+  }
+
+  TempStats = (PEI_MEMORY_BUCKET_INFORMATION *)MemoryBuckets;
```
makubacki: Can the `MemoryBuckets` parameter to the function just be of `PEI_MEMORY_BUCKET_INFORMATION *` type? If that's the only type expected, it will make the API more explicit.

makubacki:
Please update the PR description. It still describes things like `MemoryBucketHelperLib` which are no longer present.

makubacki:
It looks like `MemoryBucketLib` could write to global variables before main memory is initialized. Is that correct?

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
kenlautner: Added.

MdeModulePkg/MdeModulePkg.dec
kenlautner: Updated.

MdeModulePkg/MdeModulePkg.dec
kenlautner: Fixed the white space changes.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: Updated to say the function shouldn't be called outside of the PEI CORE.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: Updated.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: That makes sense.  Updated.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.inf
kenlautner: Fixed.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
kenlautner: I removed my initial logic and made it so that none of the PEI memory bucket stuff happens in pre-mem PEI.   It makes more sense to do this in my opinion.

kenlautner:
> It looks like `MemoryBucketLib` could write to global variables before main memory is initialized. Is that correct?

The global variables shouldn't be written in pre-mem PEI.

MdeModulePkg/Core/Pei/Memory/MemoryServices.c
```diff
@@ -650,6 +652,28 @@ PeiAllocatePages (
     FreeMemoryBottom = &(Hob.HandoffInformationTable->EfiFreeMemoryBottom);
   }
 
+  // MU_CHANGE START
+  MemBucketHob = GetFirstGuidHob (&gMemoryBucketInformationGuid);
```
apop5, line 656: Should this handle the case where the hob doesn't exist (i.e. GetFirstGuidHob returns NULL)?

kenlautner: It's checked in the SyncMemoryBuckets function that it's passed into.

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
```diff
@@ -22,6 +22,13 @@ EFI_MEMORY_TYPE  mMemoryTypes[PEI_BUCKETS] = {
   EfiACPIMemoryNVS
 };
 
+enum {
```
mikeytdisco, line 26: We already have an ENUM of memory types in include\Uefi\UefiMultiPhase.h --> EFI_MEMORY_TYPE.  Another enum with different values for the same thing shouldn't be created.

kenlautner: Because we're not keeping track of all the memory types with the PEI memory buckets the enum in UefiMultiPhase doesn't match up with the indexes in the structures I created.  I originally hard coded numbers but it was recommended to me to make another enum to make the code more readable.

makubacki:
Note: Plan is to get this merged by Friday, 2/3/2023 PST

MdeModulePkg/Library/MemoryBucketLib/MemoryBucketLib.c
mikeytdisco: Or, you could sparsely populate your arrays - it isn't that much memory.

makubacki: A related but larger issue I see is lack of cohesion in data structures leading to implicit dependencies between indices and redundant information.

I think it'd be easier for me to propose a refactor than explain so I'll do that and send it soon.

makubacki:
@kenlautner I'm going to work on another PR into this one that has functional changes. For now, there's a few finishing touches (mostly non-functional) that were easier to directly change than explain individually in the following PR.

Can you please check it and merge it into your branch? It should automatically show up here after that.

https://github.com/kenlautner/mu_basecore/pull/1

makubacki:
> @kenlautner I'm going to work on another PR into this one that has functional changes.

Based on our most recent offline discussion, I'm not going to make that PR and wait for your rough draft of the changes discussed and we'll go from there. Feel free to let me know if there's any pieces you'd like me to help out with.

PR #168, opened 2022-11-11
Enable Data Terminal Ready and Request to Send at the end of a Serial…

apop5:
## Description
When a Serial device reset, the Modem Control Register Data Terminal Ready and Request to Send need to be cleared also. Otherwise the registers will be left in their previous state, and the connected device will not be able to transmit data. 

A particular device was not able to receive data after a SerialReset call because the Dtrc and Rts bits were left in their previous states. 

- [ ] Breaking change?
This should not break existing implementations.

## How This Was Tested
Tested using a 16550 "compatible" pci serial device. 

## Integration Instructions
N/A


makubacki (approved):
It looks like `MU_CHANGE` tags are needed. Can you please add those?

apop5:
> It looks like `MU_CHANGE` tags are needed. Can you please add those?

Tags have been added.

spbrogan (approved):

PR #182, opened 2022-11-18
UefiBootManagerLib: Update assert condition in BmFindBootOptionInVariable()

makubacki:
## Description

Currently the following assertion is made in
`BmFindBootOptionInVariable()`:

```c
  if (OptionNumber == LoadOptionNumberUnassigned) {
    BootOptions = EfiBootManagerGetLoadOptions (&BootOptionCount, LoadOptionTypeBoot);

    if (BootOptions == NULL) {
      ASSERT (BootOptions != NULL);
      return LoadOptionNumberUnassigned;
    }

    Index = EfiBootManagerFindLoadOption (OptionToFind, BootOptions, BootOptionCount);
```

The reason behind this is to prevent passing a null pointer to
`EfiBootManagerFindLoadOption()`.

However, `EfiBootManagerFindLoadOption()` accepts a null pointer
argument to be passed as the second argument ('Array') as long as
the `Count` argument is `0`.

In that case `LoadOptionNumberUnassigned` will still be returned from
the function but an assert is not necessary.

This change updates the condition to only assert if the pointer is
null and the boot option count is non-zero.

- [ ] Breaking change?
  **No**

## How This Was Tested

Verified build and boot on Qemu35Pkg.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
+++ b/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
@@ -147,7 +147,7 @@ BmFindBootOptionInVariable (
   if (OptionNumber == LoadOptionNumberUnassigned) {
     BootOptions = EfiBootManagerGetLoadOptions (&BootOptionCount, LoadOptionTypeBoot);
 
-    if (BootOptions == NULL) {
+    if ((BootOptions == NULL) && (BootOptionCount > 0)) {
       ASSERT (BootOptions != NULL);
       return LoadOptionNumberUnassigned;
     }


kuqin12 (approved):

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -147,7 +147,7 @@ BmFindBootOptionInVariable (
   if (OptionNumber == LoadOptionNumberUnassigned) {
     BootOptions = EfiBootManagerGetLoadOptions (&BootOptionCount, LoadOptionTypeBoot);
 
-    if (BootOptions == NULL) {
```
apop5, line 150: OptionToFind is not NULL checked before its dereferenced in the start of the function (outside of your change)

if (OptionToFind->OptionNumber != LoadOptionNumberUnassigned)

apop5 (approved):

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
makubacki: I filed an issue to track that separately - https://github.com/microsoft/mu_basecore/issues/183

As far as I can tell, the only caller does at least check if it's null before passing it here.

https://github.com/microsoft/mu_basecore/blob/cd15357d98e9e29288040555b3f85ce6ce2e2a32/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c#L1885

PR #184, opened 2022-11-19
Remove .cargo/config.toml file

Javagedes:
Previously Z flags were set in the projects .cargo/config.toml file. This file has been deleted as it forced host target builds (for tests) to use compiler flags meant for target based builds. To ensure the flags continue to be used during target builds, they are added to the build command itself.

## Description

<_Please include a description of the change and why this change was made._>

- [ ] Breaking change?

## How This Was Tested

<_Please describe the test(s) that were run to verify the changes._>

## Integration Instructions

### DIFF
--- a/BaseTools/Conf/build_rule.template
+++ b/BaseTools/Conf/build_rule.template
@@ -281,7 +281,7 @@
         $(OUTPUT_DIR)(+)$(MODULE_NAME)rust.lib
 
     <Command>
-        $(CARGO) rustc $(CARGO_FLAGS) --manifest-path ${s_path}(+)Cargo.toml -Z unstable-options --target-dir $(CARGO_OUTPUT_DIR) --timings=html
+        $(CARGO) rustc $(CARGO_FLAGS) --manifest-path ${s_path}(+)Cargo.toml $(RUST_ZFLAGS) --target-dir $(CARGO_OUTPUT_DIR) --timings=html
         $(CP) $(CARGO_BINDIR)(+)*.a $(OUTPUT_DIR)(+)$(MODULE_NAME)rust.lib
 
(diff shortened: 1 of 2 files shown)


makubacki:
How about running cargo directly in the rust package folder (folder with Cargo.toml) after this change?

Javagedes:
> How about running cargo directly in the rust package folder (folder with Cargo.toml) after this change?

If running cargo manually, the user would have to add these flags to the command themselves.

Javagedes:
Instead of removing the .cargo/config I can probably make it more advanced to handle these situations, however you’re now splitting build control between UEFI’s build settings and cargos build settings. We either need to give 100% control to UEFI build system (cargo build + all flags are set in the Conf files) or we need to give 100% control to cargo (all flags for all situations set in .cargo/config and conf files only call cargo build).

makubacki:
I'd like this to work like a normal Rust project as much as possible and only deviate where absolutely necessary. I understand that deviation or movement of settings might be necessary given we're building within the edk2 build system. This might be such a case, but I want to make sure we've exhausted all paths to leverage cargo configuration which would be more intuitive for those outside the edk2 community and generally comes with better documentation than the edk2 build environment.

I'm fine with the second option - giving 100% or close to it control to cargo (for what it is capable of handling).

For example, I thought build profiles in cargo files help naturally split settings there. Not saying that's applicable here but it's well documented and naturally aligns with the cargo command being used.

How complicated do you think doing this in the cargo config file would be?

Javagedes:
> How complicated do you think doing this in the cargo config file would be?

I need to research a couple of things but I don’t think it will be difficult to move more into the .cargo/config rather then remove it. I’ll move this to draft while I look into this.

PR #190, opened 2022-11-29
Add support for HTTP error 429

mikeytdisco:
## Description

HTTP status codes have to be explicitly specified in DxeHttpLib in order to be supported.  Intune may return 429 to throttle requests.

- [ No] Breaking change?

## How This Was Tested

Compiled.

## Integration Instructions

N/A


apop5 (approved):

makubacki (approved):

makubacki:
@mikeytdisco I believe you mentioned this offline earlier but the enum value `HTTP_STATUS_429_TOO_MANY_REQUESTS` seems a bit concerning for divergence from the upstream code.

Do you think you should file a TianoCore BZ describing the issue and link it in this Mu code?

PR #192, opened 2022-11-30
Refactor Policy Service to allow a name as well as a guid

cfernald:
## Description

This adds an optional NAME field to the policy service interface. This new name is used in conjunction with the GUID value to have unique GUID-name pairs. The NULL name is treated as it's own unique guid-name pair. 

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with the updated PEI/DXE module tests on the Q35 platform

## Integration Instructions

All existing references to the policy service should add a NULL value for the name field on all references to continue using it as-is.


PolicyServicePkg/PolicyService/PolicyCommon.h
```diff
@@ -9,15 +9,44 @@
...
   EFI_GUID    PolicyGuid;
   UINT64      Attributes;
   UINT16      PolicySize;
   UINT16      AllocationSize;
-  UINT32      Removed  : 1;
-  UINT32      Reserved : 31;
+  UINT16      Removed  : 1;
+  UINT16      Reserved : 15;
+  UINT16      NameSize;
+  // CHAR16 Name[NameSize]
```
kuqin12: Does it work if we leave it as `CHAR16 Name[];`?

PolicyServicePkg/PolicyService/Dxe/PolicyDxe.c
```diff
@@ -257,7 +285,7 @@ DxeRemovePolicy (
   }
 
   EfiAcquireLock (&mPolicyListLock);
-  Status = DxeGetPolicyEntry (PolicyGuid, &Entry);
+  Status = DxeGetPolicyEntry (PolicyGuid, Name, &Entry);
```
kuqin12, line 288: Free NameBuffer?

kuqin12 (approved):

PolicyServicePkg/PolicyService/PolicyCommon.c
```diff
@@ -0,0 +1,40 @@
...
+  )
+{
+  BOOLEAN  Result;
+
+  if ((Name1 == NULL) && (Name2 == NULL)) {
+    Result = TRUE;
+  } else if ((Name1 == NULL) || (Name2 == NULL)) {
+    Result = FALSE;
+  } else {
+    Result = (StrCmp (Name1, Name2) == 0);
```
os-d: Should we validate the string is not longer than `PcdMaximumUnicodeStringLength`? Or only compare that length, at any rate?

PolicyServicePkg/Library/VerifiedPolicy.c
```diff
@@ -1,247 +1,247 @@
...
+  BufferSize  = 0;
+  Header      = NULL;
+  *DataHandle = NULL;
+
+  Status = GetPolicyInterface (&PolicyService);
+  if (EFI_ERROR (Status)) {
+    goto Exit;
+  }
+
+  Status = PolicyService->GetPolicy (PolicyGuid, NULL, Attributes, &Buffer, &BufferSize);
```
os-d: VerifiedPolicy.c's APIs should be updated to include Name as a passed parameter, right?

os-d (changes requested):

makubacki:
@cfernald the PR template was briefly removed from the repo in prep for the file sync. Can you please update the PR message with the template that's there now?

PolicyServicePkg/Library/VerifiedPolicy.c
cfernald: I actually intentionally left these as is. The verified policies will all be auto-generated and therefore should have a unique GUID with no name. Do we intend to use the verified policy lib for the config work?

PolicyServicePkg/PolicyService/PolicyCommon.h
cfernald: It's variable length. So it would need at least length 1 since the compilers do not like 0-length arrays. I'd rather just leave it as an appended field personally.

PolicyServicePkg/Library/VerifiedPolicy.c
os-d: That's how I have it written up currently, exclusively using the verified policy lib. My understanding was that was the preferred mechanism to interact with policies. Have I misunderstood? I don't see where the auto-generation is happening, the caller of SetVerifiedPolicy passes the GUID that is used, right? My thought was that the caller would also pass the Name, if applicable.

os-d (approved):

PolicyServicePkg/PolicyService/PolicyCommon.c
cfernald: This should already be done implicitly by the BaseLib implementation of StrCmp

PolicyServicePkg/PolicyService/PolicyCommon.h
cfernald: Sure. I think I may actuall keep the macro though because the easy NULL check is good and will prevent accidental misuse of the array.

PR #194, opened 2022-11-30
Create memory bin override library to allow for more extensive platform customization 

cfernald:
## Description

Adds a memory bin override library that allows a platform to customize logic around memory bin locations in DXE.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on the Q35 platform with a custom memory bin override library

## Integration Instructions

Add the new NULL version of the library to the platform DSC file
```
MemoryBinOverrideLib|MdeModulePkg/Library/MemoryBinOverrideLibNull/MemoryBinOverrideLibNull.inf
```


os-d (approved):

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -616,16 +618,29 @@ CoreAddMemoryDescriptor (
     }
 
     if (gMemoryTypeInformation[Index].NumberOfPages != 0) {
+      // MU_CHANGE START Allow overriding of bin locations.
+      AllocationType = AllocateAnyPages;
+      CheckMemoryBinOverride (
+        Type,
+        &mMemoryTypeStatistics[Type].BaseAddress,
```
apop5, line 625: is the base address necessary here? 

it gets passed into CheckMemoryBinOverride, but then in CoreAllocatepages below, it will be overwritten...

apop5: Oh, Depending on the allocation type returned, it might be requesting a specific address.

apop5 (approved):

makubacki:
@cfernald have you and @kenlautner discussed how this and https://github.com/microsoft/mu_basecore/pull/167 would cohesively work together. Seems like a design plan should take both changes into account.

cfernald:
> @cfernald have you and @kenlautner discussed how this and #167 would cohesively work together. Seems like a design plan should take both changes into account.

Agreed. This change will likely be over-hauled after merging the PEI phase bucket work. Due to this change being needed to unblock other work however, I will come back and do that work after both of these changes are merged.

kenlautner:
If this change only cares about the DXE memory buckets then with how my PR is currently structured it shouldn't affect it.  The DXE memory buckets don't really care about what the PEI memory looks like.

MdeModulePkg/Include/Library/MemoryBinOverrideLib.h
```diff
@@ -0,0 +1,48 @@
+/** @file
```
makubacki: Library class header needs to be added to MdeModulePkg.dec

MdeModulePkg/Library/MemoryBinOverrideLibNull/MemoryBinOverrideLibNull.c
```diff
@@ -0,0 +1,53 @@
...
+  NULL implementation of the memory bin override lib.
+
+  Copyright (c) Microsoft Corporation. All rights reserved.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include <Uefi.h>
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
```
makubacki: Was `DebugLib` really needed in this file?

makubacki:
> If this change only cares about the DXE memory buckets then with how my PR is currently structured it shouldn't affect it. The DXE memory buckets don't really care about what the PEI memory looks like.

I'm not so concerned about whether the two can coexist but that that they present cohesive interfaces. For example, this PR adds `MemoryBinOverrideLib` and the other PR adds `MemoryBucketLib`. Do we really need to call mix & match bin & bucket terminology?

MdeModulePkg/Include/Library/MemoryBinOverrideLib.h
```diff
@@ -0,0 +1,48 @@
...
+  @param[out]   NumberOfPages   The number of pages of the bin override on return.
+  @param[out]   AllocationType  The allocation type for the bin, AllocateAddress
+                                if an override was provided.
+**/
+VOID
+EFIAPI
+CheckMemoryBinOverride (
+  IN EFI_MEMORY_TYPE         Type,
+  OUT EFI_PHYSICAL_ADDRESS   *BaseAddress,
+  OUT OUT UINT32             *NumberOfPages,
```
kuqin12: Duplicated `OUT`?

kuqin12 (approved):

MdeModulePkg/Include/Library/MemoryBinOverrideLib.h
```diff
@@ -0,0 +1,48 @@
...
+
+  @param[in]    Type            The memory type of the bin.
+  @param[out]   BaseAddress     The base address of the bin override on return.
+  @param[out]   NumberOfPages   The number of pages of the bin override on return.
+  @param[out]   AllocationType  The allocation type for the bin, AllocateAddress
+                                if an override was provided.
+**/
+VOID
+EFIAPI
+CheckMemoryBinOverride (
```
makubacki: I think `GetMemoryBinOverride()` would be more precise to what's happening here than check since you're getting back the override values rather than a bool indicating if an override is needed.

makubacki (approved):

cfernald:
> > If this change only cares about the DXE memory buckets then with how my PR is currently structured it shouldn't affect it. The DXE memory buckets don't really care about what the PEI memory looks like.
> 
> I'm not so concerned about whether the two can coexist but that that they present cohesive interfaces. For example, this PR adds `MemoryBinOverrideLib` and the other PR adds `MemoryBucketLib`. Do we really need to call mix & match bin & bucket terminology?

I think this is fair. I intend to update the functionality here once we have PEI memoru buckets as those will also be accounted for by the platform. For now I will leave it as it to and will update post check-in of the PEI memory buckets.

makubacki:
> > > If this change only cares about the DXE memory buckets then with how my PR is currently structured it shouldn't affect it. The DXE memory buckets don't really care about what the PEI memory looks like.
> > 
> > 
> > I'm not so concerned about whether the two can coexist but that that they present cohesive interfaces. For example, this PR adds `MemoryBinOverrideLib` and the other PR adds `MemoryBucketLib`. Do we really need to call mix & match bin & bucket terminology?
> 
> I think this is fair. I intend to update the functionality here once we have PEI memoru buckets as those will also be accounted for by the platform. For now I will leave it as it to and will update post check-in of the PEI memory buckets.

Sounds good. We technically should reduce thrash of public interfaces (like decide on bin vs buckets before the interface gets introduced). But I mainly wanted to raise the idea that we should make sure these things go together well. For what it's worth, I marginally prefer bucket simply because "bin" could be confused with "binary" for the unaware. But not a big deal either way.

cfernald:
> > > > If this change only cares about the DXE memory buckets then with how my PR is currently structured it shouldn't affect it. The DXE memory buckets don't really care about what the PEI memory looks like.
> > > 
> > > 
> > > I'm not so concerned about whether the two can coexist but that that they present cohesive interfaces. For example, this PR adds `MemoryBinOverrideLib` and the other PR adds `MemoryBucketLib`. Do we really need to call mix & match bin & bucket terminology?
> > 
> > 
> > I think this is fair. I intend to update the functionality here once we have PEI memoru buckets as those will also be accounted for by the platform. For now I will leave it as it to and will update post check-in of the PEI memory buckets.
> 
> Sounds good. We technically should reduce thrash of public interfaces (like decide on bin vs buckets before the interface gets introduced). But I mainly wanted to raise the idea that we should make sure these things go together well. For what it's worth, I marginally prefer bucket simply because "bin" could be confused with "binary" for the unaware. But not a big deal either way.

I too prefer bucket for this reason, though I chose bin here as page.c refers to these as bins throughout currently.

PR #196, opened 2022-12-01
[Rebase & FF] Introducing MP services test application

kuqin12:
## Description

This change introduces a new UEFI shell test application to perform various multi-threaded
operations based on MP services.

A `MU_CHANGE` in DSC file is added to pass the package entanglement.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change itself is a test application. Verified functionality on QEMU platform.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
+++ b/MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
@@ -100,6 +100,10 @@ PrintProcessorInformation (
   UINTN                      NumCpu;
   UINTN                      NumEnabledCpu;
 
+  // MU_CHANGE: Initialize local variables in case there is error in the function call.
+  NumCpu        = 0;
+  NumEnabledCpu = 0;
+
   Status = Mp->GetNumberOfProcessors (Mp, &NumCpu, &NumEnabledCpu);
   if (EFI_ERROR (Status)) {
     Print (L"GetNumberOfProcessors failed: %r\n", Status);
@@ -201,7 +205,9 @@ AllocateApFuncBufferAllAPs (
   IN UINTN                     NumCpus
   )
 {
-  INT32  Index;
+  // MU_CHANGE: Conformed to use UINTN and an extra index to clean up during failure
+  UINTN  Index;
+  UINTN  FailureIndex;
 
   ApArg->Mp = Mp;
 
@@ -215,8 +221,9 @@ AllocateApFuncBufferAllAPs (
     ApArg->Buffer[Index] = AllocateZeroPool (APFUNC_BUFFER_LEN);
     if (ApArg->Buffer[Index] == NULL) {
       Print (L"Failed to allocate buffer for AP message\n");
-      for (--Index; Index >= 0; Index++) {
-        FreePool (ApArg->Buffer[Index]);
+      // MU_CHANGE: Uses an extra index to clean up during failure
+      for (FailureIndex = 0; FailureIndex < Index; FailureIndex++) {
+        FreePool (ApArg->Buffer[FailureIndex]);
       }
 
       FreePool (ApArg->Buffer);
@@ -270,11 +277,6 @@ StartupThisAP (
 {
   EFI_STATUS  Status;
   APFUNC_ARG  ApArg;
-
-  Status = AllocateApFuncBufferSingleAP (&ApArg, Mp, ProcessorIndex);
-  if (EFI_ERROR (Status)) {
-    return Status;
-  }
 
   Status = AllocateApFuncBufferSingleAP (&ApArg, Mp, ProcessorIndex);
   if (EFI_ERROR (Status)) {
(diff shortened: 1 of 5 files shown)


MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
```diff
@@ -0,0 +1,561 @@
...
+  @param ProcessorIndex   The processor to enable.
+  @param ProcessorHealthy The health status of the processor.
+
+  @return EFI_SUCCESS on success, or an error code.
+**/
+STATIC
+EFI_STATUS
+EnableAP (
+  IN EFI_MP_SERVICES_PROTOCOL  *Mp,
+  UINTN                        ProcessorIndex,
```
cfernald: nit 'IN' for the two other params

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
```diff
@@ -0,0 +1,561 @@
...
+  @param ProcessorIndex   The processor to disable.
+  @param ProcessorHealthy The health status of the processor.
+
+  @return EFI_SUCCESS on success, or an error code.
+**/
+STATIC
+EFI_STATUS
+DisableAP (
+  IN EFI_MP_SERVICES_PROTOCOL  *Mp,
+  UINTN                        ProcessorIndex,
```
cfernald: IN for these as well

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
```diff
@@ -0,0 +1,561 @@
...
+UINTN
+PrintProcessorInformation (
+  IN   EFI_MP_SERVICES_PROTOCOL  *Mp,
+  OUT  UINTN                     *BspIndex
+  )
+{
+  EFI_STATUS                 Status;
+  EFI_PROCESSOR_INFORMATION  CpuInfo;
+  UINTN                      Index;
+  UINTN                      NumCpu;
```
cfernald, line 100: Should NumCpu be set to 0 in case there is in error in GetNumberOfProcessors?

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
```diff
@@ -0,0 +1,561 @@
...
+                 NULL
+                 );
+  if (EFI_ERROR (Status)) {
+    Print (L"failed: %r\n", Status);
+    return Status;
+  } else {
+    Print (L"done.\n");
+    for (Index = 0; Index < NumCpus; Index++) {
+      Print (ApArg.Buffer[Index]);
+    }
```
cfernald, line 363: For my education, why do we just print the value instead of programmatically checking for a response from each core?

os-d (approved):

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
kuqin12: This test is intended to provide more of a "visual" response. So each AP function, in this case, will populate the buffer with strings for the BSP to print out. But yes, if such printing is not needed, the APs can populate the result code into common buffer for the BSP to check.

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
```diff
@@ -0,0 +1,561 @@
...
+StartupThisAP (
+  IN EFI_MP_SERVICES_PROTOCOL  *Mp,
+  IN UINTN                     ProcessorIndex,
+  IN UINTN                     Timeout
+  )
+{
+  EFI_STATUS  Status;
+  APFUNC_ARG  ApArg;
+
+  Status = AllocateApFuncBufferSingleAP (&ApArg, Mp, ProcessorIndex);
```
apop5, line 281: this is being called two times in a row? 

AllocateApFuncBufferSingleAP()?

kuqin12: Nope, this should be wrong. Will remove.

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
kuqin12: Updated.

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
kuqin12: Updated.

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
kuqin12: Removed. Thanks.

MdeModulePkg/Application/MpServicesTest/MpServicesTest.c
kuqin12: Thanks, updated to initialize both NumCpu and NumEnabledCpu.

apop5 (approved):

PR #200, opened 2022-12-05
Update FillInMemoryMap() to Ignore Regions of GCD Memory Type Non-Existent

TaylorBeebe:
## Description

The FillInMemoryMap() routine which fills in gaps in the EFI memory map based on the GCD memory map. This PR updates the behavior of FillInMemoryMap() to ignore regions not present in the EFI memory map but present in the GCD memory map which are of type EfiGcdMemoryTypeNonExistent.

Platforms don't need to do anything special to integrate this change, but it may solve problems caused by the memory protection initialization routine attempting to set access attributes on memory ranges which aren't configurable via a 4-level page table.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Comparing the updated EFI memory map with and without the changes.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -1301,15 +1301,16 @@ FillInMemoryMap (
                           &OverlapLength,
                           &GcdType
                           );
+      if (GcdType != EfiGcdMemoryTypeNonExistent) {
+        POPULATE_MEMORY_DESCRIPTOR_ENTRY (
+          NewMemoryMapCurrent,
+          StartOfAddressSpace,
+          EfiSizeToPages (OldMemoryMapCurrent->PhysicalStart - StartOfAddressSpace - RemainingLength),
+          GcdTypeToEfiType (&GcdType)
+          );
+        NewMemoryMapCurrent = NEXT_MEMORY_DESCRIPTOR (NewMemoryMapCurrent, *DescriptorSize);
+      }
 
-      POPULATE_MEMORY_DESCRIPTOR_ENTRY (
-        NewMemoryMapCurrent,
-        StartOfAddressSpace,
-        EfiSizeToPages (OldMemoryMapCurrent->PhysicalStart - StartOfAddressSpace - RemainingLength),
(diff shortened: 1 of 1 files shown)


apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1301,15 +1301,16 @@ FillInMemoryMap (
...
                           );
+      if (GcdType != EfiGcdMemoryTypeNonExistent) {
+        POPULATE_MEMORY_DESCRIPTOR_ENTRY (
+          NewMemoryMapCurrent,
+          StartOfAddressSpace,
+          EfiSizeToPages (OldMemoryMapCurrent->PhysicalStart - StartOfAddressSpace - RemainingLength),
+          GcdTypeToEfiType (&GcdType)
+          );
+        NewMemoryMapCurrent = NEXT_MEMORY_DESCRIPTOR (NewMemoryMapCurrent, *DescriptorSize);
+      }
```
os-d, line 1312: To clarify, in the case of holes in the memory map that get remapped later, we would not map them in this first pass (as they are EfiGcdMemoryTypeNonExistent) but when the memory gets added, later, we will simply insert the memory region into the proper place in the memory map? I.e. this won't be an updating of region type in the memory map, but actually a physical addition of a region to the map.

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
apop5: This code is running as part of CpuDxe.

CpuDxe is operating on a copy of the memory map, as it exists at this point in execution.
The local copy of the memory map will be used to update the cpu page tables.

When PciHostBridge runs, https://github.com/microsoft/mu_basecore/blob/release/202208/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c#L545
it should add memory space, set attributes, and allocate memory space, which should follow mu memory protections, as reported for the hob.

PR #203, opened 2022-12-09
Update CryptoPkg binary generation to always include EC

kenlautner:
## Description

Updates the prebuilt crypto to always have Elliptic curve cryptographic functionality.  This is necessary to use TLS in any capacity.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested a TLS connection on Q35.  Before the change a connection would fail to establish due to lacking any cryptographic function to use during the initial handshake.  After this change the connection can be established as intended.

## Integration Instructions

Update your cryptographic binaries to 2022.08.05


kenlautner:
Note:  I need to update the binary being pointed to.

makubacki:
I suggest putting all of the files that only have auto generated timestamp updates in a single commit to separate them from functional changes. Then rebase & ff the commits instead of squash.

It will make it easier later to distinguish the actual changes from those.

makubacki:
Please use a fork instead of directly pushing to a personal branch on this repo.

CryptoPkg/Library/Include/openssl/opensslconf.h
```diff
@@ -17,7 +17,7 @@ extern "C" {
 #endif
 
 /* Autogenerated conditional openssl feature list starts here */
-#if !FixedPcdGetBool (PcdOpensslEcEnabled)
```
kuqin12, line 20: dumb question: why is this change needed?

kuqin12 (approved):

CryptoPkg/Library/Include/openssl/opensslconf.h
kenlautner: I'm updating the commit history to make this cleaner but I removed it to ensure that we're always including Elliptic curve cryptography.  It's assumed we always have it in our current implementation.  This is a bit redundant with me setting the PCD to be always TRUE but I thought I might as well be thorough.  I'll add a comment.

makubacki (approved):

makubacki:
Please update the PR title from "Personal/klautner/ec fix" to something more relevant. Will make it easier to know what it changes in the list of closed PRs later.

makubacki:
> Please update the PR title from "Personal/klautner/ec fix" to something more relevant. Will make it easier to know what it changes in the list of closed PRs later.

@kenlautner, notifying for visibility since this does not count as PR feedback

PR #208, opened 2022-12-13
Add Readme Section on Memory Protection Special Regions

TaylorBeebe:
## Description

Memory protection special region functionality was added last month. This PR adds documentation for this feature to the memory protection doc.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


Docs/feature_memory_protection.md
```diff
@@ -401,3 +401,66 @@ manner similar to below:
 ```
 
 before building the HOB.
+
+## **Memory Protection Special Regions**
+
+Memory protection is not activated until the CPU Architecture Protocol has been installed because the
+protocol facilitates access to the attribute set functions in CpuDxe. Many allocations and image loads
```
apop5: "facilitates access to the attribute set functions in CpuDxe."

I think this should mention that the "attribute set function in CpuDxe" means that it goes to the Page Tables.

Docs/feature_memory_protection.md
```diff
@@ -401,3 +401,66 @@ manner similar to below:
...
 
 before building the HOB.
+
+## **Memory Protection Special Regions**
+
+Memory protection is not activated until the CPU Architecture Protocol has been installed because the
+protocol facilitates access to the attribute set functions in CpuDxe. Many allocations and image loads
+will have occurred by the time the protocol is published so careful accounting is required to ensure
+appropriate attributes are applied. An event notification triggered on the CPU Architecture Protocol
+installation will combine the GCD and EFI memory maps to create a full map of memory. Because image memory
```
apop5: "full map of memory"

Can you add that this full map of memory is internal to the memory protection code, and that it is not reflected in the GCD/EFI Memory map?

apop5 (approved):

Docs/feature_memory_protection.md
TaylorBeebe: Updated

Docs/feature_memory_protection.md
TaylorBeebe: Updated

kenlautner (approved):

Docs/feature_memory_protection.md
```diff
@@ -401,3 +401,67 @@ manner similar to below:
...
+protocol facilitates access to the attribute manipulation functions in CpuDxe which update the translation/page tables.
+Many allocations and image loads will have occurred by the time the protocol is published so careful accounting
+is required to ensure appropriate attributes are applied. An event notification triggered on the CPU Architecture
+Protocol installation will combine the GCD and EFI memory maps to create a full map of memory for use internally
+by the memory protection initialization logic. Because image memory is allocated for the entire image and not each
+section (code and data), the images are separated within
+the combined map so NX can be applied to data regions and RO can be applied to code regions. After breaking
+up the map so each DXE image section has its own descriptor, every non-image descriptor will have its attributes
+set based on its EFI memory type. There are cases where the platform will want to attributes applied to
+a region of memory which is different than what would be applied based on its EFI memory type. In this case,
```
makubacki:
```suggestion
set based on its EFI memory type. There are cases where the platform will want to apply attributes to
a region of memory which is different than what would be applied based on its EFI memory type. In this case,

```

Docs/feature_memory_protection.md
```diff
@@ -401,3 +401,67 @@ manner similar to below:
...
+platforms can utilize the Memory Protection Special Region interface to specify regions which should have specific
+attributes applied during memory protection initialization.
+
+### In PEI:
```
makubacki: If this is a PEI example, I think the section would be a bit clearer named as "Example: Declaring special memory protection region".

makubacki: Same for the DXE example.

makubacki (approved):

Docs/feature_memory_protection.md
TaylorBeebe: Updated

PR #211, opened 2022-12-14
Set MMIO NX Based on Memory Protection Policy

TaylorBeebe:
## Description

We currently apply NX to MMIO ranges converted during PciHostBridge initialiation. This update makes NX application dependent upon the memory protection policy.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting to shell

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
+++ b/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
@@ -557,6 +557,7 @@ InitializePciHostBridge (
         if (gDxeMps.NxProtectionPolicy.Fields.EfiMemoryMappedIO != 0) {
           Attributes |= EFI_MEMORY_XP;
         }
+
         Status = gDS->SetMemorySpaceAttributes (
                         HostAddress,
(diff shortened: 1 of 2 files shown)


MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
```diff
@@ -548,14 +550,18 @@ InitializePciHostBridge (
                    EFI_MEMORY_UC
```
os-d: I think you are out of date with upstream, this should have XP passed here, too, but could just be `Attributes` if the memory protection policy check is moved up.

os-d (approved):

MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciHostBridge.c
TaylorBeebe: I think the capabilities should include EFI_MEMORY_XP regardless of if we actually set it in the following call

kenlautner (approved):

makubacki (approved):

PR #212, opened 2022-12-16
[bug] Absolute path support for DscCompleteCheck

bkarstens:
## Description

Fixing a bug to properly read absolute paths in dsc files (on Windows). Absolute paths get introduced when using components within ext_deps and referencing them in the dsc like `$(MY_EXT_DEP_PATH)/MyComponent.inf`. The previous version was checking for `'MyPkg/My_extdep/MyComponent.inf' in 'C:\git\my_repo\MyPkg\My_extdep/MyComponent.inf'` and raising errors even though the component was referenced.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
On Windows:
+ Ran the DscCompleteCheck with components listed in the dsc like `$(MY_EXT_DEP_PATH)/MyComponent.inf`: DscComplete check no longer returns errors
+ Ran the DscCompleteCheck with normal components listed in the dsc: Does not return errors
+ Ran the DscCompleteCheck with normal and ext_dep components left out of the dsc: returned errors about both

## Integration Instructions

N/A

### DIFF
--- a/.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
+++ b/.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
@@ -130,7 +130,8 @@ class DscCompleteCheck(ICiBuildPlugin):
         return overall_status
 
     @staticmethod
-    def _module_in_dsc(inf: str, dsc: DscParser, Edk2pathObj) -> bool:
+    def _module_in_dsc(inf: str, dsc: DscParser, Edk2pathObj: Edk2Path) -> bool:
+
         """Checks if the given module (inf) is in the given dsc.
 
         Args:
--- a/.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
+++ b/.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
@@ -96,19 +96,9 @@ class DscCompleteCheck(ICiBuildPlugin):
         dp.SetInputVars(environment.GetAllBuildKeyValues())
         dp.ParseFile(wsr_dsc_path)
 
-        def module_in_dsc(inf: str) -> bool:
-            for module_type in (dp.ThreeMods, dp.SixMods, dp.OtherMods):
-                for module in module_type:
-                    if Path(module).is_absolute():
-                        module = Edk2pathObj.GetEdk2RelativePathFromAbsolutePath(module)
-                    if inf in module:
-                        return True
-            return False
-
         # Check if INF in component section
         for INF in INFFiles:
-            if not module_in_dsc(INF):
-
+            if not DscCompleteCheck._module_in_dsc(INF, dp, Edk2pathObj):
                 infp = InfParser().SetBaseAbsPath(Edk2pathObj.WorkspacePath)
                 infp.SetPackagePaths(Edk2pathObj.PackagePathList)
                 infp.ParseFile(INF)
@@ -138,3 +128,23 @@ class DscCompleteCheck(ICiBuildPlugin):
         else:
             tc.SetSuccess()
         return overall_status
+
+    @staticmethod
+    def _module_in_dsc(inf: str, dsc: DscParser, Edk2pathObj) -> bool:
+        """Checks if the given module (inf) is in the given dsc.
+
+        Args:
+            inf (str): The inf file to check for
+            dsc (DscParser): The parsed dsc file.
+            Edk2pathObj (Edk2Path): The path object capturing the workspace and package paths.
+
+        Returns:
+            bool: if the module is in the dsc.
+        """
+        for module_type in (dsc.ThreeMods, dsc.SixMods, dsc.OtherMods):
+            for module in module_type:
+                if Path(module).is_absolute():
+                    module = Edk2pathObj.GetEdk2RelativePathFromAbsolutePath(module)
+                if inf in module:
+                    return True
+        return False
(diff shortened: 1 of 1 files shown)


.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
```diff
@@ -58,7 +59,7 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
                 "DscPath not found in config file.  Nothing to check.")
             return -1
 
-        abs_pkg_path = Edk2pathObj.GetAbsolutePathOnThisSystemFromEdk2RelativePath( # MU_CHANGE
+        abs_pkg_path = Edk2pathObj.GetAbsolutePathOnThisSystemFromEdk2RelativePath(  # MU_CHANGE
```
makubacki, line 63: As far as I can tell, you just moved the comment here. Was that to align it on a multiple of 4 column?

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
```diff
@@ -82,7 +83,7 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
                 try:
                     tc.LogStdOut("Ignoring INF {0}".format(a))
                     INFFiles.remove(a)
-                except:
+                except Exception:
```
makubacki, line 87: Is this functionally related to the bug or being fixed as general best practice (avoid catching system exiting exceptions)?

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
```diff
@@ -95,11 +96,18 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
         dp.SetInputVars(environment.GetAllBuildKeyValues())
         dp.ParseFile(wsr_dsc_path)
 
+        def module_in_dsc(inf: str) -> bool:
```
apop5: This function is declared in the middle of RunBuildPlugin

Could it be moved into a function as part of the DscCompleteCheck() class?

makubacki: This can be a bit subjective but based on the length of the `RunBuildPlugin()` function, I could see a case made for it being more readable as a private class function.

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
bkarstens: That was probably my formatter adhering to [PEP8](https://peps.python.org/pep-0008/#inline-comments): "Inline comments should be separated by at least two spaces from the statement."

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
bkarstens: This change is not related to the bug, just best practices pointed out by my formatter. My formatter automatically made the except explicitly for `BaseException` and since it was pointed out, I changed it to just `Exception` since excepting `BaseException` is unnecessary and not best practice.

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
bkarstens: This is all subjective and I have no strong opinions on the matter. If you'd like it moved out to a private function of the `DscCompleteCheck` class, I have no objections. I can see arguments for readability either way.

I declared the function inside `RunBuildPlugin` because it a) made the arguments cleaner by capturing `dp` and b) was only needed for the `RunBuildPlugin` function.

bkarstens: Moved.

apop5 (approved):

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
```diff
@@ -131,3 +128,23 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
         else:
             tc.SetSuccess()
         return overall_status
+
+    @staticmethod
+    def _module_in_dsc(inf: str, dsc: DscParser, Edk2pathObj) -> bool:
```
makubacki: Was the type hint for `Edk2pathObj` left off intentionally?

```suggestion
    def _module_in_dsc(inf: str, dsc: DscParser, Edk2pathObj: Edk2Path) -> bool:

```

makubacki (approved):

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
makubacki: That's fine. Just wanted to double check.

.pytool/Plugin/DscCompleteCheck/DscCompleteCheck.py
```diff
@@ -9,6 +9,7 @@
 from edk2toollib.uefi.edk2.parsers.dsc_parser import DscParser
 from edk2toollib.uefi.edk2.parsers.inf_parser import InfParser
 from edk2toolext.environment.var_dict import VarDict
+from pathlib import Path
```
makubacki: For convenience.

```suggestion
from pathlib import Path
from edk2toollib.uefi.edk2.path_utilities import Edk2Path

```

PR #213, opened 2022-12-16
Increase Temp Buffer Size in FillMemoryMap() Routine

TaylorBeebe:
## Description

Platforms with many differences between the EFI and GCD memory maps can overflow the temp buffer allocated at the start of FillInMemoryMap(). This PR updates the size of the temp buffer to be quadruple the size of the input EFI memory map to accommodate extra entries. At the end of the routine, the contents of the temp buffer are copied into a new buffer which is strictly the required size and the temp buffer is freed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting to shell

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -1280,7 +1280,7 @@ FillInMemoryMap (
   NewMemoryMapStart = NULL;
 
   // Quadruple the size of the input memory map to accomodate extra entries
-  NewMemoryMapStart = AllocatePool ((*MemoryMapSize * 4) + *DescriptorSize);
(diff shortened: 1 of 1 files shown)


apop5:
Would it be better to create a PCD that determines the multiplier for additional entries and default it to 2 or 4?

Flickdm (approved):

os-d (approved):

TaylorBeebe:
> Would it be better to create a PCD that determines the multiplier for additional entries and default it to 2 or 4?

This seems like it would be confusing. The buffer allocated at the start of the routine is just temporary. At the end of the routine, the contents of the temp buffer are copied to a new buffer which is only the required size and the temp buffer is freed.

PR #225, opened 2023-01-09
Restore BSP Exception Stack Initialization

TaylorBeebe:
## Description

The BSP switch stack initialization was lost during the 202208 integration.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting to shell

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
+++ b/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
@@ -244,6 +244,7 @@ DxeMain (
   EFI_VECTOR_HANDOFF_INFO       *VectorInfoList;
   EFI_VECTOR_HANDOFF_INFO       *VectorInfo;
   VOID                          *EntryPoint;
+  VOID                          *Ptr; // MU_CHANGE
 
   //
   // Setup the default exception handlers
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
```diff
@@ -244,6 +244,7 @@ DxeMain (
   EFI_VECTOR_HANDOFF_INFO       *VectorInfoList;
   EFI_VECTOR_HANDOFF_INFO       *VectorInfo;
   VOID                          *EntryPoint;
+  VOID                          *Ptr; // MU_CHANGE
```
spbrogan: Instead of using VOID ptr it might be better to use something like
 EFI_HOB_GUID_TYPE       *GuidHob;

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
```diff
@@ -260,10 +261,17 @@ DxeMain (
...
   // Setup Stack Guard
   //
-  // MU_CHANGE START
+  // MU_CHANGE START: Check Memory Protection HOB
   // if (PcdGetBool (PcdCpuStackGuard)) {
-  //  Status = InitializeSeparateExceptionStacks (NULL, NULL);
-  //  ASSERT_EFI_ERROR (Status);
+  Ptr = GetFirstGuidHob (&gDxeMemoryProtectionSettingsGuid);
+  if ((Ptr != NULL) &&
+      (*((UINT8 *)GET_GUID_HOB_DATA (Ptr)) == (UINT8)DXE_MEMORY_PROTECTION_SETTINGS_CURRENT_VERSION) &&
```
spbrogan: And this would be more readable if you use the field names and structure types.

makubacki: I agree. It would also be a bit less error prone in case of a non-backward compatible structure change in the future (which shouldn't happen anyway).

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
TaylorBeebe: Updated

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
TaylorBeebe: Updated

makubacki (approved):

apop5:
does this need backported to 202202 as well?

apop5 (approved):

TaylorBeebe:
> does this need backported to 202202 as well?

No, this bug was introduced with the 202208 integration due to the changes to the exception handler setup logic flow

os-d (approved):

PR #230, opened 2023-01-24
Add extra RestoreTpl() call in DiskIo to maintain TPL raise/restore symmetry

joschock:
## Description

Adds a call to RestoreTpl() in DiskIo2ReadWriteDisk(). While the current implementation does not technically violate spec on raise/restore TPL, this extra call ensures symmetry between RaiseTpl and RestoreTpl calls, which makes analysis of TPL correctness simpler and permits certain non-standard TPL usages that some platforms require.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted a system using this driver for disk access, observed that symmetry between Raise/Restore TPL was maintained as expected.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
+++ b/MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIo.c
@@ -847,6 +847,7 @@ DiskIo2ReadWriteDisk (
   DISK_IO_SUBTASK         *Subtask;
   DISK_IO2_TASK           *Task;
   EFI_TPL                 OldTpl;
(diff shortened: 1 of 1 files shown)


makubacki (approved):

antklein (approved):

apop5 (approved):

makubacki:
@joschock do you just need this merged now?

joschock:
> @joschock do you just need this merged now?

Yes, that would be great, if things are good from MsCoreUefi side.

PR #245, opened 2023-02-05
Initial PEI memory buckets commit.  Integration with DXE will be adde…

kenlautner:
## Description

Added memory buckets to PEI for all memory types that persist into runtime. This is to reduce memory fragmentation for memory that is passed into the OS. Currently only EfiRuntimeServicesData is allocated in PEI.

MemoryBucketLib is added to keep track of PEI memory allocations of relevant types (currently only memory that persists into runtime) and consolidate their allocations to reduce fragmentation.

MemoryBucketLib keeps track of all the relevant information about the PEI memory buckets. It considers the newly added PeiMemoryBucket PCDs to decide what memory allocations to keep track of and possibly not be used at all. If all the PCDs remain at their default value of 0 PEI memory buckets will not be used and PEI memory allocation will function the same as it has before this change. Otherwise PEI memory buckets will be created for the memory types with non-zero PCD values.

These buckets are created in post-mem PEI and there is currently no support for pre-mem buckets.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Currently there is an exception occurring on Physical and Virtual platforms.  Will continue to fine tune to get memory behaving correctly.

## Integration Instructions

N/A

PR #267, opened 2023-02-15
Add PEI Memory Bucket Support

makubacki:
## Description

Added memory buckets to PEI for all memory types that are allocated
and persist into runtime. This is to reduce memory fragmentation for
memory that is passed into the OS.

This change updates PEI Core to contain "memory bucket" aka "memory
bin" logic that produces a memory type statistics HOB upon runtime
memory type allocation that is consumed by the DXE Core.

These memory buckets are allocated off the PEI heap and used as the
persistent memory location range through OS runtime.

DXE Core will consume the memory type statistics HOB produced if
such a runtime allocation occurs in PEI and use the runtime
memory ranges defined in the HOB for its memory bin locations
instead of allocating new ranges.

As before, the total allocation size per bucket must remain under
the total bucket size to prevent memory from being allocated
outside the PEI bucket range.

Testing has only been done in runtime memory buckets allocated
in post-memory PEI.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

In progress.

## Integration Instructions

N/A

Supersedes #245 and #167 

Co-authored-by: Ken Lautner <klautner@microsoft.com>

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain.h
+++ b/MdeModulePkg/Core/Dxe/DxeMain.h
@@ -49,6 +49,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Protocol/SmmBase2.h>
 #include <Protocol/PeCoffImageEmulator.h>
 #include <Guid/MemoryTypeInformation.h>
+#include <Guid/MemoryTypeStatistics.h>  // MU_CHANGE - Save memory allocations for the PEI memory buckets
 #include <Guid/FirmwareFileSystem2.h>
 #include <Guid/FirmwareFileSystem3.h>
 #include <Guid/HobList.h>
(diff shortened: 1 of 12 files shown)


makubacki:
I originally proposed https://github.com/microsoft/mu_feature_mm_supv/issues/85 as a near term, less risky alternative to carrying over runtime memory allocations of bucket memory types from PEI (what is done in this PR). After thinking about it more, I actually prefer that solution longer term as well.

1. This is only needed within affected platforms due to MM Supervisor comm and related buffers right now so a generic solution is not necessary.
2. Memory protections are not (currently) aligned between PEI and DXE. So, trying to persist PEI allocations into DXE leads to conditionalized accommodations to constructs like guard pages, pool alignment and buffering deltas, etc. that complicates DXE memory management logic.
3. DXE already carries runtime memory allocations into OS runtime. PEI is architecturally disjoint from DXE and DXE is the environment that directly supports loading OS runtime. Synchronizing runtime memory expectations between PEI and DXE is not as clean as DXE managing that alone.
4. The memory allocation needed for MM Supervisor PEI IPL can be boot services data so runtime allocation can be consolidated in DXE which leads to concentration of functional and security best practices there.

PR #269, opened 2023-02-17
Allow PciBus to tolerate a CRS response by ignoring the device

mikeytdisco:
# Preface

## Description

If there is a slow device on the PCI Bus, and the HostBridge is programmed to allow CRS, the slow device may return 0x0001 to inform the config space reader of the Vendor Id that pci device is not ready.  The current PciBus enumerator will treat 0001 as a valid Vendor Id, but it is not.  It indicates that all other config space is invalid. This code changes that operation to skip slow devices.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified that a platform without CRS still boots.

## Integration Instructions

N/A


spbrogan (approved):

os-d (approved):

kuqin12 (approved):

MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
```diff
@@ -57,8 +57,13 @@ PciDevicePresent (
                                   1,
                                   Pci
                                   );
-
-  if (!EFI_ERROR (Status) && ((Pci->Hdr).VendorId != 0xffff)) {
+  // MU_CHANGE Begin
+  // The host bridge may be programmed to accept CRS.  If the PCI device is slow, and CRS is enabled,
```
makubacki, line 62: Since this has 3 approvals, I wanted to ask before being able to do a bit more investigation into the call stack.

Does it make sense for a unique status code like `EFI_NOT_READY` to be returned? Right now `EFI_NOT_FOUND` could mean legitimately not found which is different than "not ready yet", right?

mikeytdisco: I think that depends upon the requirement for the change.  If the desire is supporting CRS, then that is a bigger change and would require more control (ie what to do about NOT_READY - spin? For how long?)  The short term fix is to allow CRS to be enabled, and fail gracefully.  The current problem is that PciBus treats 0x0001 as a valid ID, and uses 0xFFFF from all the other config registers as the resources required.  So, this fix makes 0x0001 behave the same as 0xFFFF and return NOT_FOUND.

makubacki: I think the spin delay could be sorted out, but it seems okay for the time being to prevent mishandling of devices.

It might be helpful to add a warning message if this case is encountered so someone debugging why a device was not found can see this case occurred.

makubacki (approved):

MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
mikeytdisco: Added debug warning message.

PR #292, opened 2023-03-01
Ensure DevicePath is FilePath Prior to Accessing PathName

TaylorBeebe:
## Description

If the Shell logic attempts to locate the startup.nsh script on a drive other than FS0:, it will check the path to the image for shell and the path to the file for shell which does not work correctly for the internal shell because the shell file path is located in the firmware volume. To ensure we don't try to read an invalid string, check the type and subtype of the device path header to ensure it is a filepath device type.

This bug causes a crash on SBSA when guard pages are active because startup.nsh is located in the virtual drive mapped to FS2:.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booting on Q35 and SBSA

## Integration Instructions

N/A

### DIFF
--- a/ShellPkg/Application/Shell/Shell.c
+++ b/ShellPkg/Application/Shell/Shell.c
@@ -1259,8 +1259,14 @@ LocateStartupScript (
     }
 
     InternalEfiShellSetEnv (L"homefilesystem", StartupScriptPath, TRUE);
+    // MU_CHANGE START: Ensure FileDevicePath is a FILEPATH_DEVICE_PATH prior to accessing PathName
+    if ((FileDevicePath->Type == MEDIA_VENDOR_DP) && (FileDevicePath->SubType == MEDIA_FILEPATH_DP)) {
+      StartupScriptPath = StrnCatGrow (&StartupScriptPath, &Size, ((FILEPATH_DEVICE_PATH *)FileDevicePath)->PathName, 0);
+    } else {
+      StartupScriptPath = StrnCatGrow (&StartupScriptPath, &Size, L"\\", 0);
+    }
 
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

apop5 (approved):

mikeytdisco (approved):

ShellPkg/Application/Shell/Shell.c
```diff
@@ -1259,9 +1259,15 @@ LocateStartupScript (
     }
 
     InternalEfiShellSetEnv (L"homefilesystem", StartupScriptPath, TRUE);
+    // MU_CHANGE START: Ensure FileDevicePath is a FILEPATH_DEVICE_PATH prior to accessing PathName
+    if ((FileDevicePath->Type == MEDIA_VENDOR_DP) && (FileDevicePath->SubType == MEDIA_FILEPATH_DP)) {
```
makubacki, line 1263: Just wanted to double check that you confirmed the type and subtype. I seem to recall type=`MEDIA_DEVICE_PATH` and subtype=`MEDIA_FILEPATH_DP` being used more often for file device paths.

makubacki (approved):

ShellPkg/Application/Shell/Shell.c
mikeytdisco: That is correct.  The code assumed the device path node is such as evidenced by the cast to FILEPATH_DEVICE_PATH.  It is a FILEPATH_DEVICE_PATH on "normal" systems that do not expose FV's as mapped file systems.  If you map FV's, and boot to the shell from an FvGuid/FvName, the DEVIDE_PATH node is a FvGuid/FvName node, and casting that to a FILEPATH_DEVICE_PATH causes StrCat to run into the weeds.

PR #326, opened 2023-03-20
Add Volatile Keyword to NVMe CQs and SQs

### LINKED ISSUE #324: [Bug]: NvmExpressDxe and NvmExpressPei are missing 'volatile' for pointer to struct located in memory updated by Nvme controller

### Current Behavior

From NvmExpressDxe\NvmExpressPassthru.c, function NvmExpressPassThru:
```
  while (EFI_ERROR (gBS->CheckEvent (TimerEvent))) {
    if (Cq->Pt != Private->Pt[QueueId]) {
      Status = EFI_SUCCESS;
      break;
    }
  }
```
Before this 'while' loop, the code has initiated an NVMe transaction via a submission queue.
The 'while' loop is looking for a state change of 'Cq->Pt' indicating that the transaction is complete.

'Cq' is a pointer to 'NVME_CQ' which is a structure that is updated by the Nvme controller. This structure should have the 'volatile' keyword to ensure that the memory is re-read within the 'while' loop.
Without 'volatile', the compiler is only reading the 'Cq->Pt' once outside of the 'while' loop and then doing register compares within the 'while' loop.

It has been observed (on AARCH64 target hardware) that in some cases, 'Cq->Pt' has not yet been updated by the NVMe controller by the time the code reaches the 'while' loop. In those cases, the 'while' loop will only terminate due to timeout because the external memory is no longer being read within the 'while' loop.

A similar code construct is in NvmExpressPei\NvmExpressPassThru.c and there may be other areas where 'volatile' is missing.

### Expected Behavior

Using 'volatile' with declarations of pointer to NVME_CQ results in code that re-reads the memory within the 'while' loop ensuring that the updated state of 'Cq->Pt' is detected.

### Steps To Reproduce

The problem is intermittent and may depend on underlying hardware and toolchain optimization flags.

Issue any commands/procedures to read from an NVMe drive.
Observe NVMe command timeout errors through logs and addition of debug prints in NvmExpressPassThru.c.
Code inspection and analysis of assembly listing of execution binaries with and without 'volatile' keyword.

### Build Environment

```markdown
- OS(s): Windows 11 WSL2
- Tool Chain(s): GCC5
- Targets Impacted: ALL
```

### Version Information

```text
Release/202208
```

### Urgency

Medium

Someone else needs to fix it

Maintainer feedback requested

### Anything else?

There may be other modules besides NvmExpressDei and NvmExpressPei that are missing the 'volatile' keyword.

os-d:
## Description

Fixes bug #324. There are two NVMe CQs that get updated by hardware. In these two files, we do not have a volatile keyword on these CQ structs, so when we run them in a loop to check if they have changed, we do not continue to fetch the HW updated queues. Adding the volatile keyword allows us to continue to check the queues during the timer period (otherwise the timer period elapses without being able to operate on queues with descriptors in them.

I checked for other instances of this pattern in the NVMe code that are not reported in this bug, I only found the fixed instances in NvmExpressPassthru.c and NvmExpressPeiPassThru.c. There is a similar if statement in NvmExpress.c, but the CQ gets updated in that loop and this is just a timer callback, process what we found and start the timer again.

This PR also updates SQs to have the volatile keyword, per PR feedback, to ensure that HW accesses are done for each read of the SQ. Current code patterns do not appear to have an issue with reading SQs, but adding the volatile keyword will help ensure that patterns are not introduced that have the same issue as the CQs, as well as maintaining the same pattern with respect to HW queues.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by bug submitter on an ARM64 platform.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
@@ -516,7 +516,7 @@ NvmExpressPassThru (
   EFI_STATUS                     PreviousStatus;
   EFI_PCI_IO_PROTOCOL            *PciIo;
   NVME_SQ                        *Sq;
-  NVME_CQ                        *Cq;
+  volatile NVME_CQ               *Cq;      // MU_CHANGE: Add volatile so that timer loop below is getting updated CQ
   UINT16                         QueueId;
   UINT16                         QueueSize;
   UINT32                         Bytes;
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
@@ -940,7 +940,7 @@ NvmExpressPassThru (
       // Dump every completion entry status for debugging.
       //
       DEBUG_CODE_BEGIN ();
-      NvmeDumpStatus ((NVME_CQ)Cq);
+      NvmeDumpStatus (Cq);
       DEBUG_CODE_END ();
     }
 
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
@@ -18,7 +18,7 @@
 **/
 VOID
 NvmeDumpStatus (
-  IN NVME_CQ  *Cq
+  IN volatile NVME_CQ  *Cq     // MU_CHANGE: Add volatile keyword to NVME_CQ
   )
 {
   DEBUG ((DEBUG_VERBOSE, "Dump NVMe Completion Entry Status from [0x%x]:\n", Cq));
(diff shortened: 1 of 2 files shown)


makubacki (approved):

MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
```diff
@@ -516,7 +516,7 @@ NvmExpressPassThru (
   EFI_STATUS                     PreviousStatus;
   EFI_PCI_IO_PROTOCOL            *PciIo;
   NVME_SQ                        *Sq;
```
apop5: does Sq have the same protentional problem? Both Sq and Cq would be mapped into the Nvme's MMIO space, so both could theoretically be changed independently.

kuqin12 (approved):

apop5 (approved):

MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressPassthru.c
os-d: Sq shouldn't have this problem because we don't try to read anything from it (it is our submission queue, so we post some descriptors to the Sq, ring the doorbell, then wait for completions in the Cq). There is a possibility that we think the Sq has free descriptors when it is full, but I believe when we check if the Sq is full or not, we are the only ones submitting to it.

os-d: Per offline discussion with @apop5, added the volatile keyword for SQs as well, more description in edited PR description.

PR #335, opened 2023-03-28
Adding MM Supervisor debugging print bit

### LINKED ISSUE #99: [Feature]: Control Syscall Enter/Exit Message Separately from DEBUG_VERBOSE
### Feature Overview

The `SysCallDispatcher()` function in [MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c](https://github.com/microsoft/mu_feature_mm_supv/blob/8b8aeef82a98410e61a632f2eaf41dd161087501/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c#L245-L256) prints a message upon every entry and exit to the function.

https://github.com/microsoft/mu_feature_mm_supv/blob/8b8aeef82a98410e61a632f2eaf41dd161087501/MmSupervisorPkg/Core/PrivilegeMgmt/SyscallDispatcher.c#L245-L256

The print error level is `DEBUG_VERBOSE` but the message frequency greatly exceeds the number of messages typically printed in a single function at that level. It effectively makes a boot with `DEBUG_VERBOSE` level set impossible to use in a reasonable developer cycle.

On one system, it has printed 86,000+ times on a single boot (before being powered off). On the UART at 115200, this has consumed 39+ minutes and the text dominates the serial log.

This is useful information, but it's not general verbose information. It's extreme detail for a very specific scenario that I believe is best handled with a special flag.

This request does not specify how this flag is implemented.

> Note: This is submitted as a feature request, but could also be considered a bug.

### Solution Overview

A specific solution is not given. Options could include:

- Define a print error level for "MM Supervisors" (or something more relevant) in `gEfiMdePkgTokenSpaceGuid.PcdFixedDebugPrintErrorLevel`

- Define a print error level PCD in 'MmSupervisorPkg` and add a bit for the sys call dispatcher

- Define another mechanism like a dedicated (e.g. boolean) PCD to control whether this message is printed

### Alternatives Considered

1. Override the `PrintErrorLevel` on the MmSupervisorCore module to remove `DEBUG_VERBOSE`

Not desirable as other verbose messages in the module are missed.

2. Comment out this message.

Not desirable since source code modification should not be necessary for this specific case.

### Urgency

Medium

### Are you going to implement the feature request?

Someone else needs to implement the feature

Maintainer feedback requested

### Anything else?

_No response_

kuqin12:
# Preface

## Description

This change is made to resolve [this issue](https://github.com/microsoft/mu_feature_mm_supv/issues/99).

Logs from certain MM supervisor logs are causing due to some highly repetitive prints, making the system barely bootable. This change will add a new bit to the debug mask macros to enable these category of prints individually.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change only expands bitmask definition, non-functional.

## Integration Instructions

N/A.

### DIFF
--- a/MdePkg/Include/Library/DebugLib.h
+++ b/MdePkg/Include/Library/DebugLib.h
@@ -51,6 +51,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define DEBUG_CACHE     0x00200000  // Memory range cachability changes
 #define DEBUG_VERBOSE   0x00400000  // Detailed debug messages that may
                                     // significantly impact boot performance
+#define DEBUG_MM_SUPV   0x00800000  // MU_CHANGE: This is separately defined to allow repetetive
+                                    // supervisor debugging prints to be enabled individually.
 #define DEBUG_ERROR  0x80000000     // Error
 
 //
--- a/MdePkg/MdePkg.dec
+++ b/MdePkg/MdePkg.dec
@@ -2218,6 +2218,7 @@
   #  BIT20 - Global Coherency Database changes message.<BR>
   #  BIT21 - Memory range cachability changes message.<BR>
   #  BIT22 - Detailed debug message.<BR>
+  #  BIT23 - Added for repetetive MM supervisor debugging prints.<BR> # MU_CHANGE
   #  BIT31 - Error message.<BR>
   # @Prompt Fixed Debug Message Print Level.
   gEfiMdePkgTokenSpaceGuid.PcdFixedDebugPrintErrorLevel|0xFFFFFFFF|UINT32|0x30001016


MdePkg/Include/Library/DebugLib.h
```diff
@@ -51,6 +51,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define DEBUG_CACHE     0x00200000  // Memory range cachability changes
 #define DEBUG_VERBOSE   0x00400000  // Detailed debug messages that may
                                     // significantly impact boot performance
+#define DEBUG_MM_SUPV   0x00800000  // MU_CHANGE: This is separately defined to allow repetetive
```
makubacki, line 54: The debug print driving this change can print so frequently that it will dominate the log and add considerable time to the boot (far more than other "verbose" messages).

So, something that should be considered is, are other messages bundled under the same flag missed if the frequency of the syscall message is a deterrent to use the flag?

What other (if any) debug prints in the MM Supervisor would you expect to use this level?

kuqin12: I was expecting the mm entrypoint prints to be added under this flag as well. But on the other hand, it may just qualify a verbose flag.

makubacki: Alright. The bit can of course represent anything within its scope over time (i.e., be added to new MM Supervisor debug prints), so I want to make sure that the scope of "MM Supervisor" is suitable at this time.

Main concern is that "MM Supervisor" is a large scope and "syscall dispatch" messages are a small/specific piece of that so everything else in the MM Supervisor would need to be covered under pre-existing bits (like info, warn, error, verbose, etc) leaving this to represent the syscall messages for it to be useful in containing the message output.

If we go with this approach, I think this is one change that really needs to go into TianoCore first to avoid a bit assignment conflict and ensure it matches the upstream without conflict.

kuqin12: Synced offline, and determined that this is not the proper to proceed and could cause unnecessary ripples when upstreaming to the edk2. Will change the supervisor code to cover this issue instead.

PR #346, opened 2023-04-07
SetupBrowserDxe: Initialize the variable 'BrowserStorage'

pohanch:
# Preface

Fix potential error from CodeQL fixes ([commit df0aadd](https://github.com/microsoft/mu_basecore/commit/df0aadd15f3b5beade40728b87fd0765f58bec5b)) that resulted in an uninitialized variable attempting to be used.

## Description

Initialize the variable 'BrowserStorage' to avoid it's used but uninitialized when the logic route through an if statement.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built several times with the change locally, and the system boots.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
+++ b/MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
@@ -438,8 +438,9 @@ CreateStorage (
   EFI_GUID         *StorageGuid;
   CHAR8            *StorageName;
 
-  UnicodeString = NULL;
-  StorageName   = NULL;
+  UnicodeString  = NULL;
+  StorageName    = NULL;
+  BrowserStorage = NULL; // MU_CHANGE - Initialize variable that might not be updated due to error checking
   switch (StorageType) {
     case EFI_HII_VARSTORE_BUFFER:
       StorageGuid = (EFI_GUID *)(CHAR8 *)&((EFI_IFR_VARSTORE *)OpCodeData)->Guid;
--- a/MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
+++ b/MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
@@ -535,11 +535,10 @@ ErrorExit:
     FreePool (BrowserStorage);
   }
 
-  if (Storage->ConfigRequest != NULL) {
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
```diff
@@ -438,8 +438,9 @@ CreateStorage (
   EFI_GUID         *StorageGuid;
   CHAR8            *StorageName;
 
-  UnicodeString = NULL;
-  StorageName   = NULL;
+  UnicodeString  = NULL;
+  StorageName    = NULL;
+  BrowserStorage = NULL; // MU_CHANGE - Initialize variable that might not be updated due to error checking
```
os-d, line 443: This is a good catch, it looks like we need to do the same for Storage, as if we fail on UnicodeString being NULL, we would attempt to check if Storage->ConfigRequest is NULL, free it if not, then do the same with Storage. So can you also update Storage to be initialized to NULL and update the error checking to check if Storage is NULL first, then if Storage->ConfigRequest is NULL, free it, then free Storage?

os-d (changes requested):

MdeModulePkg/Universal/SetupBrowserDxe/IfrParse.c
os-d: Sorry, misread the error flow. We shouldn't need to initialize Storage, but we should change the order of the error checking to check Storage before Storage->ConfigRequest (and only check inside Storage if it is non-null).

os-d:
```c
   if (Storage != NULL) {
     if (Storage->ConfigRequest != NULL) {
      FreePool (Storage->ConfigRequest);
    }
    FreePool (Storage);
  }
```

pohanch: Thank you. Changed the checking flow.

os-d (approved):

kuqin12 (approved):

PR #347, opened 2023-04-07
Infinite boot retries

apop5:
## Description

This is a draft of infinite boot retries based on a newly created PCD.  When true, the system will never stop retrying the boot options. 

PCD default is FALSE to match existing functionality. 

- [X] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## Integration Instructions

n/a

### DIFF
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -1272,6 +1272,14 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportAlternativeQueueSize|FALSE|BOOLEAN|0x40000151
   # MU_CHANGE [END]
 
+  # MU_CHANGE [BEGIN] - Support infinite boot retries
+  # #  If enabled, Bds will retry boot options indefinitely
+  # #   TRUE  - Efi boot options will be retried indefinitely.
+  # #   FALSE - Efi boot options will not be retried.
+  # # @Prompt Enable UEFI Stack Guard.
+  gEfiMdeModulePkgTokenSpaceGuid.PcdSupportInfiniteBootRetries|FALSE|BOOLEAN|0x40000152
+  # MU_CHANGE [END]
+
 [PcdsFixedAtBuild, PcdsPatchableInModule]
   ## Dynamic type PCD can be registered callback function for Pcd setting action.
   #  PcdMaxPeiPcdCallBackNumberPerPcdEntry indicates the maximum number of callback function
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -1272,11 +1272,12 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportAlternativeQueueSize|FALSE|BOOLEAN|0x40000151
   # MU_CHANGE [END]
 
-  # MU_CHANGE [BEGIN] - Support infinite boot retries
-  # #  If enabled, Bds will retry boot options indefinitely
+  # MU_CHANGE [BEGIN] - Support indefinite boot retries
+  # # Some platforms require that all EfiLoadOptions are retried until one of the options 
+  # # succeeds. When True, this Pcd will force Bds to retry all the valid EfiLoadOptions 
+  # # indefinitely until one of the options succeeds. 
   # #   TRUE  - Efi boot options will be retried indefinitely.
   # #   FALSE - Efi boot options will not be retried.
-  # # @Prompt Enable UEFI Stack Guard.
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportInfiniteBootRetries|FALSE|BOOLEAN|0x40000152
   # MU_CHANGE [END]
 
--- a/MdeModulePkg/Universal/BdsDxe/BdsEntry.c
+++ b/MdeModulePkg/Universal/BdsDxe/BdsEntry.c
@@ -418,6 +418,9 @@ BootBootOptions (
     PlatformBootManagerProcessBootCompletion (&BootOptions[Index]);        // MSCHANGE 00076 - record boot status
 
     // MU_CHANGE [BEGIN] - Support infinite boot retries
+    //  Changes for PcdSupportInfiniteBootRetries are meant to minimize upkeep in mu repos.
+    //   If/when upstreaming this change, refactoring calling loop in BdsEntry() would be
+    //   better location.
     if (!PcdGetBool (PcdSupportInfiniteBootRetries)) {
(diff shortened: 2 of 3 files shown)


kuqin12 (approved):

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1272,6 +1272,14 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportAlternativeQueueSize|FALSE|BOOLEAN|0x40000151
   # MU_CHANGE [END]
 
+  # MU_CHANGE [BEGIN] - Support infinite boot retries
+  # #  If enabled, Bds will retry boot options indefinitely
+  # #   TRUE  - Efi boot options will be retried indefinitely.
+  # #   FALSE - Efi boot options will not be retried.
+  # # @Prompt Enable UEFI Stack Guard.
```
makubacki: Drop the `@prompt Enable UEFI Stack Guard.` line.

```suggestion

```

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1272,6 +1272,14 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdSupportAlternativeQueueSize|FALSE|BOOLEAN|0x40000151
   # MU_CHANGE [END]
 
+  # MU_CHANGE [BEGIN] - Support infinite boot retries
+  # #  If enabled, Bds will retry boot options indefinitely
```
makubacki: I think this PCD needs more documentation. Mainly explaining why this PCD exists. MdeModulePkg is one of the most common package dependencies so many people will see this so I think it should be written to precisely explain why the PCD exists and when it should be used.

MdeModulePkg/MdeModulePkg.dec
makubacki, line 1: Ideally, a feature request issue should be filed explaining this option and linked to the PR (in the PR description.

makubacki, line 1: Also, would be nice to add something like the `type:enhancement` label. That will recognize the change as a "feature" level change which will cause it to be categorized as a "feature" in the release notes and roll the minor version appropriately.

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -417,16 +417,31 @@ BootBootOptions (
...
+      if ((BootManagerMenu != NULL) && (BootOptions[Index].Status == EFI_SUCCESS)) {
+        EfiBootManagerBoot (BootManagerMenu);
+        break;
+      }
+
+      // MU_CHANGE [BEGIN]- Support infinite boot retries
+    }
+
+    if (PcdGetBool (PcdSupportInfiniteBootRetries)) {
+      if (Index == (BootOptionCount -1)) {
```
makubacki:
```suggestion
      if (Index == (BootOptionCount - 1)) {
```

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -417,16 +417,31 @@ BootBootOptions (
...
+        Index = 0;
```
makubacki: This is not equivalent to assigning `Index` to `0` in the loop initialization step. On the next iteration, `Index` will be incremented and then compared against `BootOptionCount` with the value of `1`.

Is that intentional?

MdeModulePkg/MdeModulePkg.dec
apop5: good point.

apop5:
Fixes #360

MdeModulePkg/MdeModulePkg.dec
apop5: #360 created.

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
apop5: It was an oversight. Updated to set Index to (UINTN)-1.

MdeModulePkg/MdeModulePkg.dec
apop5: Removed.

MdeModulePkg/MdeModulePkg.dec
apop5: Updated documentation to try to better describe the usage and intention.

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
apop5: Updated to match.

makubacki (approved):

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -417,16 +417,31 @@ BootBootOptions (
...
+        EfiBootManagerBoot (BootManagerMenu);
+        break;
+      }
+
+      // MU_CHANGE [BEGIN]- Support infinite boot retries
+    }
+
+    if (PcdGetBool (PcdSupportInfiniteBootRetries)) {
+      if (Index == (BootOptionCount - 1)) {
+        Index = (UINTN)-1;
```
spbrogan, line 447: Please write a quick comment that setting to -1 so that for loop incrementing logic will set Index to zero on next iteration.  It will avoid someone searching to see if -1 is some kind of flag only to realize it is a hack to get index back to 0.  :)

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -417,16 +417,31 @@ BootBootOptions (
...
+      // supports boot manager menu, and if firmware is configured to boot in an
+      // interactive mode, the boot manager will stop processing the BootOrder variable and
+      // present a boot manager menu to the user.
+      //
+      if ((BootManagerMenu != NULL) && (BootOptions[Index].Status == EFI_SUCCESS)) {
+        EfiBootManagerBoot (BootManagerMenu);
+        break;
+      }
+
+      // MU_CHANGE [BEGIN]- Support infinite boot retries
```
spbrogan, line 438: Please add comment that when/if this change is upstreamed the loop should be redesigned so that such a hacky change isn't needed.  The current hacky solution makes sense in that the MU_CHANGE is trying to minimize the differences.

spbrogan (approved):

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
apop5: Will do.

PR #352, opened 2023-04-12
Converge on 4k Section Alignment for AARCH64 Builds

### LINKED ISSUE #386: [Bug]: AARCH64 LD Can Misinterpret Multiple common-page-size Parameters

### Current Behavior

It has been observed that on aarch64 package builds, GenFw can fail at this line: https://github.com/microsoft/mu_basecore/blob/1bd4b3bfd657ad5a3940d072be6540f7c9c23c5d/BaseTools/Source/C/GenFw/Elf64Convert.c#L1540.

What appears to be happening is that ld is interpreting the multiple common-page-size parameters passed to it in a different order, sometimes at 0x20, sometimes at 0x1000, which breaks GenFw's expectation of alignment. It does not appear to be valid to pass common-page-size to ld multiple times.

There are complicating factors: the same toolchain will work on the same code, then it will break the next day, with no code changed. Manually editing tools_def.txt (or tools_def.template and regenerating tools_def.txt) to set all common-page-size calls for AARCH64 GCC to 0x1000 causes the issue to go away, however, on a system that is not experiencing the issue, setting all the common-page-size parameters to 0x20 does not repro the issue.

The issue does repro occasionally across multiple machines, both local dev boxes of different devs (and on different packages) and on pipelines. So far, this has only been observed in non-public code; I am attempting to get a public repro, however, with the unpredictability of getting a repro, I have been unsuccessful so far.

At the very least, not passing common-page-size multiple times would seem to be a step in the right direction, although further investigation here seems warranted.

### Expected Behavior

I expect GenFw to not fail and properly be able to convert ELF files to PE/COFF.

### Steps To Reproduce

Unfortunately, this is not easily reproducible, but the signature is clear.

### Build Environment

```markdown
- OS(s): Linux
- Tool Chain(s): GCC
- Targets Impacted: AARCH64
```

### Version Information

```text
Version v2022080001.4.2
```

### Urgency

High

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

os-d:
# Preface

## Description

Currently tools_def.template sets two different common-page-sizes for GCC/Clang AARCH64 builds. This is not a supported configuration. This leads to ambiguity and relying on the toolchain to interpret the arguments in a certain way, which they are not guaranteed to do. This leads to the case where the same code can fail to be aligned to the required 4k boundary because the incorrect common-page-size was chosen.

This PR removes the incorrect common-page-size for AARCH64 GCC/Clang builds and only uses the 4kb alignment to remove the ambiguity and fix errors seen where the wrong common-page-size gets chosen by the toolchain. This PR also removes some of the duplicate instances of setting the common-page-size that exist to try to reduce issues in the future if it is changed. However, it was not attempted to reduce all of the duplicated instances in order to not miss edge cases and instead only the clearly duplicated cases where a new macro expanded another macro that contained common-page-size and then set common-page-size directly afterwards.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by running in a local CI build that was failing due to the local GCC toolchain choosing the non-4k alignment argument passed to it and GenFw failing due to incorrect section alignment. With this fix, the CI build passes.

## Integration Instructions

To integrate this change, pull the latest mu_basecore and delete the top level Conf directory. This will get regenerated from the updated mu_basecore tools_def.template.


BaseTools/Conf/tools_def.template
```diff
@@ -2042,8 +2042,8 @@ DEFINE GCC_DLINK2_FLAGS_COMMON     = -Wl,--script=$(EDK_TOOLS_PATH)/Scripts/GccB
 DEFINE GCC_IA32_X64_DLINK_COMMON   = DEF(GCC_DLINK_FLAGS_COMMON) --gc-sections
 DEFINE GCC_ARM_AARCH64_DLINK_COMMON= -Wl,--emit-relocs -nostdlib -Wl,--gc-sections -u $(IMAGE_ENTRY_POINT) -Wl,-e,$(IMAGE_ENTRY_POINT),-Map,$(DEST_DIR_DEBUG)/$(BASE_NAME).map
 DEFINE GCC_ARM_DLINK_FLAGS         = DEF(GCC_ARM_AARCH64_DLINK_COMMON) -z common-page-size=0x20 -Wl,--pic-veneer
-DEFINE GCC_AARCH64_DLINK_FLAGS     = DEF(GCC_ARM_AARCH64_DLINK_COMMON) -z common-page-size=0x20
-DEFINE GCC_ARM_AARCH64_ASLDLINK_FLAGS = -Wl,--defsym=PECOFF_HEADER_SIZE=0 DEF(GCC_DLINK2_FLAGS_COMMON) -z common-page-size=0x20
+DEFINE GCC_AARCH64_DLINK_FLAGS     = DEF(GCC_ARM_AARCH64_DLINK_COMMON) -z common-page-size=0x1000
```
kuqin12, line 2045: Will we still see this flag twice on the command line when using GCC? I saw that CLANG removed one instance though.

kuqin12 (approved):

apop5 (approved):

makubacki (approved):

BaseTools/Conf/tools_def.template
os-d: We will see it twice, but I was hesitant to remove it from more places without chasing it down completely, as I wasn't sure if there were cases where we would only use one of the macros that sets the common-page-size

os-d:
Closing as the issue does not currently reproduce. Continuing to do investigation into the proper tools_def changes here.

os-d:
I am reopening this PR because the behavior has continued to be seen intermittently. A Tianocore bugzilla has been filed for this issue, but for now, this seems to workaround the issue.

makubacki:
> I am reopening this PR because the behavior has continued to be seen intermittently. A Tianocore bugzilla has been filed for this issue, but for now, this seems to workaround the issue.

Can you file an issue in mu_basecore and link it to this PR? In the issue include a link to the BZ.

makubacki:
> > I am reopening this PR because the behavior has continued to be seen intermittently. A Tianocore bugzilla has been filed for this issue, but for now, this seems to workaround the issue.
> 
> Can you file an issue in mu_basecore and link it to this PR? In the issue include a link to the BZ.

Please put some of the info shared offline (as appropriate) into the issue as well to give context if it's looked up in the future.

os-d:
Sorry, I missed this. I will file an issue and link it.

PR #354, opened 2023-04-13
Remove UefiBootServicesTableLib from inclusion file

kuqin12:
# Preface

## Description

The inclusion file change will potentially change the library linkage of `UefiBootServicesTableLib`, which could be an incompatible instance compared to the existing consumers.

This change temporarily removes this inclusion until all unit tests from the consumers migrated their unit tests to the common implementation.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on consumer packages that uses `UefiBootServicesTableLib`.

## Integration Instructions

For packages uses `UefiBootServicesTableLib|UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLib.inf` for unit tests, one needs to update the host test dsc to include the entry. Otherwise, no changes needed.

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -20,7 +20,9 @@
   UnitTestLib|UnitTestFrameworkPkg/Library/UnitTestLib/UnitTestLibCmocka.inf
   DebugLib|UnitTestFrameworkPkg/Library/Posix/DebugLibPosix/DebugLibPosix.inf
   MemoryAllocationLib|UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.inf
-  UefiBootServicesTableLib|UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLib.inf
+  # MU_CHANGE [BEGIN] - Disable UefiBootServicesTableLib until all unit tests converge to this common instance
+  # UefiBootServicesTableLib|UnitTestFrameworkPkg/Library/UnitTestUefiBootServicesTableLib/UnitTestUefiBootServicesTableLib.inf
+  # MU_CHANGE [BEGIN] - Disable UefiBootServicesTableLib until all unit tests converge to this common instance
 
 [BuildOptions]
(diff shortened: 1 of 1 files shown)


os-d (approved):

makubacki:
Library instances in include files are typically meant to be included at the top of a DSC and overridden as needed by the package. Is that a problem here?

kuqin12:
> Library instances in include files are typically meant to be included at the top of a DSC and overridden as needed by the package. Is that a problem here?

The issue is that if the include file is more specific of a type than the end consumer, the more specific instance will supersede. For this case, the consuming dscs just have `[LibraryClasses]` for the `UefiBootServicesLib`

makubacki:
> > Library instances in include files are typically meant to be included at the top of a DSC and overridden as needed by the package. Is that a problem here?
> 
> The issue is that if the include file is more specific of a type than the end consumer, the more 

> > Library instances in include files are typically meant to be included at the top of a DSC and overridden as needed by the package. Is that a problem here?
> 
> The issue is that if the include file is more specific of a type than the end consumer, the more specific instance will supersede. For this case, the consuming dscs just have `[LibraryClasses]` for the `UefiBootServicesLib`

We can make the change but some notes:

1. UnitTestFrameworkPkgHost.dsc.inc has several common library classes assigned to instances (e.g. `BaseLib`, `DebugLib`, `CpuLib`, etc.) in `[LibraryClasses.common.HOST_APPLICATION]` so those would also be a problem if the platform wanted to override them.
2. The platform should be aware of that the include DSC specifies its library class instances in `[LibraryClasses.common.HOST_APPLICATION]` and specify libraries for that module type in a module-specific section to prevent future occurrences of this (new library instances being added to the DSC include file they don't override).
3. We should reduce Mu deltas and get back to parity with edk2 when possible.

makubacki (approved):

kuqin12:
> > Library instances in include files are typically meant to be included at the top of a DSC and overridden as needed by the package. Is that a problem here?
> 
> The issue is that if the include file is more specific of a type than the end consumer, the more specific instance will supersede. For this case, the consuming dscs just have `[LibraryClasses]` for the `UefiBootServicesLib`

Agreed. An issue is logged here: https://github.com/microsoft/mu_basecore/issues/356. Will remove this deviation after resolving the consuming dsc issues.

PR #372, opened 2023-04-24
Updated Crypto Tests

kenlautner:
## Description

Updated the crypto tests to reference the crypto functionality PCDs before attempting to test them.  This will only test the enabled functionality for the different crypto flavors of the shared crypto binaries.  Tests for functionality that are not enabled are skipped and labelled that way in the report.  

Additionally, a new PCD dsc file was created for test purposes where all the Crypto PCDs are enabled.  This should only be used for testing.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran the test on QemuQ35 and Adl physical platforms.  Checked the different flavors to make sure tests were passed/failed/skipped as expected without any problems.

## Integration Instructions

When building the shell app or  host based unit test include the relevant PCDs that correspond to the flavor of Shared Crypto used for DXE.


CryptoPkg/Test/Crypto.pcd.ALL.inc.dsc
```diff
@@ -0,0 +1,145 @@
+# List of Crypto PCDs with all functionality enabled.  This should only be used for testing.
```
os-d, line 1: I think this should be `*.dsc.inc` to match the usual naming convention?

Typically I see `*.dsc.inc` files with the section header defined in them, so no matter where they are included (basically) it is valid. Now, this being said, you are adding it not in a proper PcdsFixedAtBuild section, but inside of an override for a module, so perhaps this is not valid here.

os-d (approved):

kuqin12 (approved):

CryptoPkg/Test/Crypto.pcd.ALL.inc.dsc
kenlautner: All the other flavor PCD include files are *inc.dsc so I followed that convention.  I have no problems switching it if that's what people prefer.

apop5 (approved):

PR #393, opened 2023-05-06
Google bad example

kuqin12:
# Preface

## Description

<_Please include a description of the change and why this change was made._>

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

<_Please describe the test(s) that were run to verify the changes._>

## Integration Instructions

### DIFF
--- a/UnitTestFrameworkPkg/Library/GoogleTestBadExamples/GoogleTestBadExamplesSrc.c
+++ b/UnitTestFrameworkPkg/Library/GoogleTestBadExamples/GoogleTestBadExamplesSrc.c
@@ -1,0 +1,59 @@
+/** @file
+  Bad example for Google Test.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Uefi/UefiBaseType.h>
+#include <PiDxe.h>
+
+#include <Guid/ZeroGuid.h>
+#include <Protocol/Policy.h>
+
+#include <Library/DebugLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+
+/**
+  Get policy protocol from our database.
+
+  @retval EFI_SUCCESS       Policy is located successfully.
+  @retval EFI_ERROR         Failed to locate policy protocol.
+**/
+EFI_STATUS
+EFIAPI
+GetPolicy (
+  VOID
+  )
+{
+  EFI_STATUS       Status;
+  UINT16           ActualPolicySize;
+  VOID             *TestPolicy = NULL;
+  UINT64           PolicyAttribute;
+  POLICY_PROTOCOL  *PolicyProtocol;
+
+  Status = gBS->LocateProtocol (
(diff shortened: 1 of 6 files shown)


UnitTestFrameworkPkg/Library/GoogleTestBadExamples/Googletest/GoogleTestBadExamplesUT.inf
```diff
@@ -0,0 +1,32 @@
...
+  VERSION_STRING  = 1.0
+
+[Sources]
+  GoogleTestBadExamplesUT.cpp
+  ../GoogleTestBadExamplesSrc.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+  PolicyServicePkg/PolicyServicePkg.dec
```
makubacki: Is there a plan to upstream PolicyServicePkg?

One of the problems with keeping it in mu_basecore long term is that it extends its reach into other packages as a dependency (like this) which increases the number of Mu overrides that can't be reconciled until the package is upstreamed or the code is refactored to eliminate the PolicyServicePkg dependency.

spbrogan: I would not consider PolicyServicePkg a baseline package that other edk2 based packages should take a dependency on.  If this is just for writing a bad unit test we should move the unit test and/or find a different scenario that produces the same bad result.

makubacki: Agree

UnitTestFrameworkPkg/Library/GoogleTestBadExamples/GoogleTestBadExamplesSrc.c
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+  Get policy protocol from our database.
+
+  @retval EFI_SUCCESS       Policy is located successfully.
+  @retval EFI_ERROR         Failed to locate policy protocol.
+**/
+EFI_STATUS
+EFIAPI
+GetPolicy (
```
VivianNK, line 25: Why is a real function being used instead of a mock?

UnitTestFrameworkPkg/Library/GoogleTestBadExamples/Googletest/GoogleTestBadExamplesUT.inf
kuqin12: The usage of PolicyService is confusing matters. Removed reference to that. The real focus is on mocking the locate protocol here.

kuqin12:
Will not go through this route for death test. The other ideas are disabling asserts or enable that through unit test debug assert routine. Anyway, close this for now.

Issue #369, opened 2023-04-21
[Feature]: Supporting stuart build for Linux ARM


kuqin12:
### Feature Overview

Related to https://github.com/microsoft/mu_basecore/issues/305, this feature request is to add support for stuart build on Linux ARM, which involves building binaries for external dependencies and pipeline setup.

### Solution Overview

1. Adding basetool builds for base core.
2. Updating pipeline to run Linux ARM builds.

### Urgency

Low

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

kuqin12:
The binary dependencies are all resolved. Still need to onboard pipelines.

Also, the SBSA builds does not work as is due to GLOBAL_OFFSET_TABLE cannot be linked for Standalone MM Core.

The rest of the work is staged here: https://github.com/kuqin12/mu_tiano_platforms/tree/test_arm64

kuqin12:
Still need 202302 to pick up the latest BaseTools to fix gcc builds for AARCH64 targets.

Issue #305, opened 2023-03-07
[Feature]: Support stuart building on Windows ARM systems


kuqin12:
### Feature Overview

With better support of Windows ARM ecosystem, more dependencies required by stuart build there were not available a few years ago are ready for consumption. This issue is created to support native stuart build on Windows ARM systems.

### Solution Overview

The main pieces of this support will include:

1. EDK2 pytool change to support VS discovery on ARM systems
2. BaseTools change to support VS toolchain query and native builds of BaseTools
3. IASL builds inside ARM environments
4. Reuse of x86 version of nasm (at least for now)

The changes will be validated on with integration test on Q35 and SBSA virtual platforms.

### Urgency

Medium

### Are you going to implement the feature request?

I will implement the feature

### Do you need maintainer feedback?

No maintainer feedback needed

kuqin12:
IASL build and basetool builds are staged in this branch: https://github.com/kuqin12/mu_basecore/tree/base_tool_arm. Will continue to experiment more changes needed.

EDK2 changes are stages in this branch: https://github.com/kuqin12/edk2-pytool-library/tree/arm_support

kuqin12:
The binaries are resolved. Building should be functional now. Still need pipeline support.

kuqin12:
This is done.

PR #404, opened 2023-05-16
Onboarding ARM64 builds on selfhosted Azure pipeline agents

kuqin12:
# Preface

## Description

This change added a few new matrix entries to support building mu_tiano_platforms on native ARM64 systems. The PR will cover both https://github.com/microsoft/mu_basecore/issues/369 and https://github.com/microsoft/mu_basecore/issues/305.

The PR should also be incorporated with mu_devops change.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on selfhost-agents and existing agents.

## Integration Instructions

Pipeline changes, N/A for integration.


.azurepipelines/Ubuntu-GCC5.yml
kuqin12, line 1: To be integrated to mu_devops.

.azurepipelines/Windows-VS.yml
kuqin12, line 1: Ditto.

kuqin12: Done

.azurepipelines/Ubuntu-GCC5.yml
kuqin12: Done.

.azurepipelines/Matrix-Build-Job.yml
kuqin12, line 1: Need to remove the CLANGPDB files in this PR.

kuqin12: This is removed.

.azurepipelines/MuDevOpsWrapper.yml
```diff
@@ -0,0 +1,88 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+resources:
+  repositories:
+    - repository: mu_devops
+      type: github
+      endpoint: microsoft
+      name: kuqin12/mu_devops
+      ref: test_arm64
```
makubacki: I think you can go ahead and complete https://github.com/microsoft/mu_devops/pull/154.

The Mu repos will continue to use the 3.0.0 version of the repo that they're fixed against. Then, you can update these PRs to match the 4.0.0 version tag on the microsoft org repo (which will be found after that PR completes) so it matches what would be synced to the repo.

.azurepipelines/MuDevOpsWrapper.yml
```diff
@@ -0,0 +1,88 @@
...
+  - Windows_NT
+  - Linux
+  default: Windows_NT
+- name: build_matrix
+  displayName: Build matrix for this repository
+  type: object
+- name: pool_name
+  displayName: Variable name that hosts pool name to be used for self-hosted agents
+  type: string
+  default: pool_name
```
makubacki: Is the default really meant to be `"pool_name"`? Checking because typically the parameter would be defined/overridden with a pipeline value or be empty as a default.

UnitTestFrameworkPkg/ReadMe.md
makubacki, line 1: Was this change intentionally added to the PR?

makubacki (approved):

UnitTestFrameworkPkg/ReadMe.md
kuqin12: Yeah, it was caught on the selfhosted agents. Partly because the lowest supported version of nodejs tools on ARM system was higher than what we use on x86 pipelines. I can check again to see if ARM systems now have better support, but the only difference spotted so far are this spell check and a markdown lint error here: https://github.com/microsoft/mu_tiano_plus/pull/134/files#diff-0f5af5a188764e351b669698e5b9673f67c181f78905f832a7f15562712dc514.

makubacki: Okay, thanks for the explanation.

.azurepipelines/MuDevOpsWrapper.yml
makubacki: This is updated to `v4.0.0` now so marking resolved.

cfernald (approved):

.azurepipelines/MuDevOpsWrapper.yml
kuqin12: Yeah, I think we do not have to set this default to `pool_name`. This was set to `pool_name` in synced file so that the applicable repos do not have to supply the variable of `pool_name` in their top level yaml across all repos.

Another catch is that this is actually the "variable name" instead of the "variable value". :) So that it is indeed driven by a variable named "pool_name". This was done because the pool name _value_ will be used in 2 different scenarios: one is in the condition, so that we can skip the jobs that indicates selfhost run but does not have selfhost pool specified; the other is to specify the pool name. In order to resolve the variables properly during the pipeline run, we set it this way so that the variables have the correct value when it is being evaluated.

UnitTestFrameworkPkg/ReadMe.md
kuqin12: I updated the cspell version on selfhost agents to be consistent with what we use for normal pipeline, did not see any issue with the building. will revert this change (and other applicable ones)

kuqin12: MD change reverted.

.azurepipelines/MuDevOpsWrapper.yml
kuqin12: Resolving this for now given 1. the consumer using the Jobs/PRGate can consume the default empty parameter, 2. the sycned repos are in a more controlled environment which could use the "pool_name" as default to simplify the top level scripts. We can revisit this if deem needed later.

PR #407, opened 2023-05-19
Fix some miscellaneous issues in cryptopkg.

kenlautner:
## Description

Fixed some miscellaneous issues in cryptopkg including the x509 test causing a page fault and the version for the crypto files being the incorrect version.  Changing the crypto version makes this a breaking change.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built and tested on QemuQ35 including the basecrypto tests.

## Integration Instructions

Update to the latest basecore commit.


apop5 (approved):

CryptoPkg/Library/BaseCryptLib/Pk/CryptX509.c
```diff
@@ -1071,6 +1071,7 @@ X509GetSerialNumber (
   }
 
   if (SerialNumber != NULL) {
+    *SerialNumberSize = MIN ((UINTN)Asn1Integer->length, *SerialNumberSize);
```
kuqin12: I think the line 1079 just nullifies your change here?

kenlautner: That's true but this is necessary for CopyMem.  There's a case where Asn1Integer->length is smaller than SerialNumberSize and we're copying more data than what exists.  I'm accounting for that here.

CryptoPkg/Library/BaseCryptLib/Pk/CryptX509.c
```diff
@@ -1071,6 +1071,7 @@ X509GetSerialNumber (
   }
 
   if (SerialNumber != NULL) {
+    *SerialNumberSize = MIN ((UINTN)Asn1Integer->length, *SerialNumberSize);
     CopyMem (SerialNumber, Asn1Integer->data, *SerialNumberSize);
```
kuqin12:
```suggestion
    CopyMem (SerialNumber, Asn1Integer->data, (UINTN)Asn1Integer->length);
```
After offline discussion, the change itself is not an issue given the check performed a few lines above. But it is suggested to change the update to above shape instead of updating SerialNumberSize twice for better clarity.

kenlautner: Updated.

kuqin12 (approved):

PR #420, opened 2023-05-26
Add support for IAD-style USB input devices in ConPlatform

joschock:
## Description

Some multi-function input devices (e.g. combo keyboard and mouse) present as IAD-style devices (see: https://learn.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-interface-association-descriptor). Historically, multi-function devices would report a DeviceClass of 0, indicating that interface matching should be done on the interface descriptor rather than the global device descriptor. 

IAD-style devices us DeviceClass of 0xEF, so they don't match MatchUsbClass() for keyboard (DeviceClass=3, SubClass=1, Proto=1). If they are treated as if they had a DeviceClass of zero, which is more traditional for legacy multi-function devices, then the interface descriptors are used instead and these types of devices will "just work" without needing to add a custom USB device path to ConIn.

- [x] Impacts functionality?
  - IAD-style devices will now use interface descriptor matching rather than device descriptor matching when evaluating ConIn contents. Mostly this means that composite input devices that didn't used to work without ConIn customization in UEFI will now work out of the box.
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified this with an arduino-based test device that exposes an IAD-style descriptor with keyboard and mouse HID interfaces. Prior to this change, device keyboard input was not usable in FP or shell because it didn't match the default ConIn USB input descriptor. After this change, the keyboard input worked as expected. 

## Integration Instructions

No integration required.

### DIFF
--- a/MdeModulePkg/Universal/Console/ConPlatformDxe/ConPlatform.c
+++ b/MdeModulePkg/Universal/Console/ConPlatformDxe/ConPlatform.c
@@ -808,10 +808,15 @@ MatchUsbClass (
   DeviceClass    = DevDesc.DeviceClass;
   DeviceSubClass = DevDesc.DeviceSubClass;
   DeviceProtocol = DevDesc.DeviceProtocol;
-  if (DeviceClass == 0) {
+
+  // MSCHANGE - BEGIN - Add support for IAD-type multi-function devices.
+  if ((DeviceClass == 0) ||
+      ((DeviceClass == 0xEF) && (DeviceSubClass == 0x02) && (DeviceProtocol == 0x01)))
+  {
     //
-    // If Class in Device Descriptor is set to 0, use the Class, SubClass and
-    // Protocol in Interface Descriptor instead.
+    // If Class in Device Descriptor is set to 0 (Device), or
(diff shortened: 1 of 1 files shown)


kenlautner (approved):

VivianNK (approved):
Curious - Are there currently or could there foreseeably be any other cases of Device Class/SubClass/Protocol combinations that would call for being treated like traditional device class 0 devices?

makubacki (approved):

joschock:
> Curious - Are there currently or could there foreseeably be any other cases of Device Class/SubClass/Protocol combinations that would call for being treated like traditional device class 0 devices?

I'm not sure - no one has brought me one yet that doesn't work besides the IAD one I have ;). A quick scan of the defined classes on USB.org doesn't yield any obvious candidates, but IAD wouldn't have been obvious either. Note that this particular driver is mostly concerned with things that look like input devices (e.g. provide SimpleTextIn), so things like composite video capture USB devices with audio/camera etc. aren't in scope.

PR #423, opened 2023-05-30
Integrate PrEval Policy 5

Javagedes:
## Description

Integrates edk2-pytool-extension's PrEval's Policy 5 that runs whenever a INF has been changed. It will parse each package's dsc (as defined in `<pkg>.ci.yaml::PrEval.DscPath`) and mark it to be build if that package has a dependency on the changed INF.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/CryptoPkg.ci.yaml
+++ b/CryptoPkg/CryptoPkg.ci.yaml
@@ -5,7 +5,11 @@
 # Copyright (c) 2020 - 2022, Intel Corporation. All rights reserved.<BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
-{
+{   # MU_CHANGE begin
+    "PrEval": {
+        "DscPath": "CryptoPkg.dsc",
+    },
+    # MU_CHANGE end
     "LicenseCheck": {
         "IgnoreFiles": [
             # These directories contain auto-generated OpenSSL content
--- a/CryptoPkg/CryptoPkg.ci.yaml
+++ b/CryptoPkg/CryptoPkg.ci.yaml
@@ -5,7 +5,8 @@
 # Copyright (c) 2020 - 2022, Intel Corporation. All rights reserved.<BR>
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
-{   # MU_CHANGE begin
+{
+    # MU_CHANGE begin
     "PrEval": {
         "DscPath": "CryptoPkg.dsc",
     },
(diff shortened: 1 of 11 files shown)


kenlautner (approved):

os-d (approved):
This LGTM, my only comment is that with additional plugins, we are starting to accumulate the same DscPath over and over again (with one case using the host test dsc). Would it be worthwhile to have a global, hey, this is the default DscPath that realistically every plugin is going to use except the one(s) that use the host test dsc?

makubacki (approved):

Javagedes:
> This LGTM, my only comment is that with additional plugins, we are starting to accumulate the same DscPath over and over again (with one case using the host test dsc). Would it be worthwhile to have a global, hey, this is the default DscPath that realistically every plugin is going to use except the one(s) that use the host test dsc?

I agree that having the DscPath in multiple places is not the best, but edk2-pytool-extensions has to work under the expectation that no plugins exist. We don't want to add dependencies between PREval and the compiler plugin or vice versa.

os-d:
> I agree that having the DscPath in multiple places is not the best, but edk2-pytool-extensions has to work under the expectation that no plugins exist. We don't want to add dependencies between PREval and the compiler plugin or vice versa.

Sure, I totally agree. What I mean though is a global section to pull the DscPath from. I.e., don't pull it from a different plugin's DscPath, but have a global DscPath that all plugins inherit unless they override it (like for the host test plugin).

All this to say, it doesn't really matter too much, just occurred to me. With some of the discussions going on in edk2-pytools land, it may cause this to not be an issue at all.

Javagedes:
> > I agree that having the DscPath in multiple places is not the best, but edk2-pytool-extensions has to work under the expectation that no plugins exist. We don't want to add dependencies between PREval and the compiler plugin or vice versa.
> 
> Sure, I totally agree. What I mean though is a global section to pull the DscPath from. I.e., don't pull it from a different plugin's DscPath, but have a global DscPath that all plugins inherit unless they override it (like for the host test plugin).
> 
> All this to say, it doesn't really matter too much, just occurred to me. With some of the discussions going on in edk2-pytools land, it may cause this to not be an issue at all.

Depending on if the toml config changes go through the compiler plugin won't be necessary anymore anyway :) The only reason it exists now is because we don't have a settings file for each package with a UefiBuild object attached, so we have to rely on the compiler plugin to instantiate a generic UefiBuild object to build a non platform repo :)

PR #425, opened 2023-05-31
Explicitly set execute bits for linux basetools binaries during publish

cfernald:
## Description

The artifacts published from the build stage of the basetools pipelines will lose their executable attributes when downloaded in the published phase. This change explicitly makes the binaries executable as part of the publishing job for linux binaries. This could also be done by zipping the binaries in the first stage of the pipeline, but the added steps add little value over the simple chmod fix. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested in pipeline on personal basecore fork. 

## Integration Instructions

N/A


cfernald:
Closing as I discovered the python implementation of the zip decompress does not respect the external file attributes so it makes no difference for an extdep.

PR #469, opened 2023-06-26
MdeModule: resolve invalid library override

Javagedes:
## Description

MediaSanitizeUnitTestHost used an invalid library override, attempting to override a non-existent library with a component that does not have an associated library.  Comparing generated output of the inf files with and without a library override showed no changes.  Therefore, this commit removes the invalid library override.

Additionally, the two files provided byt NvmExpressDxe.inf (  NvmExpressMediaSanitize.c
  NvmExpressMediaSanitize.h) are already included in MediaSanitizeUnitTestHost.inf.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Diff'd the build artifacts of the component with and without the library class override and saw no changes in the following:

- deps.txt
- Makefile
- static_library_files.lst
- NvmExpressMdeiaSanitize.c.deps
- MediaSanitizeUnitTestHost.inf
- MediaSanitizeUnitTest.c.deps

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Test/MdeModulePkgHostTest.dsc
+++ b/MdeModulePkg/Test/MdeModulePkgHostTest.dsc
@@ -83,7 +83,7 @@
       UefiBootServicesTableLib|MdePkg/Library/UefiBootServicesTableLib/UefiBootServicesTableLib.inf
   }
 
-  MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTestHost.inf  # MU_CHANGE - Invalid Library override
+  MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTestHost.inf  # MU_CHANGE - Add Additional Testing
 
   # MU_CHANGE [BEGIN]
   MdeModulePkg/Library/VariablePolicyLib/VariablePolicyUnitTest/VariablePolicyUnitTest.inf {
--- a/MdeModulePkg/Test/MdeModulePkgHostTest.dsc
+++ b/MdeModulePkg/Test/MdeModulePkgHostTest.dsc
@@ -83,10 +83,7 @@
       UefiBootServicesTableLib|MdePkg/Library/UefiBootServicesTableLib/UefiBootServicesTableLib.inf
   }
 
-  MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTestHost.inf {
-    <LibraryClasses>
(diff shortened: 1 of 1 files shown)


Javagedes:
@makubacki not sure if this should be a NULL override? As I mentioned above this is a fake library being overwritten by a component and does not make sense. I don't see any changes in build artifacts, however if this absolutely needs added, it would make sense as NULL override

makubacki:
> @makubacki not sure if this should be a NULL override? As I mentioned above this is a fake library being overwritten by a component and does not make sense. I don't see any changes in build artifacts, however if this absolutely needs added, it would make sense as NULL override

No, I think you did the right thing to remove it. It is possible that `NvmExpressMediaSanitize.c` links functions from other `NvmExpressDxe` files, but running through a build and the other files you looked at should have shown that.

MdeModulePkg/Test/MdeModulePkgHostTest.dsc
```diff
@@ -83,10 +83,7 @@
       UefiBootServicesTableLib|MdePkg/Library/UefiBootServicesTableLib/UefiBootServicesTableLib.inf
   }
 
-  MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTestHost.inf {
-    <LibraryClasses>
-      NvmExpressDxe|MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressDxe.inf
-  }
+  MdeModulePkg/Bus/Pci/NvmExpressDxe/UnitTest/MediaSanitizeUnitTestHost.inf  # MU_CHANGE - Invalid Library override
```
makubacki: It looks like this is not in [edk2/master/MdeModulePkg/Test/MdeModulePkgHostTest.dsc](https://raw.githubusercontent.com/tianocore/edk2/master/MdeModulePkg/Test/MdeModulePkgHostTest.dsc). The `MU_CHANGE` comment is needed, but it could just mention adding the test since the invalid library class override is irrelevant now.

makubacki (approved):

MdeModulePkg/Test/MdeModulePkgHostTest.dsc
Javagedes: Updated comment - thanks!

TaylorBeebe (approved):

PR #480, opened 2023-06-28
Change default alignment for ramdisk booting.

ddeptford:
The ramdisk is modelled as an NVDIMM which have a naturally higher alignment than 4K. Operating systems may wish to map NVDIMMs using large pages, so force the allocation alignment to 2MB.

- [X] Impacts functionality?
  - Alignment of ramdisk is changed from 4KB to 2MB.
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested using HTTP ramdisk boot, targeting a flat windows image.

## Integration Instructions

N/A


MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
```diff
@@ -1505,7 +1509,7 @@ BmExpandLoadFile (
 
   Status = LoadFile->LoadFile (LoadFile, FilePath, TRUE, &BufferSize, FileBuffer);
   if (EFI_ERROR (Status)) {
-    FreePages (FileBuffer, EFI_SIZE_TO_PAGES (BufferSize));
+    FreeAlignedPages (FileBuffer, EFI_SIZE_TO_PAGES (BufferSize));
```
os-d, line 1517: I think we also need to change this to FreeAlignedPages, also? https://github.com/microsoft/mu_basecore/blob/release/202302/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c#L1416C12-L1416C12. Not sure if there is another destruction path that needs to be updated.

apop5 (approved):

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
makubacki, line 1: `MU_CHANGE` tags are needed.

makubacki: `MU_CHANGE` tags are specific to our contribution process for changes we make to files that exist in the upstream edk2 package. Let me know if you have any questions or need help with those.

ddeptford: Done, thanks for the help!

MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
ddeptford: Done, I think that's it.  Note that the body for FreeAlignedPages is a carbon copy of FreePages, so I'm not sure we'd get any kind of runtime error in case of mismatch, which is a bit unfortunate.

os-d: I had a suspicion that was the case, but didn't check it :).

os-d (approved):

makubacki (approved):

PR #497, opened 2023-07-19
BaseTools: Detect library class mismatch [REBASE&FF]

Javagedes:
## Description

Performs a check that will verify that the library instance implements the specified library by ensuring a LIBRARY_CLASS definition exists in the INF [Defines] section matches the library.

That is to say, for the following example:
`TestLib|Path/To/BaseTestLib.inf`, that BaseTestLib.inf has `LIBRARY_CLASS = TestLib` defined in the [Defines] section.

Changes the previous check for library overrides to be a warning instead of an error.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A


BaseTools/Source/Python/Workspace/DscBuildData.py
```diff
@@ -1190,6 +1203,17 @@ def __ParsePcdFromCommandLine(self):
             for item in delete_assign:
                 GlobalData.BuildOptionPcd.remove(item)
 
+    def _ValidateLibraryClass(self, LibraryClass: str, LibraryInstance: PathClass) -> bool:
+        if LibraryClass =='NULL':
```
makubacki, line 1207: nite: Space before `'NULL'`

```suggestion
        if LibraryClass == 'NULL':

```

BaseTools/Source/Python/Workspace/DscBuildData.py
```diff
@@ -763,6 +763,12 @@ def Modules(self):
                 LibraryPath = PathClass(NormPath(Record[1], Macros), GlobalData.gWorkspace, Arch=self._Arch)
                 LineNo = Record[-1]
 
+                # Validate that the Library instance implements the Library Class
+                if not self._ValidateLibraryClass(LibraryClass, LibraryPath):
+                    EdkLogger.warn("build",
```
makubacki, line 768: There was an open earlier about warning visibility from edk2 tools. Are these visible in our current builds?

BaseTools/Source/Python/Workspace/DscBuildData.py
makubacki, line 1: My understanding is that at this time we should keep `MU_CHANGE` tags around the changes.

makubacki (approved):

kenlautner (approved):

BaseTools/Source/Python/Workspace/DscBuildData.py
Javagedes: This is correct. RunCmd() does no parsing of the output, only redirecting to the specified log level. The `scan_compiler_output` could be updated to detect Build.py warnings and split them out after the build. It would be the same situation as `path_utilities`. It spews the warning multiple times as end up parsing, and throwing the warning, twice per build. Here is an example of building mu_tiano_platforms:


image

Javagedes: I have a small update queued that can fix the spamming, if you want it added.

Javagedes:
Added an update that applies a filter so we only log a particular warning in the dsc once:

image

PR #499, opened 2023-07-20
BaseTools: Detect library class mismatch [REBASE&FF]

Javagedes:
## Description
Performs a check that will verify that the library instance implements
the specified library by ensuring a LIBRARY_CLASS definition exists in
the INF [Defines] section matches the library.

That is to say, for the following example:
`TestLib|Path/To/BaseTestLib.inf`, that BaseTestLib.inf has
`LIBRARY_CLASS = TestLib` defined in the [Defines] section.

Changes the previous check for library overrides to be a warning instead
of an error.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built multiple platforms to confirm warning showed properly.

## Integration Instructions

While no Integration is required, a new build warning will be seen in the build log. The warning is in the following format:
`$(DSC_PATH): warning: $(INF) does not support LIBRARY_CLASS $(LIBRARY_CLASS)` where `$(DSC_PATH)` is the DSC being built, `$(INF)` is the INF described in the DSC, and `$(LIBRARY_CLASS)` is the name of the library class that the INF is attempting to represent. (i.e., `$(LIBRARY_CLASS)|$(INF)` or `TestLib|Path\to\BaseTestLib.inf`).

To resolve these errors, verify the library class is real (The Library class should have a header file associated with it in the DEC of the defining package). If it is real, then ensure the library instance truly implements the library class and add it to the DEFINE section of the library INF. If not, update the DSC to use the correct library class.

If this warning appears, it will appear at the beginning of the build command:
INFO - Processing meta-data .
INFO - Architecture(s)  = IA32 X64
INFO - Build target     = DEBUG
INFO - Toolchain        = VS2022
INFO - 
INFO - Active Platform          = c:\src\Path\Platform.dsc
INFO - build...
INFO - c:\src\Path\Platform.dsc(...): warning: c:\src\Path\BaseTestLib.inf does not support LIBRARY_CLASS TestLib

### DIFF
--- a/BaseTools/Source/Python/AutoGen/AutoGenWorker.py
+++ b/BaseTools/Source/Python/AutoGen/AutoGenWorker.py
@@ -218,6 +218,7 @@ class AutoGenWorkerInProcess(mp.Process):
             GlobalData.gEnableGenfdsMultiThread = self.data_pipe.Get("EnableGenfdsMultiThread")
             GlobalData.gPlatformFinalPcds = self.data_pipe.Get("gPlatformFinalPcds")
             GlobalData.file_lock = self.file_lock
+            GlobalData.gLogLibraryMismatch = False # MU_CHANGE
             CommandTarget = self.data_pipe.Get("CommandTarget")
             pcd_from_build_option = []
             for pcd_tuple in self.data_pipe.Get("BuildOptPcd"):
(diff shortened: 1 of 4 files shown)


BaseTools/Source/Python/AutoGen/AutoGenWorker.py
makubacki, line 1: In the integration instructions section of the PR description, can you show an example of what the warning looks like, where to find it, etc.?

Javagedes: Updated! Thanks.

makubacki (approved):

kenlautner (approved):

PR #507, opened 2023-08-01
Set OpenCppCoverage working dir to avoid files in root

cfernald:
## Description

Set the working dir for OpenCppCoverage. OpenCppCoverage will ALWAYS create a file called LastCoverageResults.log in it's current working directory, regardless of it's working_dir parameter. This change makes sure this file is opened in the build dir instead.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with local builds

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
+++ b/BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
@@ -207,14 +207,14 @@ class HostBasedUnitTestRunner(IUefiBuildPlugin):
         # Generate coverage file
         coverageFile = ""
         for testFile in testList:
-            ret = RunCmd("OpenCppCoverage", f"--source {workspace} --export_type binary:{testFile}.cov -- {testFile}")
+            ret = RunCmd("OpenCppCoverage", f"--source {workspace} --export_type binary:{testFile}.cov -- {testFile}", workingdir=f"{workspace}Build/")
             coverageFile += " --input_coverage=" + testFile + ".cov"
             if ret != 0:
                 logging.error("UnitTest Coverage: Failed to collect coverage data.")
                 return 1
 
         # Generate and XML file if requested.by each package
-        ret = RunCmd("OpenCppCoverage", f"--export_type cobertura:{os.path.join(buildOutputBase, 'coverage.xml')} --working_dir={workspace}Build {coverageFile}")
+        ret = RunCmd("OpenCppCoverage", f"--export_type cobertura:{os.path.join(buildOutputBase, 'coverage.xml')} --working_dir={workspace}Build/ {coverageFile}", workingdir=f"{workspace}Build/")
         if ret != 0:
             logging.error("UnitTest Coverage: Failed to generate cobertura format xml in single package.")
             return 1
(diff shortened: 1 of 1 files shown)


BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -207,14 +207,14 @@ def gen_code_coverage_msvc(self, thebuilder):
...
-            ret = RunCmd("OpenCppCoverage", f"--source {workspace} --export_type binary:{testFile}.cov -- {testFile}")
+            ret = RunCmd("OpenCppCoverage", f"--source {workspace} --export_type binary:{testFile}.cov -- {testFile}", workingdir=f"{workspace}Build/")
             coverageFile += " --input_coverage=" + testFile + ".cov"
             if ret != 0:
                 logging.error("UnitTest Coverage: Failed to collect coverage data.")
                 return 1
 
         # Generate and XML file if requested.by each package
-        ret = RunCmd("OpenCppCoverage", f"--export_type cobertura:{os.path.join(buildOutputBase, 'coverage.xml')} --working_dir={workspace}Build {coverageFile}")
+        ret = RunCmd("OpenCppCoverage", f"--export_type cobertura:{os.path.join(buildOutputBase, 'coverage.xml')} --working_dir={workspace}Build/ {coverageFile}", workingdir=f"{workspace}Build/")
```
Javagedes: Is it necessary to set the working dir both via the OpenCppCoverage cli flag and via RunCmd? I looks like either or would work fine.

cfernald: Ah, yes we can probably remove the CLI flag since we are not doing it in the popen.

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
```diff
@@ -207,14 +207,14 @@ def gen_code_coverage_msvc(self, thebuilder):
...
         for testFile in testList:
-            ret = RunCmd("OpenCppCoverage", f"--source {workspace} --export_type binary:{testFile}.cov -- {testFile}")
+            ret = RunCmd("OpenCppCoverage", f"--source {workspace} --export_type binary:{testFile}.cov -- {testFile}", workingdir=f"{workspace}Build/")
             coverageFile += " --input_coverage=" + testFile + ".cov"
             if ret != 0:
                 logging.error("UnitTest Coverage: Failed to collect coverage data.")
                 return 1
 
         # Generate and XML file if requested.by each package
-        ret = RunCmd("OpenCppCoverage", f"--export_type cobertura:{os.path.join(buildOutputBase, 'coverage.xml')} --working_dir={workspace}Build {coverageFile}")
```
Javagedes, line 217: Instead of doing the complicated mess of checking if workspace ends with the os path separator (L206), you could use the pathlib library which will handle that for you. `workspace = Path(builder.env.GetValue("WORKSPACE"))`. Then you can do things like `workingdir=str(workspace / "Build")`

Pathlib could be used instead of os.path.join in many places here.

Up to you if you want to make those changes or not. It technically works and it would be outside the scope of this PR, so :shrug:

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cfernald: To clarify though, we CLI one is not sufficient since it ignores that flags value for LastCoverageResults.log 😢

Javagedes (approved):

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
cfernald: Yeah, this one came from upstream. I'm not sure if I care too much one way or the other. I'm happy to throw in the change though.

Javagedes:
If it’s from upstream, it’s probably not worth the diff.

apop5 (approved):

makubacki (approved):

BaseTools/Plugin/HostBasedUnitTestRunner/HostBasedUnitTestRunner.py
Javagedes: If the change is upstream, it's not worth the diff or time to bring it upstream, in my opinion.

PR #529, opened 2023-08-16
SpellCheck: force no color codes

Javagedes:
## Description

The SpellCheck plugin runs cspell, which uses ANSI color codes when displaying results. When ANSI color codes are used in a terminal that does not support ANSI, the color codes are displayed as text, which causes the rest of SpellCheck plugin to fail as the expected path is not a real path as it starts and ends with color codes:

i.e. `\[90m.\Path\To\File.txt[39m` instead of `.\Path\To\File.txt`

This change forces cspell to not colorize the output.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/.pytool/Plugin/SpellCheck/SpellCheck.py
+++ b/.pytool/Plugin/SpellCheck/SpellCheck.py
@@ -212,7 +212,7 @@ class SpellCheck(ICiBuildPlugin):
     def _check_spelling(self, abs_file_to_check: str, abs_config_file_to_use: str) -> []:
         output = StringIO()
         ret = RunCmd(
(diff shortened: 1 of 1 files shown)


makubacki (approved):

makubacki:
From what I understand there was a regression in the pipelines. It would be good to make sure that is understood. But I don't generally have a problem with disabling color codes from cspell.

apop5 (approved):

Issue #503, opened 2023-07-27
[Feature]: Provide a hook API when stack overflow occur with stack protection enabled


MarcChen46:
### Feature Overview

**[Background]**
Currently we have EFI_CPU_ARCH_PROTOCOL.RegisterInterruptHandler() for DXE_DRIVER and EFI_SMM_CPU_SERVICE_PROTOCOL.RegisterExceptionHandler() for DXE_SMM_DRIVER to register EXCEPT_IA32_STACK_FAULT type that can perform some customized actions when stack overflow occurred with stack protection enabled, but it has a dependency that the customized code can only get performed after these protocols get installed, it is the gap we want to resolve.

**[Feature request]**
When stack overflow occurred in any protected modules types(ex: DXE_DRIVER, DXE_CORE, SMM_CORE, DXE_SMM_DRIVER...), it should provide a hook API or mechanism that allow platform to add customized code (ex: report the error to somewhere as telemetry), the customized should not have dependency on protocols.

### Solution Overview

Below is just a proposal for reference, it is ok to have any other kind of methods to achieve it.
**[Proposal]**
Implement a MU version of BaseStackCheckLib instance and implement below code in BaseStackCheckMsft.c, use MSFT VS compiler to compile this C code that has an customized hook API inside __report_gsfailure.

```
NORETURN VOID __cdecl
__report_gsfailure (
  void
  )
{
  DEBUG ((DEBUG_ERROR, "(__report_gsfailure) STACK PROTECTION : Violation detected\n"));

  CustomizedHook ();  // API provided by another new added library class
  CpuDeadLoop ();
}
```

### Urgency

Low

### Are you going to implement the feature request?

Someone else needs to implement the feature

### Do you need maintainer feedback?

Maintainer feedback requested

MarcChen46:
@TaylorBeebe, sorry, I did not notice the issue would be closed automatically, may I know if there is any plan for this feature request? can we re-open this issue?

TaylorBeebe:
> @TaylorBeebe, sorry, I did not notice the issue would be closed automatically, may I know if there is any plan for this feature request? can we re-open this issue?

This feature request is in progress -- reopened the issue

TaylorBeebe:
@MarcChen46
This feature will be completed soon. I posted a draft PR which will be cleaned up over the coming days. I'll ping this issue when the PR is out of draft status. https://github.com/microsoft/mu_basecore/pull/610

[...]

PR #533, opened 2023-08-21
Add DXE Stack Cookie Failure Hook

TaylorBeebe:
## Description

Reference: https://github.com/microsoft/mu_basecore/issues/503

The above issue requests the ability to hook into a stack cookie check failure and recommends a library class link within the failure check. To accomplish the same goal and minimize the memory boat, a new instance of the security cookie check logic has been added which, before calling the interrupt, will attempt to locate a stack cookie check failure hook protocol and execute its published function.

This implementation:
1. Allows any platform to add handler logic with the usual levers (a new module and possibly an APRIORI entry) 
2. Allows the stack cookie failure hook logic to be located in one place in memory instead of in every DXE module
3. Calls the hook before the interrupt meaning the hook logic will have normal access to system resources

However, because the library is called "BaseBinSecurityLib" instead of "BinSecurityLib" the DXE instance of BaseBinSecurityLib doesn't fit usual code conventions. Ideally, BaseBinSecurityLib should be renamed to BinSecurityLib, but currently every package DSC references this library so it would need to be renamed across several repos making this a breaking change. **If anyone has thoughts on this, please share!**

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by calling __security_check_cookie(1) at the end of DXE main to ensure the hook is properly being invoked on Q35.

## Integration Instructions

N/A


MarcChen46:
Looks like the `BaseBinSecurityLibRng` only supports X64 components now, is that expected? How about IA32?

Second question is that using protocol may have race condition that the hook procedure is only available after the protocol get installed, is it possible that we can add another null library with a hook API, then developers could feel free to implement their hook procedure either in the library's API or install a hook protocol depends on their requirement.

TaylorBeebe:
> 

It seems like this code should be compliable with ARM, AARCH64 and IA32 targets. I'll double check and remove the architecture restriction from this and the original BaseBinSecurityLib if so.

For your second question, I think what you want is to ensure the module producing the protocol is the first dispatched by the DXE core. In this case, the preferable method is to create the module which produces the protocol and add it to the top of the APRIORI list so it's at the front of the dispatch queue.

MarcChen46:
> > 
> 
> It seems like this code should be compliable with ARM, AARCH64 and IA32 targets. I'll double check and remove the architecture restriction from this and the original BaseBinSecurityLib if so.
> 
> For your second question, I think what you want is to ensure the module producing the protocol is the first dispatched by the DXE core. In this case, the preferable method is to create the module which produces the protocol and add it to the top of the APRIORI list so it's at the front of the dispatch queue.

yes, please add support for all kinds of targets, and yes, the APRIORI can guarantee the dispatch order, but I'm thinking if we want to add stack protect to all kinds of module type including PEI_CORE and DXE_CORE, then using a hook API from a library may be better than Protocol.

BTW, current implementation only provides protocol definition, for supporting PEI, please also create PPI for hook.

MdePkg/Include/Protocol/StackCookieCheckFailureHook.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef STACK_COOKIE_CHECK_FAILURE_HOOK_H__
+#define STACK_COOKIE_CHECK_FAILURE_HOOK_H__
+
+#define STACK_COOKIE_CHECK_FAILURE_HOOK_PROTOCOL_GUID \
+  { \
+    0xC1F15632, 0x4E68, 0x4503, { 0x98, 0x6C, 0x61, 0x89, 0x64, 0xD1, 0x1A, 0xB5 } \
+  }
+
+/**
+ BEEBE TODO
```
MarcChen46, line 16: please remember to remove the todo before complete.

MdePkg/Include/Protocol/StackCookieCheckFailureHook.h
```diff
@@ -0,0 +1,30 @@
+/** @file
+  Defines the protocol interface which can be called before the stack cookie interrupt
+  when a stack cookie check fails.
+
```
makubacki, line 4: Missing copyright and SPDX ID.

MdePkg/Library/DxeBinSecurityLibRng/DxeBinSecurityLibRng.inf
makubacki, line 1: From a maintenance/integration standpoint, the pre-existing `BaseBinSecurityLibRng` should be converged with the DXE-specific changes here, assuming the changes keep a DXE-specific instance. This is also why library class names should not include phase-binding details.

We often advocate to others to not grow technical debt in fear of breaking the build. Although disruptive, I think we should use the tools we have with "breaking change" tags, semantic version convention, and release notes with clear integration instructions to structure the code in a sustainable way. Since this is only in Project Mu, it removes the upstream burden which would be broader impacting coupled with poorer release awareness mechanisms.

The libraries can share a single directory (e.g. `BinSecurityLibRng`) that contains:

1. A BASE instance INF - `BaseBinSecurityLibRng.inf`
2. A DXE instance INF - `DxeBinSecurityLibRng.inf`
3. A common .c file - `BinSecurityLibRng.c`
4. A DXE-specific .c file - `BinSecurityLibRngDxe.c`
5. Assembly files - Applied in INFs per arch as needed

Now, integrators are aware of both instance's existence. Common code sharing makes functionality delta obvious, bug fixing less error prone, etc.

`__security_check_cookie_Failure` is currently calling `StackCookieFailure()` in the DXE `security_check_cookie.asm` file, that could also be converged for example where a common `StackCookieFailure()` simply returns in the BASE instance.

MdePkg/Library/DxeBinSecurityLibRng/DxeBinSecurityLibRng.inf
```diff
@@ -0,0 +1,41 @@
+## @file DxeBaseBinSecurityLibRng.c
```
makubacki, line 1:
```suggestion
## @file DxeBinSecurityLibRng.inf

```

MdePkg/MdePkg.dec
```diff
@@ -1993,6 +1993,10 @@
   ## Include/Protocol/MemoryProtectionNonstopMode.h
   # 896306BA-1722-414D-93CF-14244F5E315D
   gMemoryProtectionNonstopModeProtocolGuid = {0x896306BA, 0x1722, 0x414D, {0x93, 0xCF, 0x14, 0x24, 0x4F, 0x5E, 0x31, 0x5D }}
+  
+  ## Include/Protocol/StackCookieCheckFailureHook.h
```
makubacki, line 1997: This DEC addition should have `MU_CHANGE` tags.

MdePkg/Library/DxeBinSecurityLibRng/DxeBinSecurityLibRng.inf
makubacki: Similar change needed in the .c file.

makubacki:
> yes, please add support for all kinds of targets, and yes, the APRIORI can guarantee the dispatch order, but I'm thinking if we want to add stack protect to all kinds of module type including PEI_CORE and DXE_CORE, then using a hook API from a library may be better than Protocol.

I think Marc's feedback is valid. The ultimate goal of this change is to provide an optional platform hook.

There are at least a few considerations like MdePkg bloat and platform integration complexity.

#### MdePkg Bloat

This change currently adds a new library instance + a new protocol to MdePkg. The protocol is for a very narrow use case. I don't think this is necessarily a problem as long as the protocol is well-defined and cohesive, but it does expand the number of interfaces exposed by MdePkg which is high visibility.

#### Platform Integration Complexity

Dynamic interfaces tend to be a little more complex due to their dynamic nature as Marc is suggesting. They can help control dynamic order with DEPEX's, share external functionality across binaries, and so on, but if the remainder of the stack is otherwise simple static logic, exposing that through a dynamic interface can create a bit more overall complexity between producer/consumer order.

---

With this change, I look at whether the core code objectives can be met while minimizing bloat and retaining platform integration flexibility.

#### Core code objectives (mentioned):

1. Add simple platform hook for platform action on stack cookie exception
2. Minimize overall firmware size impact

The core code must implement (1). The core or platform code can contribute to (2).

For example, the core code can call a simple library API (without regard to boot phase), that can be implemented in a phase-specific library instance at the platform level that calls a dynamic interface like a protocol or PPI. The platform may choose to use a PPI/Protocol specific to platform service (e.g. `EXTERNAL_ENGINE_COMM_PROTOCOL->ReportError (...)`), just set a bit somewhere, etc. But the core does need to maintain more interfaces than necessary (e.g. PPI and Protocol) to accomplish the same thing. Along with that interface maintenance comes things like future deprecation, API modifications, etc. which can have more overhead with dynamic interfaces (i.e. protocol GUID rolling or versioning) since it must be accounted for at runtime. Tracking this across a PPI and a protocol can increase the work.

PR #534, opened 2023-08-22
DXE and SMM core performance lib codeql changes.

kenlautner:
## Description

Added a few more codeql fixes for the DXE and SMM core performance libs.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on intel physical platforms.  No issues building and booting.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
+++ b/MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
@@ -1191,7 +1191,11 @@ InsertFpdtRecord (
     case PERF_INMODULE_END_ID:
     case PERF_CROSSMODULE_START_ID:
     case PERF_CROSSMODULE_END_ID:
-      GetModuleInfoFromHandle ((EFI_HANDLE)CallerIdentifier, ModuleName, sizeof (ModuleName), &ModuleGuid);
+      Status = GetModuleInfoFromHandle ((EFI_HANDLE)CallerIdentifier, ModuleName, sizeof (ModuleName), &ModuleGuid);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "Failed to get Module Info from Handle! Status = %r\n", Status));
+      }
+
       if (String != NULL) {
         StringPtr = String;
       } else {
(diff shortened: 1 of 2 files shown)


MdeModulePkg/Library/DxeCorePerformanceLib/DxeCorePerformanceLib.c
```diff
@@ -1191,7 +1191,14 @@ InsertFpdtRecord (
     case PERF_INMODULE_END_ID:
     case PERF_CROSSMODULE_START_ID:
     case PERF_CROSSMODULE_END_ID:
-      GetModuleInfoFromHandle ((EFI_HANDLE)CallerIdentifier, ModuleName, sizeof (ModuleName), &ModuleGuid);
+      // MU_CHANGE [BEGIN] - CodeQL change
+      Status = GetModuleInfoFromHandle ((EFI_HANDLE)CallerIdentifier, ModuleName, sizeof (ModuleName), &ModuleGuid);
```
makubacki, line 1195: The main thing being output from `GetModuleInfoFromHandle()` is `ModuleName`, right?

Below this line, `ModuleName` may be assigned to `StringPtr`:

```c
      if (String != NULL) {
        StringPtr = String;
      } else {
        StringPtr = ModuleName;
      }
```

If the error code returned from `GetModuleInfoFromHandle()` was something like `EFI_INVALID_PARAMETER` because the pointer argument was `NULL`, then `StringPtr` could be set to `NULL` in the assignment from `ModuleName` above and then dereferenced in the call to `AsciiStrLen()`:

```c
if (AsciiStrLen (StringPtr) == 0) {
```

It looks like the literal `"unknown name"` could be assigned to `StringPtr` if both `String` and `ModuleName` are `NULL`.

kenlautner: Correct.  If there is an error in GetModuleInfoFromHandle `StringPtr` might be assigned the value `unknown name`.  I added the check to satisfy CodeQL and to give additional information for traceability.  The code already accounted for the failure so functionally I changed nothing.  If we'd like to make a functional change I could do that.

makubacki: GitHub won't leave a suggestion over the actual code, but I think the logic below should be:

```c
      StringPtr = NULL;

      if (String != NULL) {
        StringPtr = String;
      } else if (ModuleName != NULL) {
        StringPtr = ModuleName;
      }

      if ((StringPtr == NULL) || (AsciiStrLen (StringPtr) == 0)) {
        StringPtr = "unknown name";
      }
```

It looks like `StringPtr` might already be `NULL` but the function is so large and unwieldy it doesn't hurt to ensure that going into the logic here.

kenlautner: Updated.

makubacki (approved):

VivianNK (approved):

PR #557, opened 2023-09-07
Add New CodeQL Queries [Rebase & FF]

makubacki:
## Description

Enables 17 new CodeQL queries by updating the queries opted into
by Project Mu available in the standard CodeQL query package
`codeql/cpp-queries`.

I tested many queries and found the included set here to be the
most useful with the least number of false positives. Some queries
had a number of issues that led to them being placed on the exclusion
list so they are not considered in the future without the notes there
being taken into account.

General details about queries available in the pack are available
here:
https://codeql.github.com/codeql-query-help/cpp/

The changelog for the query pack is available here:
https://github.com/github/codeql/blob/codeql-cli/latest/cpp/ql/src/CHANGELOG.md

The following new queries are enabled organized alphabetically
within sections grouped by type:

Errors:

- `cpp/missing-return`
- `cpp/no-space-for-terminator`
- `cpp/pointer-overflow-check`
- `cpp/redundant-null-check-simple`
- `cpp/very-likely-overrunning-write`

 Warnings:

- `cpp/comparison-precedence`
- `cpp/implicit-bitfield-downcast`
- `cpp/offset-use-before-range-check`
- `cpp/overflow-calculated`
- `cpp/overflow-destination`
- `cpp/return-stack-allocated-memory`
- `cpp/static-buffer-overflow`
- `cpp/unsigned-comparison-zero`

Recommendations:

- `cpp/missing-header-guard`
- `cpp/unused-local-variable`
- `cpp/unused-static-function`
- `cpp/unused-static-variable`

The following queries are newly explicitly excluded as noted in the
file:

- `cpp/allocation-too-small`
- `cpp/commented-out-code`
- `cpp/duplicate-include-guard`
- `cpp/invalid-pointer-deref`
- `cpp/useless-expression`

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- CI build
- Run CodeQL locally

## Integration Instructions

Most downstream repos directly use the CodeQL queries defined
here in mu_basecore because the default queries used by the
CodeQL plugin come from `MuCodeQlQueries.qls`.

Most downstream repos also import a list of CodeQL filters
from the `CodeQlFilters.yml` file in this repo.

Given there's modifications to those files in this change, it
is recommended to understand the changes and determine how they
would impact platform execution of CodeQL.

### DIFF
--- a/.pytool/Plugin/CodeQL/MuCodeQlQueries.qls
+++ b/.pytool/Plugin/CodeQL/MuCodeQlQueries.qls
@@ -4,50 +4,132 @@
 - queries: '.'
   from: codeql/cpp-queries
 
-# "Core" Queries
+##########################################################################################
+# "Core" Queries - Part of Core SDL
+##########################################################################################
+
+## Errors
 - include:
     id: cpp/badoverflowguard
+- include:
+    id: cpp/likely-bugs/memory-management/v2/conditionally-uninitialized-variable
+- include:
+    id: cpp/pointer-overflow-check
+- include:
+    id: cpp/unguardednullreturndereference
+
+## Warnings
 - include:
     id: cpp/comparison-with-wider-type
 - include:
     id: cpp/conditionallyuninitializedvariable
 - include:
-    id: cpp/likely-bugs/memory-management/v2/conditionally-uninitialized-variable
-- include:
     id: cpp/paddingbyteinformationdisclosure
-- include:
-    id: cpp/pointer-overflow-check
-- include:
-    id: cpp/redundant-null-check-param
-- include:
-    id: cpp/unguardednullreturndereference
 - include:
     id: cpp/uselesstest
 
-# Additional Fundamental Queries Not Originally Run
+## Recommendations
 - include:
-    id: cpp/infinite-loop-with-unsatisfiable-exit-condition
+    id: cpp/redundant-null-check-param
+
+##########################################################################################
+# Extra Queries - Outside Core SDL
+##########################################################################################
+
+## Enable When Time is Available to Fix Issues
+# Hundreds of issues. Most appear valid. Type: Recommendation.
+#- include:
+#    id: cpp/missing-null-test
+
+## Errors
 - include:
     id: cpp/infiniteloop
 - include:
-    id: cpp/overflow-buffer
+    id: cpp/missing-return
+- include:
+    id: cpp/no-space-for-terminator
+- include:
+    id: cpp/pointer-overflow-check
+- include:
+    id: cpp/redundant-null-check-simple
 - include:
     id: cpp/sizeof/const-int-argument
 - include:
     id: cpp/sizeof/sizeof-or-operation-as-argument
+- include:
+    id: cpp/very-likely-overrunning-write
 
+## Warnings
+- include:
+    id: cpp/comparison-precedence
+- include:
+    id: cpp/implicit-bitfield-downcast
+- include:
+    id: cpp/infinite-loop-with-unsatisfiable-exit-condition
+- include:
+    id: cpp/offset-use-before-range-check
+- include:
+    id: cpp/overflow-buffer
+- include:
+    id: cpp/overflow-calculated
+- include:
+    id: cpp/overflow-destination
+- include:
+    id: cpp/return-stack-allocated-memory
+- include:
+    id: cpp/static-buffer-overflow
+- include:
+    id: cpp/unsigned-comparison-zero
+
+## Recommendations
+- include:
+    id: cpp/missing-header-guard
+- include:
+    id: cpp/unused-local-variable
+- include:
+    id: cpp/unused-static-function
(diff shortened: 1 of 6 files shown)


apop5 (approved):

TaylorBeebe (approved):

.pytool/Plugin/CodeQL/MuCodeQlQueries.qls
```diff
@@ -4,50 +4,144 @@
...
+    id: cpp/likely-bugs/memory-management/v2/conditionally-uninitialized-variable
 - include:
-    id: cpp/unguardednullreturndereference
+    id: cpp/paddingbyteinformationdisclosure
 - include:
-    id: cpp/uselesstest
+    id: cpp/unguardednullreturndereference
+
+##########################################################################################
+# Additional Fundamental Queries
```
VivianNK: What's the criteria for "additional fundamental queries" that differentiate from "core queries" ? Can we get a blurb here like we have in "Extra Queries"?
Also what is the "##Unknown" category? Could it potentially produce an error or warning?

MdeModulePkg/Library/BootMaintenanceManagerUiLib/UpdatePage.c
```diff
@@ -527,9 +527,12 @@ UpdateConsolePage (
         ((NewTerminalContext->IsStdErr != 0) && (UpdatePageId == FORM_CON_ERR_ID))
         )
     {
-      CheckFlags         |= EFI_IFR_CHECKBOX_DEFAULT;
-      ConsoleCheck[Index] = TRUE;
-    } else {
+      CheckFlags |= EFI_IFR_CHECKBOX_DEFAULT;
+
+      if (Index < MAX_MENU_NUMBER) {
```
VivianNK, line 532: Why do we need to check this condition that was asserted on line 524?

VivianNK (approved):
Left a question and suggestion for another comment in the .qls file but lgtm

MdeModulePkg/Library/BootMaintenanceManagerUiLib/UpdatePage.c
makubacki: Asserts are a form of debug aid but cannot be used for ensuring functional correctness in production code.

The `ASSERT()` macro is defined as [this](https://github.com/tianocore/edk2/blob/b240eab03530f063ef5438497d70a731b19a201e/MdePkg/Include/Library/DebugLib.h#L395-L407)  in `DebugLib.h`:

```c
#if !defined (MDEPKG_NDEBUG)
#define ASSERT(Expression)        \
    do {                            \
      if (DebugAssertEnabled ()) {  \
        if (!(Expression)) {        \
          _ASSERT (Expression);     \
          ANALYZER_UNREACHABLE ();  \
        }                           \
      }                             \
    } while (FALSE)
#else
#define ASSERT(Expression)
#endif
```

So, it will be removed entirely if `DebugAssertEnabled()` is `FALSE`.

Every instance of `DebugLib` can implement `DebugAssertEnabled()` differently. Most (non-Null instances) follow the same pattern by depending on `DEBUG_PROPERTY_DEBUG_ASSERT_ENABLED` in `PcdDebugPropertyMask`:

```c
BOOLEAN
EFIAPI
DebugAssertEnabled (
  VOID
  )
{
  return (BOOLEAN)((PcdGet8 (PcdDebugPropertyMask) & DEBUG_PROPERTY_DEBUG_ASSERT_ENABLED) != 0);
}
```

In general, it's very common for different build types (like a `RELEASE` build) to either use [`BaseDebugLibNull`](https://github.com/tianocore/edk2/blob/b240eab03530f063ef5438497d70a731b19a201e/MdePkg/Library/BaseDebugLibNull/DebugLib.c#L155-L162), in which case it is implemented as:

```c
BOOLEAN
EFIAPI
DebugAssertEnabled (
  VOID
  )
{
  return FALSE;
}
```

Or disable asserts in the PCD.

In any case, the code cannot depend on asserts to be active to prevent fatal errors.

.pytool/Plugin/CodeQL/MuCodeQlQueries.qls
makubacki: The first distinction has to do with how query lists are treated internally. I can try to add more details, but it will be limited by that context.

The "Unknown" section came about because I couldn't initially find the severity level details of those after breaking them out, but I have that info now, so I'll move them to other sections.

makubacki: I removed the "Additional Fundamental Queries" section, converging its content with "Extra Queries" and moved "Unknown" queries to the appropriate severity level sections.

PR #558, opened 2023-09-08
Don't Lock Variable Policy if Device is in Unit Test Mode

TaylorBeebe:
## Description

To enable testing variable policy functionality in the shell, it needs to be unlocked when the test starts. This PR adds a check to the system device state and avoids locking if it is in unit test mode.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by running the variable policy test app in Q35.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariablePolicyLockingCommon.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariablePolicyLockingCommon.c
@@ -62,7 +62,7 @@ LockPolicyInterfaceAtReadyToBoot (
 {
   EFI_STATUS  Status;
 
-  if ((GetDeviceState() & DEVICE_STATE_UNIT_TEST_MODE) != 0) {
+  if ((GetDeviceState () & DEVICE_STATE_UNIT_TEST_MODE) != 0) {
     DEBUG ((DEBUG_INFO, "[%a] Unit test mode is enabled. Skipping lock.\n", __FUNCTION__));
     return;
   }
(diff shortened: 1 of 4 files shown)


MdeModulePkg/Universal/Variable/RuntimeDxe/VariablePolicyLockingCommon.c
```diff
@@ -61,6 +62,11 @@ LockPolicyInterfaceAtReadyToBoot (
 {
   EFI_STATUS  Status;
 
+  if ((GetDeviceState () & DEVICE_STATE_UNIT_TEST_MODE) != 0) {
+    DEBUG ((DEBUG_INFO, "[%a] Unit test mode is enabled. Skipping lock.\n", __FUNCTION__));
```
os-d: I would be tempted to put this at DEBUG_ERROR level, because the only time you would be using this is running unit tests, in which case, great, whatever, another print, but if you happened to his this in prod code, that would be a very big error.

I know there has been discussion in this I have not been involved with, but should this be enclosed in DEBUG_CODE so that we don't have any possibility of doing this on a release build? Well, no possibility if the PcdDebugPropertyMask is set correctly.

makubacki: I think `DEBUG_CODE` would be a good addition.

makubacki (approved):
My recollection of the latest discussion around this is that DeviceStateLib is an acceptable dependency so approving given the other comment is addressed.

apop5 (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariablePolicyLockingCommon.c
TaylorBeebe: This call is blocked behind DEBUG_CODE. The print level is unchanged

PR #562, opened 2023-09-12
Add support for initializing Max Payload Size during PCIe enumeration

cfernald:
## Description

PCIe's Max Payload Size controls the maximum packet size that can be handled by a given device. The device's maximum supported MPS is provided by the PCIe Device Capabilities Register, and the current MPS is programmed in the Control Register. For a given Root Port, the MPS should be the same value for all devices and bridges under that Root Port. When the new PCD `PcdPcieInitializeMps` is set to TRUE, the supported MPS will be collected as part of PCI device creation and after all devices are enumerated under the Root Port, all devices and bridges will have their MPS set to the optimum value. 

This allows platforms to ensure that the optimum MPS is set across the PCI tree before PCI devices are started. This can significantly improve performance for PCI devices such as NVMe.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on virtual and physical platforms, OS boot tested.

## Integration Instructions

N/A

 Co-authored-by: abrahamg <abrahamg@microsoft.com>


MdeModulePkg/Bus/Pci/PciBusDxe/PciLib.c
```diff
@@ -1802,6 +1802,18 @@ PciHostBridgeEnumerator (
       return Status;
     }
 
+    // MU_CHANGE BEGIN: Add support for initializing PCIe MPS
+    if (PcdGetBool (PcdPcieInitializeMps) && gFullEnumeration) {
+      UINT8  MaxPayloadSize;
+      Status =  PciGetMaxPayloadSize (RootBridgeDev, &MaxPayloadSize);
+      if (!EFI_ERROR (Status)) {
+        Status =  PciProgramMps (RootBridgeDev, MaxPayloadSize);
+        ASSERT (!EFI_ERROR (Status));
```
apop5: Can this be turned into a error message and allow boot to continue?

MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
```diff
@@ -207,6 +207,155 @@ PciPciDeviceInfoCollector (
...
+
+      Status = PciIoDevice->PciIo.Pci.Write (
+                                        &PciIoDevice->PciIo,
+                                        EfiPciIoWidthUint16,
+                                        PciIoDevice->PciExpressCapabilityOffset + EFI_PCIE_CAPABILITY_DEVICE_CONTROL,
+                                        1,
+                                        &DeviceControl.Uint16
+                                        );
+      if (EFI_ERROR (Status)) {
+        return Status;
```
apop5: Let's say that a root bridge has another bridge behind it, and beyond that bridge, there are two devices. One device fails to program max payload (I don't see how that would happen because most pci devices just discard writes to read only registers, but for the sake of this scenario, let's say it does). We would leave one additional device that would not have the correct payload size.  Maybe we should output a debug message and continue going through the child devices?

An an ascii example the confusing paragraph above:
Rp0 - Bridge0 - device 1
              - device 2 <- write error
              - device 3 <- skipped by return

apop5 (approved):

MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
cfernald: Yeah, I considered the failure scenarios for a bit, but I'm not sure the full implications of this in practice. I'm going to talk to the OS folks a bit on this. The alternative would be to re-set all values down to the baseline 0 value.

cfernald: Looking into it more, this should only happen if the device's information is bad or we are in the critical failure path anyways. I will log an error and continue on  the other devices.

MdeModulePkg/Bus/Pci/PciBusDxe/PciEnumeratorSupport.c
```diff
@@ -207,6 +207,161 @@ PciPciDeviceInfoCollector (
...
+        __FUNCTION__,
+        PciIoDevice->BusNumber,
+        PciIoDevice->DeviceNumber,
+        PciIoDevice->FunctionNumber,
+        Status
+        ));
+    } else if (DeviceControl.Bits.MaxPayloadSize != MaxPayloadSize) {
+      DeviceControl.Bits.MaxPayloadSize = MaxPayloadSize;
+      DEBUG ((
+        DEBUG_INFO,
```
os-d: nit: Probably can be verbose?

os-d (approved):

PR #581, opened 2023-09-29
Add FlatPageTableLib

TaylorBeebe:
## Description

Creates a new library, FlatPageTableLib, which works on X64 and AARCH64 platforms and converts the page table to a "flat" version. The flat version is a one-dimensional array where each entry is an address, a length, and attributes. The library will walk the page/translation table and combine blocks/leaves with the same attributes into a single entry in the flat array. The attributes mask for each architecture is defined in the header and includes both the upper and lower block/leaf attributes. On both X64 and AARCH64, the hierarchical inheritance of attributes is factored into the determination of block/leaf attributes. This allows the consumer of the library to easily check the attributes of any region in the page/translation table.

A future update will add "map" functionality which will allow the user to manipulate the attributes of the page/translation table with correct hierarchical updates. That functionality will require allocation to work because spitting blocks larger than 4K will require additional table memory.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by running the paging audit on SBSA and Q35 and by comparing the output against the Memory Attribute Protocol

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Include/Library/FlatPageTableLib.h
+++ b/MdeModulePkg/Include/Library/FlatPageTableLib.h
@@ -27,6 +27,53 @@ typedef struct {
 #define AARCH64_PAGE_MAP_SIGNATURE  SIGNATURE_32 ('A','A','6','4')
 #define X64_PAGE_MAP_SIGNATURE      SIGNATURE_32 ('X','6','4',' ')
 
+// This union can be used to interpret each PAGE_MAP_ENTRY on an AARCH64 system.
+typedef union {
+  struct {
+    UINT64    Valid                      : 1;  // BIT0
+    UINT64    BlockOrTable               : 1;  // BIT1
+    UINT64    AttributeIndex             : 3;  // BIT2-4
+    UINT64    NonSecure                  : 1;  // BIT5
+    UINT64    AccessPermissions          : 2;  // BIT6-7
+    UINT64    Sharability                : 2;  // BIT8-9
+    UINT64    AccessFlag                 : 1;  // BIT10
+    UINT64    NonGlobal                  : 1;  // BIT11
+    UINT64    Oa                         : 4;  // BIT12-15
+    UINT64    Nt                         : 1;  // BIT16
+    UINT64    OutputAddress              : 33; // BIT17-49
+    UINT64    Guarded                    : 1;  // BIT50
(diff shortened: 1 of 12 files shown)


MdeModulePkg/Library/FlatPageTableLib/X64/FlatPageTableX64.c
```diff
@@ -0,0 +1,108 @@
...
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Map->ArchSignature = X64_PAGE_MAP_SIGNATURE;
+
+  // Poll CR4 to deterimine the page table depth
+  Cr4.UintN = AsmReadCr4 ();
+
+  if (Cr4.Bits.LA57 != 0) {
+    PagingMode = Paging5Level;
```
VivianNK, line 51: Where do these Paging5/4Level values come from?

MdeModulePkg/Include/Library/FlatPageTableLib.h
```diff
@@ -0,0 +1,166 @@
+/** @file
```
VivianNK, line 1: nit: Honestly not sure if we follow this but for the comment documentation tags @file, they should be followed by the filename.
e.g.
@file FlatPageTableLib.h

Or do we use a different system?

MdeModulePkg/Library/FlatPageTableLib/X64/FlatPageTableX64.c
TaylorBeebe: The definitions come from CpuPageTableLib

MdeModulePkg/Include/Library/FlatPageTableLib.h
TaylorBeebe: In EDK2 headers, I only see @file. In some Project Mu headers, I see @file <filename>. I'll have to check if the EDK2 code format guidelines weigh in on this :)

TaylorBeebe:
Moved to mu_plus: https://github.com/microsoft/mu_plus/pull/322

PR #582, opened 2023-09-29
Add a Panic Library to MdePkg

kenlautner:
## Description

Added a panic library to MdePkg.  This provides an API to use for when an unrecoverable error is hit during boot.  It will print a message to the debug console and have the system deadloop indefinitely.

It can be used by including the BasePanicLib implementation in your module or library and calling `PANIC`.  You can input your own custom message to indicate what's triggering the `PANIC`.  

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Intel physical devices.  Confirmed that when a `PANIC` is hit we get the expected message and we halt boot.

## Integration Instructions

Include the panic library in your platform dsc.


MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,78 @@
...
+  of size reduction when compiler optimization is disabled. If MDEPKG_NDEBUG is
+  defined, then panic related macros wrapped by it are the NULL implementations.
+
+Copyright (c) 2023, Microsoft Corporation. All rights reserved.<BR>
+SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef __PANIC_LIB_H__
+#define __PANIC_LIB_H__
```
os-d:
```suggestion
#ifndef PANIC_LIB_H_
#define PANIC_LIB_H_

```

nit: This seems to be the more canonical style.

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,78 @@
...
+/**
+  Macro that calls PanicReport().
+
+  If MDEPKG_NDEBUG is not defined PanicReport() is called passing
+  in the source filename, source line number, and Message.
+
+  @param  Expression  A format string
+
+**/
+#if !defined (MDEPKG_NDEBUG)
```
os-d: I guess this is a question on the panic library, whether it is for debug only scenarios or it is used to say, this is too dangerous to continue operation, even in a release build. If it is tied to NDEBUG, it seems to have less value over an ASSERT (if you want an ASSERT and then a deadloop, that is already an option).

If you wanted null panics, then presumably you would use the BasePanicLibNull, which sure you would have calls into and include the code, but I thought that was the value of panic and if it is tied to NDEBUG the platform maintainer would have to choose between keeping panics and also keeping all the other debug code controlled by NDEBUG, or not having panics.

MdePkg/Library/BasePanicLib/BasePanicLib.inf
```diff
@@ -0,0 +1,35 @@
+## @file
+#  Panic Library.
+#
+#  Copyright (c) 2023, Microsoft Corporation. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+#
```
os-d: nit: delete extra line

MdePkg/Library/BasePanicLib/BasePanicLib.inf
```diff
@@ -0,0 +1,35 @@
...
+  BASE_NAME                      = BasePanicLib
+  MODULE_UNI_FILE                = BasePanicLib.uni
+  FILE_GUID                      = DF456972-AF4A-4F3A-B33C-DD31368BF490
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = PanicLib
+
+
+#
+#  VALID_ARCHITECTURES           = IA32 X64 EBC
```
os-d: I'd assume this would be ANY?

MdePkg/Library/BasePanicLib/PanicLib.c
```diff
@@ -0,0 +1,57 @@
...
+
+  //
+  // Generate the PANIC message in ASCII format
+  //
+  AsciiSPrint (Buffer, sizeof (Buffer), "PANIC [%a] %a(%d): %a\n", gEfiCallerBaseName, FileName, LineNumber, Description);
+
+  //
+  // Send the print string to the Logging device device
+  //
+  SerialPortWrite ((UINT8 *)Buffer, AsciiStrLen (Buffer));
```
os-d: I think we need to call `SerialPortInitialize()` before this (which could be in the library constructor for PanicLib, although we are only ever going to enter this once, so calling init here is probably fine).

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.inf
```diff
@@ -0,0 +1,30 @@
+## @file
+#  Panic Library with empty functions.
+#
+#  Copyright (c) 2023, Microsoft Corporation. All rights reserved.<BR>
+#
+#  SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+#
```
os-d: nit: extra line can be removed

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.inf
```diff
@@ -0,0 +1,30 @@
...
+  BASE_NAME                      = BasePanicLibNull
+  MODULE_UNI_FILE                = BasePanicLibNull.uni
+  FILE_GUID                      = 0A097ABB-3869-4CF8-B5C0-AE798F135C46
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = PanicLib
+
+
+#
+#  VALID_ARCHITECTURES           = IA32 X64 EBC
```
os-d:
```suggestion
#  VALID_ARCHITECTURES           = ANY

```

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.uni
```diff
@@ -0,0 +1,16 @@
+// /** @file
+// Panic Library with empty functions.
+//
+// Panic Library with empty functions.
```
os-d:
```suggestion
```

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.uni
```diff
@@ -0,0 +1,16 @@
+// /** @file
+// Panic Library with empty functions.
+//
+// Panic Library with empty functions.
+//
+// Copyright (c) 2023, Microsoft Corporation. All rights reserved.<BR>
```
os-d: I think that all the license text should just be:

```suggestion
// Copyright (c) Microsoft Corporation.

```

But you can confirm.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.uni
```diff
@@ -0,0 +1,16 @@
...
+//
+// SPDX-License-Identifier: BSD-2-Clause-Patent
+//
+// **/
+
+
+#string STR_MODULE_ABSTRACT             #language en-US "Panic Library with empty functions"
+
+#string STR_MODULE_DESCRIPTION          #language en-US "Panic Library with empty functions"
```
os-d: Probably don't need the uni files at all?

MdePkg/Library/BasePanicLibNull/PanicLib.c
```diff
@@ -0,0 +1,40 @@
...
+#include <Library/PanicLib.h>
+
+//
+// Define the maximum panic message length that this library supports
+//
+#define MAX_PANIC_MESSAGE_LENGTH  0x100
+
+/**
+  Prints a panic message containing a filename, line number, and description.
+  This is always followed by a dead loop.
```
os-d: This statement isn't true in the null case, probably worth calling out.

MdePkg/MdePkg.dec
```diff
@@ -293,6 +293,13 @@
   #
   ArmTrngLib|Include/Library/ArmTrngLib.h
 
+# MU_CHANGE [START]
+  ##  @libraryclass  Provides services panic functionality.
```
os-d:
```suggestion
  ##  @libraryclass  Provides panic functionality.

```

MdePkg/Include/Library/PanicLib.h
makubacki: And required by PatchCheck now.

kenlautner: Thanks.  Updated.

MdePkg/Include/Library/PanicLib.h
kenlautner: I accidentally left this in so I removed it.

MdePkg/Library/BasePanicLib/BasePanicLib.inf
kenlautner: Deleted

MdePkg/Library/BasePanicLib/BasePanicLib.inf
kenlautner: Updated.

MdePkg/Library/BasePanicLib/PanicLib.c
kenlautner: Got it.  Added.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.uni
kenlautner: I wasn't sure so I included them.  I removed them with my latest commit.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.uni
kenlautner: I removed the file with my latest commit.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.uni
kenlautner: I forgot to update these before publishing the PR.  Thanks for pointing it out.

MdePkg/Library/BasePanicLibNull/PanicLib.c
kenlautner: Updated for the NULL implementation.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.inf
kenlautner: Updated.

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
...
+#if defined (__clang__) && defined (__FILE_NAME__)
+#define _PANIC(Message)  PanicReport (__FILE_NAME__, PANIC_LINE_NUMBER, Message)
+#else
+#define _PANIC(Message)  PanicReport (__FILE__, PANIC_LINE_NUMBER, Message)
+#endif
+
+/**
+  Macro that calls PanicReport().
+
+  If MDEPKG_NDEBUG is not defined PanicReport() is called passing
```
os-d: This should be removed now that you've dropped the MDEPKG_NDEBUG

os-d (approved):

MdePkg/MdePkg.dec
kenlautner: Fixed.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.inf
kenlautner: Removed.

MdePkg/Include/Library/PanicLib.h
kenlautner: Fixed.

MdePkg/Include/Library/PanicLib.h
makubacki: I think panic lib should simply panic or not panic (NULL instance). It should be used in select cases where it doesn't need to be optimized for size to the degree DebugLib is.

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
+/** @file
+  Provides services to print panic messages to a debug output device.
+
+  Note that a reserved macro named MDEPKG_NDEBUG is brought from debuglib for the intention
+  of size reduction when compiler optimization is disabled. If MDEPKG_NDEBUG is
+  defined, then panic related macros wrapped by it are the NULL implementations.
```
makubacki: Remove this as well.

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
+/** @file
```
makubacki, line 1: The header should describe what "panic" means and the cases when panic APIs should be used.

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
...
+
+  Print a message of the form "PANIC <FileName>(<LineNumber>): <Description>\n"
+  to the debug output device.  Immediately after that CpuDeadLoop() is called.
+
+  If FileName is NULL, then a <FileName> string of "(NULL) Filename" is printed.
+  If Description is NULL, then a <Description> string of "(NULL) Description" is printed.
+
+  @param  FileName     The pointer to the name of the source file that generated the panic condition.
+  @param  LineNumber   The line number in the source file that generated the panic condition
+  @param  Description  The pointer to the description of the panic condition.
```
makubacki:
```suggestion
  @param[in]  FileName     The pointer to the name of the source file that generated the panic condition.
  @param[in]  LineNumber   The line number in the source file that generated the panic condition
  @param[in]  Description  The pointer to the description of the panic condition.

```

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
...
+PanicReport (
+  IN CONST CHAR8  *FileName,
+  IN UINTN        LineNumber,
+  IN CONST CHAR8  *Description
+  );
+
+//
+// Source file line number.
+// Default is use the to compiler provided __LINE__ macro value. The __LINE__
+// mapping can be overriden by predefining PANIC_LINE_NUMBER
```
makubacki, line 40:
```suggestion
// Default is to use the compiler provided __LINE__ macro value. The __LINE__
// mapping can be overriden by predefining PANIC_LINE_NUMBER

```

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
...
+//
+// Defining PANIC_LINE_NUMBER to a fixed value is useful when comparing builds
+// across source code formatting changes that may add/remove lines in a source
+// file.
+//
+#ifdef PANIC_LINE_NUMBER
+#else
+#define PANIC_LINE_NUMBER  __LINE__
+#endif
```
makubacki: Can consolidate logic.

```suggestion
#ifndef PANIC_LINE_NUMBER
#define PANIC_LINE_NUMBER  __LINE__
#endif

```

MdePkg/Include/Library/PanicLib.h
```diff
@@ -0,0 +1,74 @@
...
+
+#if defined (__clang__) && defined (__FILE_NAME__)
```
makubacki, line 54: Can you add a description comment above this block like the others?

MdePkg/Library/BasePanicLib/BasePanicLib.inf
```diff
@@ -0,0 +1,32 @@
+## @file
+#  Panic Library.
+#
+#  Copyright (c) Microsoft Corporation.
+#
```
makubacki: Need SPDX identifier.

```suggestion
# SPDX-License-Identifier: BSD-2-Clause-Patent

```

MdePkg/Library/BasePanicLib/PanicLib.c
```diff
@@ -0,0 +1,59 @@
+/** @file
+  Base Panic Library.
+
+  Copyright (c) Microsoft Corporation.
+
```
makubacki: Need SPDX identifier.

```suggestion
  SPDX-License-Identifier: BSD-2-Clause-Patent

```

MdePkg/Library/BasePanicLib/PanicLib.c
```diff
@@ -0,0 +1,59 @@
...
+{
+  CHAR8  Buffer[MAX_PANIC_MESSAGE_LENGTH];
+
+  //
+  // Generate the PANIC message in ASCII format
+  //
+  AsciiSPrint (Buffer, sizeof (Buffer), "PANIC [%a] %a(%d): %a\n", gEfiCallerBaseName, FileName, LineNumber, Description);
+
+  // Initialize Serial Port to write the panic message
+  SerialPortInitialize ();
```
makubacki: Why is this going to `SerialPortLib`?

A generic interface like `DebugLib` allows flexibility to hook into a number of backends including serial port through an instance like `BaseDebugLibSerialPort` or something also generic like the status code router via `PeiDxeDebugLibReportStatusCode`.

MdePkg/Library/BasePanicLib/PanicLib.c
```diff
@@ -0,0 +1,59 @@
...
+VOID
+EFIAPI
+PanicReport (
+  IN CONST CHAR8  *FileName,
+  IN UINTN        LineNumber,
+  IN CONST CHAR8  *Description
+  )
+{
+  CHAR8  Buffer[MAX_PANIC_MESSAGE_LENGTH];
+
```
makubacki: Need to validate function pointers before dereferencing.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.inf
```diff
@@ -0,0 +1,27 @@
+## @file
+#  Panic Library with empty functions.
+#
+#  Copyright (c) Microsoft Corporation.
+#
```
makubacki, line 6: Need SPDX identifier.

```suggestion
#   SPDX-License-Identifier: BSD-2-Clause-Patent
```

MdePkg/Library/BasePanicLibNull/PanicLib.c
```diff
@@ -0,0 +1,33 @@
+/** @file
+  Base Panic Library with empty functions.
+
+  Copyright (c) Microsoft Corporation.
+
```
makubacki: Need SPDX identifier.

```suggestion
    SPDX-License-Identifier: BSD-2-Clause-Patent
```

MdePkg/Library/BasePanicLib/PanicLib.c
kenlautner: I just made it work with SerialPortLib.  To be generic I should rename rename this implementation to BasePanicLibSerialPort. I don't think anything else would need to be changed.

MdePkg/Library/BasePanicLibNull/PanicLib.c
```diff
@@ -0,0 +1,33 @@
...
+**/
+
+#include <Base.h>
+#include <Library/PanicLib.h>
+
+//
+// Define the maximum panic message length that this library supports
+//
+#define MAX_PANIC_MESSAGE_LENGTH  0x100
```
makubacki: Was this needed in this instance?

MdePkg/Library/BasePanicLib/PanicLib.c
makubacki: Why not use a more generic interface?

MdePkg/Include/Library/PanicLib.h
kenlautner: Updated.

MdePkg/Include/Library/PanicLib.h
kenlautner: Updated.

MdePkg/Include/Library/PanicLib.h
kenlautner: Added.

MdePkg/Library/BasePanicLibNull/BasePanicLibNull.inf
kenlautner: Fixed.

MdePkg/Library/BasePanicLibNull/PanicLib.c
kenlautner: Fixed.

MdePkg/Include/Library/PanicLib.h
kenlautner: Removed.

MdePkg/Include/Library/PanicLib.h
kenlautner: Fixed.

MdePkg/Include/Library/PanicLib.h
kenlautner: Updated.

MdePkg/Library/BasePanicLib/BasePanicLib.inf
kenlautner: Fixed.

MdePkg/Library/BasePanicLib/PanicLib.c
kenlautner: Fixed.

MdePkg/Library/BasePanicLibNull/PanicLib.c
kenlautner: It wasn't.  Removed.

MdePkg/Library/BasePanicLib/PanicLib.c
kenlautner: I added checks for the inputs.

MdePkg/Library/BasePanicLib/PanicLib.c
makubacki: If you're explicitly calling this a serial port lib instance, it's fine to keep.

makubacki (approved):

MdePkg/Library/BasePanicLib/PanicLib.c
kenlautner: I understand the generic interface approach but for one function I'm not sure it's worth it over just creating an implementation for each backend.  If you see additional value from the generic interface that I'm missing then I can add it.

MdePkg/Library/BasePanicLibSerialPort/PanicLib.c
```diff
@@ -0,0 +1,68 @@
...
+    Description = "(NULL) Description";
+  }
+
+  //
+  // Generate the PANIC message in ASCII format
+  //
+  AsciiSPrint (Buffer, sizeof (Buffer), "PANIC [%a] %a(%d): %a\n", gEfiCallerBaseName, FileName, LineNumber, Description);
+
+  // Initialize Serial Port to write the panic message
+  SerialPortInitialize ();
```
kuqin12: nit: why do we want to initialize it here?

kuqin12 (approved):

MdePkg/Library/BasePanicLib/PanicLib.c
makubacki: People have already invested in implementing code for backends they care about under long time abstractions like `DebugLib` and the report status code router. This allows those to be reused with minimal effort. Instead of creating a new `PanicLib` instance for every output instance that might be needed, they can simply link their code that already does that (e.g. `BaseDebugLibSerialPort`) and output goes where it needs to.

However, I think starting with an instance that links against `SerialPortLib` is fine for a couple reasons.

1. Library instances link at the module level and overriding library instances for libraries is somewhat convoluted.
2. This library has the distinction of being used in scenarios when the system may be in a very incapacitated state.

Most platforms will define a `PanicLib` instance in their DSC and consider `PanicLib` done. They will also define something like a `DebugLib` that covers each phase and consider that done.

If a `PanicLib` instance depends on `DebugLib`, the platform integrator must now consider all of the modules that link `PanicLib` and determine if the default linked `DebugLib` instance makes sense in panic scenarios. It might not. In normal cases, it may make sense to route output through a more complex stack (e.g. console out), whereas something simpler, like lighting an LED or sending a POST code is more appropriate in a panic.

In the end, another `PanicLib` instance could be created (i.e. `BasePanicLibDebugLib`) and developers that understand how to use it could, but I think serial is a reasonable starting point that would likely end up being the most common choice so it's a fine.

I do suggest updating the name of the .c files to match the .inf files per convention - `BasePanicLibSerialPort.c`.

---

The other issue that sometimes comes up with `SerialPortLib` is that it gets used in places like this and then completely misused to route places other than serial ports because of its ubiquity (kind of like the UEFI variable interface). But that's outside the scope of this PR.

MdePkg/Library/BasePanicLibSerialPort/PanicLib.c
kenlautner: There's a chance that serial ports are not initialized when PANIC is called so I call SerialPortInitialize to make sure the service is available.

PR #599, opened 2023-10-19
MdeModulePkg/PciHostBridgeDxe: Add readback after final Cfg-Write

JoeLopez333:
## Description

Problem Report:

On AARCH64, there is no ordering guarantee between configuration
space (ECAM) writes and memory space reads (MMIO). ARM AMBA CHI
only guarantees ordering for reads and writes within a single address region,
however, on some systems MMIO and ECAM may be split into separate
address regions.

A problem may arise when an ECAM write is issued a completion before a subsequent
MMIO read is issued and receives a completion.

For example, a typical PCI software flow is the following:

1. ECAM write to device command register to enable memory space
2. MMIO read from device memory space for which access was enabled
   in step 1.

There is no guarantee that step 2. will not begin before the completion of step 1.
on systems where ECAM/MMIO are specified as separate address regions, even
if both spaces have the memory attributes device-nGnRnE.

- Add a read after the final PCI Configuration space write
in RootBridgeIoPciAccess. Configuration space reads should not have side-efects.

- When configuration space is strongly ordered, this ensures
that program execution cannot continue until the completion
is received for the previous Cfg-Write, which may have side-effects.

- Risk of reading a "write-only" register and causing a CA which leaves the device
unresponsive. The expectation based on the PCI Base Spec v6.1 section 7.4 is that
all PCI Spec-defined registers will be readable, however, there may exist 
design-specific registers that fall into this category.

- [X] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

AARCH64 platform on real silicon

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
+++ b/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
@@ -1246,6 +1246,14 @@ RootBridgeIoPciAccess (
     }
   }
 
+  //
+  // Perform readback after write to confirm completion was received for the last write
+  // before subsequent memory operations can be issued.
+  //
+  if (!Read) {
+  PciSegmentRead8 (Address - InStride);
+  }
+
   return EFI_SUCCESS;
 }
 
--- a/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
+++ b/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
@@ -1254,6 +1254,7 @@ RootBridgeIoPciAccess (
   if (!Read) {
     PciSegmentRead8 (Address - InStride);
   }
+
   // MU_CHANGE END
 
   return EFI_SUCCESS;
--- a/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
+++ b/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
@@ -1246,6 +1246,7 @@ RootBridgeIoPciAccess (
     }
   }
 
+  // MU_CHANGE BEGIN
   //
   // Perform readback after write to confirm completion was received for the last write
   // before subsequent memory operations can be issued.
@@ -1253,6 +1254,7 @@ RootBridgeIoPciAccess (
   if (!Read) {
   PciSegmentRead8 (Address - InStride);
   }
+  // MU_CHANGE END
 
   return EFI_SUCCESS;
 }
--- a/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
+++ b/MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
@@ -1252,7 +1252,7 @@ RootBridgeIoPciAccess (
   // before subsequent memory operations can be issued.
   //
   if (!Read) {
-  PciSegmentRead8 (Address - InStride);
+    PciSegmentRead8 (Address - InStride);
   }
   // MU_CHANGE END
 
--- a/MdePkg/Library/BaseIoLibIntrinsic/IoLibNoIo.c
+++ b/MdePkg/Library/BaseIoLibIntrinsic/IoLibNoIo.c
@@ -412,7 +412,11 @@ MmioRead8 (
 
   Flag = FilterBeforeMmIoRead (FilterWidth8, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     Value = *(volatile UINT8 *)Address;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoRead (FilterWidth8, Address, &Value);
@@ -444,7 +448,11 @@ MmioWrite8 (
 
   Flag = FilterBeforeMmIoWrite (FilterWidth8, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     *(volatile UINT8 *)Address = Value;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoWrite (FilterWidth8, Address, &Value);
@@ -479,7 +487,11 @@ MmioRead16 (
 
   Flag = FilterBeforeMmIoRead (FilterWidth16, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     Value = *(volatile UINT16 *)Address;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoRead (FilterWidth16, Address, &Value);
@@ -513,7 +525,11 @@ MmioWrite16 (
 
   Flag = FilterBeforeMmIoWrite (FilterWidth16, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     *(volatile UINT16 *)Address = Value;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoWrite (FilterWidth16, Address, &Value);
@@ -548,7 +564,11 @@ MmioRead32 (
 
   Flag = FilterBeforeMmIoRead (FilterWidth32, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     Value = *(volatile UINT32 *)Address;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoRead (FilterWidth32, Address, &Value);
@@ -582,7 +602,11 @@ MmioWrite32 (
 
   Flag = FilterBeforeMmIoWrite (FilterWidth32, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     *(volatile UINT32 *)Address = Value;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoWrite (FilterWidth32, Address, &Value);
@@ -617,7 +641,11 @@ MmioRead64 (
 
   Flag = FilterBeforeMmIoRead (FilterWidth64, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     Value = *(volatile UINT64 *)Address;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoRead (FilterWidth64, Address, &Value);
@@ -651,7 +679,11 @@ MmioWrite64 (
 
   Flag = FilterBeforeMmIoWrite (FilterWidth64, Address, &Value);
   if (Flag) {
+    MemoryFence ();
+
     *(volatile UINT64 *)Address = Value;
+
+    MemoryFence ();
   }
 
   FilterAfterMmIoWrite (FilterWidth64, Address, &Value);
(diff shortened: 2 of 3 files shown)


kuqin12:
I think the change is acceptable for ARM targets. But I am not sure if the same change will impact too much performance for other architectures using this file and thus impacting the upstreaming effort.

makubacki:
> I think the change is acceptable for ARM targets. But I am not sure if the same change will impact too much performance for other architectures using this file and thus impacting the upstreaming effort.

I've only seen `DMB` referred to in Arm architecture but this is using the `MemoryFence()` function which is abstracted across architectures to barrier types. So, it's not DMB-specific at the moment as the current title would suggest.

The PR description needs more information. It would be helpful for it to link to resources like https://developer.arm.com/documentation/100941/0101/Barriers in the PR description and then tie that information back to the architectural need so there's more context for the change.

Arch-specific macros are defined that can be used like `#if defined (MDE_CPU_AARCH64)...`.

JoeLopez333:
@makubacki @kuqin12 I agree with both your statements, I have altered the title and description to make the intention of the PR more along the lines of "create parity between IoLib.c and IoLibNoIo.c"

I'm curious why MemoryFence() was added as a wrapper around operations in IoLib.c but not the other. Is the intention that developers working on Arm/Risc-v/etc... platforms will not rely on the IoLib to ensure memory ordering but instead implement their own versions of FilterAfterMmIoRead/FilterAfterMmIoWrite which may or may not add the memory barrier? Maybe it makes sense to leave the choice of using operations like DMB up to the developer rather than forcing the protection by adding MemoryFence()

makubacki:
> @makubacki @kuqin12 I agree with both your statements, I have altered the title and description to make the intention of the PR more along the lines of "create parity between IoLib.c and IoLibNoIo.c"
> 
> Is the intention that developers working on Arm/Risc-v/etc... platforms will not rely on the IoLib to ensure memory ordering but instead implement their own versions of FilterAfterMmIoRead/FilterAfterMmIoWrite which may or may not add the memory barrier? Maybe it makes sense to leave the choice of using operations like DMB up to the developer rather than forcing the protection by adding MemoryFence()

1. The `Filter<operation>()` APIs are really an abstraction layer from `RegisterFilterLib` intended for system emulation environments. `Before` provides an opportunity to skip the device access (devices might not be emulated, etc.). `After` is intended for tracing accesses. Some platforms use this to audit resource access patterns this ends up being used for activities like ensuring legacy accesses are not present when refactoring code, optimizing software/hardware based on access frequencies, and so on.

   As a general matter of single responsibility/cohesion, I suggest not putting anything behind an implementation of that library class that does not perform those responsibilities. It will make the implementation brittle as it lies under an unrelated abstraction.

> I'm curious why MemoryFence() was added as a wrapper around operations in IoLib.c but not the other.

2. It's been in the actual IO code for 15 years. I imagine it was added to handle things non-temporal writes that relax normal memory ordering.

3. I'm not very familiar with Arm architecture. But I was reading [Sub-types of Device Memory](https://developer.arm.com/documentation/102376/0200/Device-memory/Sub-types-of-Device). It seems like device memory ranges that would not want the normal weakly ordered memory model would mark the range with the `nR` (not re-reordered) type.

   How does this change fit into that? I.e. does it still need to apply to some case when memory is `Device-nR`? Is it meant to overlap the `Device-R` cases with `Device-nR`?

TaylorBeebe:
According to the UEFI spec, EFI_MEMORY_UC on AARCH64 maps to Device-nGnRnE and PciHostBridgeDxe will apply EFI_MEMORY_UC to all mapped MMIO space.

This PR may solve a real problem, but it might be worth investigating either why memory with the EFI attribute EFI_MEMORY_UC still has weak ordering or why some MMIO space is not being mapped as device memory.

makubacki:
@JoeLopez333 can you provide more details about the device and memory types involved for the case being addressed in this PR?

JoeLopez333:
Hi @makubacki @kuqin12 @TaylorBeebe sorry for the late reply, we have been working this out internally and talking with an Arm representative.

The original issue that I am trying to fix is on an Arm64 system with a scenario of a Cfg-Write before a Mem-Read. The Cfg-Write is the the "Memory Space Enable" bit in an endpoint's configuration space. After the Cfg-Write completes, UEFI should be able to access the endpoint's memory space and read valid data.

We were seeing that the core was moving on to the Mem-Read even though the Cfg-Write was not entirely completed. This was found to be due to the fact that the CMN IO coherent Home Node (HN-I) was configured to split the MMIO and ECAM spaces into separate SAM regions.

Exact wording from the representative from Arm:
> "When they are to two different order regions, the read will not wait for the write to complete, and can return data before the write does anything. The HN-I only preserves ordering between reads and writes to the same Order Region (which implies the same Address Region). Likewise, the HN-I will only preserve ordering between multiple reads and between multiple writes within the same Order Region, and it accomplishes this by issuing the reads with the same ARID and the writes with the same AWID (i.e. it relies on the downstream device to follow AXI ordering rules). 
Issuing a CHI request with REQ.Order=EndpointOrder only guarantees ordering to the same “endpoint address range,” which the HN-I defines as an Order Region (within an Address Region)."

> "Our CMN TRM showcases an example where ECAM and MMIO are two different regions in the HN-I SAM. The implication is that we would expect a DSB between the ECAM write and MMIO read. 
I'm asking our Open Source Software group to confirm that standard PCIe software is generally expected to be aware of the need for a DSB--but my impression from talking to some of our hardware engineers is that that is indeed the expectation.
Let me know if further clarification is needed."

Here is another relavant snippet from the PCIe AMBA Integration Guide (ARM DEN0114)
image

If ECAM and MMIO are set to two different regions in the HN-I SAM, there can never be a guarantee of ordering, even if both regions are configured as Device-nGnRnE. Therefore a memory barrier is needed to guarantee ordering after all Cfg-Writes, and potentially after Mem-Writes / Cfg-Reads as well

makubacki:
> Hi @makubacki @kuqin12 @TaylorBeebe sorry for the late reply, we have been working this out internally and talking with an Arm representative.
> 
Thanks, this is the level of detail I was looking for.

At this point, I believe the change has transcended existence as a Project Mu override and should be reviewed on the edk2 mailing list by Arm code owners to ensure the change meets their expectation and understanding. Then we can immediately cherry-pick it back.

If you're not familiar with the code contribution process, I recommend:

- Sign up for the mailing list. Click the blue "Join this group" button here - https://edk2.groups.io/g/devel.
  - It's recommended to use a non-Microsoft email since IT email security techniques mangle messages.
- Once approved, send an email to the mailing list (`devel@edk2.groups.io`) where you state the problem and the proposed change. You can prefix the subject of the email with `[RFC]`. For example, `[RFC] Order Arm ECAM and MMIO Operations in IoLib`.
  - You can include a code example in your email of course.
- Once the conversation reaches the point that implementation is agreed upon, we can help with the process to submit the patch to the mailing list.

JoeLopez333:
Hi @makubacki, thanks for the suggestion. I have initiated the conversation in the edk2 devel group. The post is currently pending approval.

If possible I would like to merge the changes to MU_BASECORE so that we can consume them while pending approval and discussion from edk2.

kuqin12 (approved):

makubacki:
> Hi @makubacki, thanks for the suggestion. I have initiated the conversation in the edk2 devel group. The post is currently pending approval.

The post is publicly visible and active - https://edk2.groups.io/g/devel/message/110456

> If possible I would like to merge the changes to MU_BASECORE so that we can consume them while pending approval and discussion from edk2.

Given this has been open for 2 weeks, I'd like to wait a couple days, if possible, to determine where that discussion goes. If it's an immediate priority, you can create a local copy of the library with the change then remove & swap that with the version here when it is available.

JoeLopez333:
Per offline conversation with @makubacki, updating this PR to reflect the latest iteration tracking the upstream change to edk2 and marking as ready for review

makubacki:
+@joschock

@JoeLopez333, @joschock and I discussed the possibility of a device that has a write-only cfg register that URs on read. In that case, universally reading after could cause the device (endpoint) to decide that the read violates the device programming model leading to a completer abort and the device becoming unresponsive.

pci

We're not sure if devices that exist that map such registers to cfg space, but can you include a mention of this in your PR description, so it gets rolled up into the final commit message? Then it will be captured in history in case someone ever sees such an error and comes back to the commit.

MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
```diff
@@ -1246,6 +1246,16 @@ RootBridgeIoPciAccess (
     }
   }
 
+  // MU_CHANGE BEGIN
+  //
+  // Perform readback after write to confirm completion was received for the last write
+  // before subsequent memory operations can be issued.
+  //
+  if (!Read) {
+  PciSegmentRead8 (Address - InStride);
```
makubacki: Indent code in the conditional block two spaces.

JoeLopez333:
@makubacki @joschock will add a mention of the risk to the PR comment. It should also be noted that the In the PCI Base Spec version 6.1 section 7.4 "Configuration Register Types" all configuration space registers are assigned one of the following attributes (quoting Michael Brown in the edk2 discussion thread)
 
> HwInit - read-only, no read side effects
> 
> RO - read-only, no read side effects
> 
> RW - read-write, no read side effects
> 
> RW1C - write 1 to clear bits, no read side effects
> 
> ROS - read-only, no read side effects
> 
> RWS - read-write, no read side effects
> 
> RW1CS - write 1 to clear bits, no read side effects
> 
> RsvdP - read-write, no read side effects
> 
> RsvdZ - read-write, no read side effects

From this I don't think we can expect that there are any write-only registers mapped to cfg space

joschock:
> @makubacki @joschock will add a mention of the risk to the PR comment. It should also be noted that the In the PCI Base Spec version 6.1 section 7.4 "Configuration Register Types" all configuration space registers are assigned one of the following attributes (quoting Michael Brown in the edk2 discussion thread)
> 
> > HwInit - read-only, no read side effects
> > RO - read-only, no read side effects
> > RW - read-write, no read side effects
> > RW1C - write 1 to clear bits, no read side effects
> > ROS - read-only, no read side effects
> > RWS - read-write, no read side effects
> > RW1CS - write 1 to clear bits, no read side effects
> > RsvdP - read-write, no read side effects
> > RsvdZ - read-write, no read side effects
> 
> From this I don't think we can expect that there are any write-only registers mapped to cfg space

This is only true for registers defined in the PCI spec. It's conceivable that design-specific registers not defined in the PCI spec may have reads with side-effects; since the proposed change affects writes to _all_ addresses (not just to spec-defined registers), this should be noted. 

Edit: to be clear, I know of no such pathological devices; but it would not surprise me if there is some endpoint out there that interprets a read of some device-specific register as a "violation of the device-specific programming model".

JoeLopez333:
> This is only true for registers defined in the PCI spec. It's conceivable that design-specific registers not defined in the PCI spec may have reads with side-effects; since the proposed change affects writes to _all_ addresses (not just to spec-defined registers), this should be noted.

noted and updated the comment in PR description.

makubacki (approved):

kenlautner:
I tested these changes on an X64 system and saw no problems on that end.

cfernald (approved):

makubacki:
Rebased the PR branch on latest release/202302 and set squash auto merge.

PR #610, opened 2023-11-03
[REBASE && FF] Update Stack Cookie Libraries

TaylorBeebe:
StackCheckLib provides the stack cookie globals for both GCC and MSVC
builds. StackCheckFailureLib provides a function to call when a stack
cookie check fails to give platforms a hook to perform actions
before an interrupt is called. After StackCheckFailure() is called,
an interrupt will be generated at the vector specified by
PcdStackCookieExceptionVector.

This series transitions the core to use the new stack cookie libraries.
The stack cookie value no longer needs to be initialized before
image execution and can instead be initialized in the library
constructor. This update also allows the cookie value to be
randomized for GCC buids which was previously not possible because
GCC does not use the cookie value from the image PE/COFF header.

Because a cross-compiler implementation of stack cookies is available,
stack cookies can be enabled for GCC5 builds.

This implementation is also a net reduction in our EDK2 diff footprint.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested on GCC and VS builds of Q35

## Integration Instructions

StackCheckLib will need to be linked to every module built with /GS for MSVC and -fstack-protector for GCC. An instance of StackCheckFailureLib will need to be added to all platform DSC files which utilize StackCheckLib.

```
[LibraryClasses.X64]
  NULL|MdePkg/Library/StackCheckLib/StackCheckLib.inf
  StackCheckFailureLib|MdePkg/Library/StackCheckFailureLibNull/StackCheckFailureLibNull.inf
```


MdePkg/Library/StackCheckLib/StackCheckLib.c
```diff
@@ -0,0 +1,48 @@
...
+#include <Uefi.h>
+#include <Uefi/UefiBaseType.h>
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+#include <Library/PeCoffLib.h>
+#include <Library/RngLib.h>
+#include <Library/StackCheckFailureLib.h>
+
+#if defined (__GNUC__) || defined (__clang__)
+VOID  *__stack_chk_guard = (VOID *)0xBEEBE;
```
kuqin12, line 19: 👍🏽

MdePkg/Library/StackCheckLib/X64/StackCheckFunctionsMsvc.nasm
```diff
@@ -0,0 +1,45 @@
...
+
+extern ASM_PFX(StackCheckFailure)
+extern ASM_PFX(__security_cookie)
+extern ASM_PFX(CpuDeadLoop)
+
+; Called when a buffer check fails. This functionality is dependent on MSVC
+; C runtime libraries and so is unsupported in UEFI.
+global ASM_PFX(__report_rangecheckfailure)
+ASM_PFX(__report_rangecheckfailure):
+    call ASM_PFX(CpuDeadLoop)
```
kuqin12: What is the difference between this and `jmp $`?

MdePkg/Library/StackCheckLib/X64/StackCheckFunctionsMsvc.nasm
```diff
@@ -0,0 +1,45 @@
...
+    ret
+
+; The GS handler is for checking the stack cookie during SEH or
+; EH exceptions and is unsupported in UEFI.
+global ASM_PFX(__GSHandlerCheck)
+ASM_PFX(__GSHandlerCheck):
+    call ASM_PFX(CpuDeadLoop)
```
kuqin12: Same for this? Is it expected to handle it differently with another CPU dead loop call?

MdePkg/Library/StackCheckLibNull/StackCheckLibNull.c
```diff
@@ -0,0 +1,16 @@
...
+
+#include <Uefi.h>
+
+#if defined (__GNUC__) || defined (__clang__)
+VOID  *__stack_chk_guard = (VOID *)0x0;
+#elif defined (_MSC_VER)
+UINTN  __security_cookie = 0x0;
+#else
+  #error "Unsupported compiler"
+#endif
```
kuqin12, line 14: Is the intent of the NULL instance for pipeline build only? Otherwise, should we always err the build when it is used?

MdePkg/Library/StackCheckLib/IA32/StackCheckFunctionsMsvc.nasm
```diff
@@ -0,0 +1,44 @@
+;------------------------------------------------------------------------------
+; IA32/StackCheckFunctionsMsvc.nasm
```
kuqin12, line 2: Copyright?

kuqin12: And a few other places in the PR.

MdePkg/Library/StackCheckFailureLibNull/StackCheckFailureLibNull.inf
```diff
@@ -0,0 +1,24 @@
...
+
+[Sources]
+  StackCheckFailureLibNull.c
+
+[Packages]
+  MdePkg/MdePkg.dec
+
+[LibraryClasses]
+  BaseLib
+  DebugLib
```
kuqin12: New line at the end of the file?

kuqin12 (approved):

MdePkg/Library/StackCheckLib/X64/StackCheckFunctionsMsvc.nasm
TaylorBeebe: I don't think they are functionally different in this case because a return is the next instruction anyways. Do you think jmp is preferable?

MdePkg/Library/StackCheckLibNull/StackCheckLibNull.c
TaylorBeebe: The pipeline will use the non-null version but I agree that generating a build error in the null instance is not the best route. I removed this error.

MdePkg/Library/StackCheckFailureLibNull/StackCheckFailureLibNull.inf
TaylorBeebe: Added a newline

MdePkg/Library/StackCheckLib/IA32/StackCheckFunctionsMsvc.nasm
TaylorBeebe: Added copyrights to the errant files

MdePkg/Library/StackCheckLib/X64/StackCheckFunctionsMsvc.nasm
TaylorBeebe: Went ahead and updated this to a jmp

MdePkg/Library/StackCheckLib/X64/StackCheckFunctionsMsvc.nasm
TaylorBeebe: Updated to a jmp

apop5:
Do the other entry point libraries that need the NO_STACK_COOKIE statement added as well?

MdePkg\Library\PeimEntryPoint\PeimEntryPoint.inf
MdePkg\Library\PeiCoreEntryPoint\PeiCoreEntryPoint.inf
MdePkg\Library\DxeCoreEntryPoint\DxeCoreEntryPoint.inf
MdePkg\Library\UefiDriverEntryPoint\UefiDriverEntryPoint.inf
MdePkg\Library\StandaloneMmDriverEntryPoint\StandaloneMmDriverEntryPoint.inf
MdePkg\Library\UefiApplicationEntryPoint\UefiApplicationEntryPoint.inf

Also, for completeness, do the Include/*EntryPoint.h files also need to have NO_STACK_COOKIE statements added as well?

TaylorBeebe:
> Do the other entry point libraries that need the NO_STACK_COOKIE statement added as well?
> 
> MdePkg\Library\PeimEntryPoint\PeimEntryPoint.inf MdePkg\Library\PeiCoreEntryPoint\PeiCoreEntryPoint.inf MdePkg\Library\DxeCoreEntryPoint\DxeCoreEntryPoint.inf MdePkg\Library\UefiDriverEntryPoint\UefiDriverEntryPoint.inf MdePkg\Library\StandaloneMmDriverEntryPoint\StandaloneMmDriverEntryPoint.inf MdePkg\Library\UefiApplicationEntryPoint\UefiApplicationEntryPoint.inf
> 
> Also, for completeness, do the Include/*EntryPoint.h files also need to have NO_STACK_COOKIE statements added as well?

Yes, they should be there as well.

However, I just looked into it a bit more and it seems like because ProcessLibraryConstructorList() is defined within every module that it's not necessary to put the attributes in the header files at all. We may be able to save a bunch of MU_CHANGEs if that's the case. I'm having a hard time finding a definite ruling on this, though. Do you know off hand?

apop5 (approved):

TaylorBeebe:
FYI: I'm going to delay wrapping up this PR until the week of 11/20 because every other repo will also need an update to use this new stack cookie library.

PR #613, opened 2023-11-07
[REBASE && FF] Update GetMemoryMapWithPopulatedAccessAttributes() to Defer Free/Allocate Calls, Improve Documentation

TaylorBeebe:
## Description

During the initialization of the memory protection, a complete map
of system memory is created via a call toGetMemoryMapWithPopulatedAccessAttributes().
The first step of the function was to fetch the EFI memory map and subsequent
steps manipulated it to create the output. The subsequent steps included
freeing and allocating memory which potentially changed the memory
layout and caused the output memory map to be incorrect. Because
allocating and freeing memory only transitions memory type between
EfiConventionalMemory and EfiBootServicesMemory (which are required
to have the same memory protection policy), this nuance didn't matter.
However, as the DXE core is updated to support more strict protections
on free memory, this nuance will matter.

This patch updates the GetMemoryMapWithPopulatedAccessAttributes()
routine to calculate and allocate all required memory before
fetching the EFI memory map used in the output, and adds a function
to free that memory once the caller is done with the ouput.

Also, to improve the long-term maintainability of the memory protection
support code, this change adds code documentation in the form of inline
comments to the memory protection initialization logic.

**Test Case 1:**
image_protection_off
The only difference between before and after this change set is the absence of an EfiConventionalMemory section. The post-changes on the right prints out the location of the allocated memory map and bitmap which is where the EfiConventionalMemory section used to be. In reality, both memory maps have essentially the same layout, but the memory map post-changes accurately reflects the actual memory map because no allocations or frees are done during its creation.

**Test Case 2:**
image_protection_onpng
Same explanation as above with image protection on.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -2573,25 +2573,15 @@ GetMemoryMapWithPopulatedAccessAttributes (
     return EFI_INVALID_PARAMETER;
   }
 
-  *MemoryMapSize = 0;
-
-  Status = CoreGetMemoryMap (
-             MemoryMapSize,
-             *MemoryMap,
-             &MapKey,
-             DescriptorSize,
-             &DescriptorVersion
-             );
-
-  ASSERT (Status == EFI_BUFFER_TOO_SMALL);
-
-  do {
-    *MemoryMap = (EFI_MEMORY_DESCRIPTOR *)AllocatePool (*MemoryMapSize);
-    if (*MemoryMap == NULL) {
-      Status = EFI_OUT_OF_RESOURCES;
-      ASSERT_EFI_ERROR (Status);
-      goto Cleanup;
-    }
+  // STEP 1: Get the Initial Memory Map
+  //
+  // The memory map fetched from this routine will be used to determine
+  // the amount of memory needed to create a hybrid memory map which
+  // describes the full address space with all the attributes required
+  // to secure memory as much as possible based on the platform protection
+  // policy.
+  {
+    *MemoryMapSize = 0;
 
     Status = CoreGetMemoryMap (
                MemoryMapSize,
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1232,184 +1234,193 @@ GetOverlappingMemorySpaceRegion (
...
-
-  @param[in, out] MemoryMapSize         Size, in bytes, of MemoryMap
-  @param[in, out] MemoryMap             IN:  Pointer to the EFI memory map which will have all gaps filled. The
-                                             original buffer will be freed and updated to a newly allocated buffer
-                                        OUT: A sorted memory map describing the entire memory region
-  @param[in]      DescriptorSize        Size, in bytes, of each descriptor region in the array
-  @param[in]      StartOfAddressSpace   Starting address from which there should be contiguous entries
-  @param[in]      EndOfAddressSpace     Ending address at which the memory map should at least reach
+  max(EndOfAddressSpace, address + length of the final memory map entry). If DetermineSize
+  is TRUE, then this function will just determing the required buffer size for the output
```
kenlautner:
```suggestion
  is TRUE, then this function will just determine the required buffer size for the output

```

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2660,124 +2820,143 @@ GetMemoryMapWithPopulatedAccessAttributes (
...
-  // Break up the memory map so every image range divides evenly into some number of
-  // EFI memory descriptors
-  SeparateImagesInMemoryMap (
-    MemoryMapSize,
-    *MemoryMap,
-    *DescriptorSize,
-    MergedImageList,
-    AdditionalRecordCount
-    );
+  // STEP 14: Update the Bitmap Variables to Isolate Used Descriptors
```
kenlautner, line 2872: Multiple step labels here.

TaylorBeebe: Unless I'm totally missing what you're referring to, I think this step is just referencing step 8

kenlautner (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1232,184 +1234,194 @@ GetOverlappingMemorySpaceRegion (
...
 
-  if ((MemoryMap == NULL) || (*MemoryMap == NULL) ||
+  if ((MemoryMap == NULL) || (MemoryMap == NULL) ||
       (MemoryMapSize == NULL) || (*MemoryMapSize == 0) ||
       (*DescriptorSize == 0) || (MemorySpaceMap == NULL) ||
-      (MemorySpaceMapDescriptorSize == NULL) || (MemorySpaceMapDescriptorCount == NULL))
+      (MemorySpaceMapDescriptorSize == NULL) || (MemorySpaceMapDescriptorCount == NULL) ||
+      (!DetermineSize && (InsertionPoint == NULL)))
   {
+    DEBUG ((DEBUG_ERROR, "%a - Function had NULL input(s)!\n", __FUNCTION__));
```
makubacki: For new additions, we should go ahead and use `__func__`.

```suggestion
    DEBUG ((DEBUG_ERROR, "%a - Function had NULL input(s)!\n", __func__));
```

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1232,184 +1234,194 @@ GetOverlappingMemorySpaceRegion (
...
-  @param[in, out] MemoryMap             IN:  Pointer to the EFI memory map which will have all gaps filled. The
-                                             original buffer will be freed and updated to a newly allocated buffer
-                                        OUT: A sorted memory map describing the entire memory region
-  @param[in]      DescriptorSize        Size, in bytes, of each descriptor region in the array
-  @param[in]      StartOfAddressSpace   Starting address from which there should be contiguous entries
-  @param[in]      EndOfAddressSpace     Ending address at which the memory map should at least reach
+  max(EndOfAddressSpace, address + length of the final memory map entry). If DetermineSize
+  is TRUE, then this function will just determine the required buffer size for the output
+
+  memory map.
```
makubacki: Clean up empty line.

```suggestion
  is TRUE, then this function will just determine the required buffer size for the output
  memory map.
```

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1232,184 +1234,194 @@ GetOverlappingMemorySpaceRegion (
...
-  EFI_MEMORY_DESCRIPTOR  *OldMemoryMapCurrent, *OldMemoryMapEnd, *NewMemoryMapStart, *NewMemoryMapCurrent;
+  EFI_MEMORY_DESCRIPTOR  *MemoryMapCurrent, *MemoryMapEnd;
   EFI_PHYSICAL_ADDRESS   LastEntryEnd, NextEntryStart, StartOfAddressSpace, EndOfAddressSpace;
   EFI_GCD_MEMORY_TYPE    GcdType = 0;
   UINT64                 RemainingLength, OverlapLength;
-  UINTN                  NewMemoryMapSize, AdditionalEntriesCount, LoopIteration;
+  UINTN                  AdditionalEntriesCount;
 
-  if ((MemoryMap == NULL) || (*MemoryMap == NULL) ||
+  if ((MemoryMap == NULL) || (MemoryMap == NULL) ||
```
makubacki: `MemoryMap` is now checked twice against `NULL`.


```suggestion
  if ((MemoryMap == NULL) ||
```

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -1232,184 +1234,194 @@ GetOverlappingMemorySpaceRegion (
...
       (MemoryMapSize == NULL) || (*MemoryMapSize == 0) ||
       (*DescriptorSize == 0) || (MemorySpaceMap == NULL) ||
-      (MemorySpaceMapDescriptorSize == NULL) || (MemorySpaceMapDescriptorCount == NULL))
+      (MemorySpaceMapDescriptorSize == NULL) || (MemorySpaceMapDescriptorCount == NULL) ||
+      (!DetermineSize && (InsertionPoint == NULL)))
   {
+    DEBUG ((DEBUG_ERROR, "%a - Function had NULL input(s)!\n", __FUNCTION__));
     return EFI_INVALID_PARAMETER;
```
makubacki: Null has a special meaning that is often different than "zero-sized".

I suggest breaking apart null check and size checks into two different conditional blocks with corresponding messages.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
makubacki, line 1: I think this change would benefit from testing information in the PR description.

Can you describe what tests have been run? I know the PR description will not carry forward to the commits, but it's useful to roll in to the release notes (which picks up its info from the PR description).

makubacki (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Updated the description with comparisons between before/after the changes.

PR #619, opened 2023-11-13
Define GUID for boot manage policy to connect storage devices.

cfernald:
## Description

Some devices may require connecting storage media while booting to network or require enumerating storage protocols that were not initially enumerated during DXE. This change adds a GUID to allow implementation of boot manager's ConnectDeviceClass to connect storage media.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


MdePkg/Include/Protocol/BootManagerPolicy.h
```diff
@@ -31,6 +31,13 @@
     0x113B2126, 0xFC8A, 0x11E3, { 0xBD, 0x6C, 0xB8, 0xE8, 0x56, 0x2C, 0xBA, 0xFA } \
   }
 
+// MU_CHANGE START: Introduce disk connect guid
+#define EFI_BOOT_MANAGER_POLICY_STORAGE_GUID \
```
kuqin12, line 35: Probably also need `gEfiBootManagerPolicyConnectStorageGuid`

kuqin12: And also similar change in the dec file?

kuqin12 (approved):

MdePkg/Include/Protocol/BootManagerPolicy.h
cfernald: good point

apop5 (approved):

MdePkg/Include/Protocol/BootManagerPolicy.h
makubacki, line 35: Is this planned to go to the UEFI spec?

cfernald: Yes, we are planning to take this to the UEFI spec and ideally push an open source implementation to EDK2's boot manager policy module

makubacki (approved):

MdePkg/Include/Protocol/BootManagerPolicy.h
makubacki: Thanks. In that case, `EFI_` makes sense as a code first change and we'd get parity again with the upstream module.

PR #665, opened 2024-01-04
Update Memory Protection Special Region Logic to Handle Edge Case

TaylorBeebe:
## Description

The special region functionality enables platforms to describe what the memory attributes should be for any number of memory regions when the memory protection initialization routine is complete. The logic which ensures that the special regions are accounted for has a case which is unhandled and described below:

Thef following is a special region which says that the address range should have EFI_MEMORY_RP:
0x4000-0x8000, EFI_MEMORY_RP

When generating the memory map used as a blueprint for applying attributes to the platform, there can be some gaps. Take the following example:
0x0000-0x6000,
0x7000-0x8000

The logic before would properly apply the attributes to the region 0x4000-0x6000 but would get confused when there is a gap between the last map entry and next map entry when the special region covers both. This change fixes this issue by checking if the special region start is within the map entry interval OR the map entry start is within the special region interval (a proper overlap check).

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 and a Surface ARM platform by publishing a special region which encounters the edge case.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -2171,9 +2171,9 @@ SeparateSpecialRegionsInMemoryMap (
     {
       MapEntryStart = (UINTN)MemoryMapEntry->PhysicalStart;
       MapEntryEnd   = (UINTN)MemoryMapEntry->PhysicalStart + (UINTN)EFI_PAGES_TO_SIZE (MemoryMapEntry->NumberOfPages);
-      if ((MapEntryStart <= SpecialRegionStart) && (MapEntryEnd > SpecialRegionStart)) {
-        // Check if some portion of the map entry isn't covered by the special region
-        if (MapEntryStart != SpecialRegionStart) {
+      if (CHECK_OVERLAP (SpecialRegionStart, SpecialRegionEnd, MapEntryStart, MapEntryEnd)) {
+        // Check if some portion before the map entry isn't covered by the special region
+        if (MapEntryStart < SpecialRegionStart) {
           // Populate a new descriptor for the region before the special region. This entry can go to the end
           // of the memory map because the special region list is sorted
           POPULATE_MEMORY_DESCRIPTOR_ENTRY (


os-d (approved):
I walked through the code a bit, I think this is covered, but just to confirm, if the special region were to cover a range that terminated in a gap in the memory map, this would still work? I.e. if 0x0000-0x2000 and 0x6000 - 0x9000 were the memory map descriptors and the special region was 0x1000 - 0x4000, this would not cause an issue?

TaylorBeebe:
> I walked through the code a bit, I think this is covered, but just to confirm, if the special region were to cover a range that terminated in a gap in the memory map, this would still work? I.e. if 0x0000-0x2000 and 0x6000 - 0x9000 were the memory map descriptors and the special region was 0x1000 - 0x4000, this would not cause an issue?

It will work in that it will only update what's in the memory map. So, it would update intervals 0x1000-0x2000 but would essentially ignore the 0x2000-0x4000 interval. If a memory range is not in the map at this point, it's because it is nonexistent in the GCD memory map (which usually means that it is unmapped in the page table). Skipping nonexistent regions was done to accommodate some of our platforms which hit errors when applying attributes to upper address regions. The correct thing to do MAY be to always update the special region intervals regardless of if the region is mapped at this point. I can also see an argument in the other direction.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
makubacki, line 1: This is the type of condition that would be nice to show breaking/fixed unit tests before/after the change. Is there anything like that you could add? Or is it a significant lift?

makubacki (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: I think it would need to be a host-based unit test because this functionality is embedded in the core. A DxeCore host-based unit test currently exists in Project Mu but I was planning on removing it because it's no longer needed now that the ImagePropertiesRecordLib stuff is in EDK2.

I'd prefer to just remove the DxeCore host-based test because it's a bit hacky and requires maintenance as the DxeCore INF changes, but @os-d mentioned that he might want to use it so if he would still like to do so then I can add a unit test for this.

os-d: I do think DxeCore unit testing is critical. My immediate use case is a local test (which could be checked in, I suppose). However, if the existing test is not the right vehicle for generic DxeCore unit testing, then I'm okay with dropping it in favor of something less hacky (if we can work that out).

TaylorBeebe: I think the existing unit test is probably the right vehicle (with a rename), it's just that host-based testing DxeCore is finicky. We can add an override tag to it to ensure we revisit it whenever the DxeCore INF changes.

Writing a unit test for this logic will take a bit and this fix is critical for a product. I'm going to push this PR through and create a task for what's been discussed here.

TaylorBeebe: Issue for generic DxeCore host-based unit test: https://github.com/microsoft/mu_basecore/issues/666
Issue for creating a unit-test for this logic: https://github.com/microsoft/mu_basecore/issues/667

PR #669, opened 2024-01-08
Don't Assert if InitializeSeparateExceptionStacks() Fails

TaylorBeebe:
## Description

ARM platforms don't initialize separate exception stacks. The ARM CpuExceptionHandler libraries always return EFI_SUCCESS in InitializeSeparateExceptionStacks() to sidestep this exception. To accommodate the use of the NULL library instance which returns EFI_UNSUPPORTED, just print an error when the routine fails.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a Surface ARM platform

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
+++ b/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
@@ -261,7 +261,7 @@ DxeMain (
   //
   // Setup Stack Guard
   //
-  // MU_CHANGE START: Check Memory Protection HOB, don't ASSERT if the exception stacks can't be created
+  // MU_CHANGE START: Check Memory Protection HOB, don't ASSERT if the exception stack init returns unsupported
   // if (PcdGetBool (PcdCpuStackGuard)) {
   GuidHob2 = GetFirstGuidHob (&gDxeMemoryProtectionSettingsGuid);
   if ((GuidHob2 != NULL) &&
--- a/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
+++ b/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
@@ -270,7 +270,9 @@ DxeMain (
   {
     Status = InitializeSeparateExceptionStacks (NULL, NULL);
     // ASSERT_EFI_ERROR (Status);
-    DEBUG ((DEBUG_ERROR, "%a: Failed to create exception stacks!\n", __FUNCTION__));
+    if (EFI_ERROR (Status)) {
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
```diff
@@ -261,15 +261,16 @@ DxeMain (
...
+  // MU_CHANGE START: Check Memory Protection HOB, don't ASSERT if the exception stacks can't be created
   // if (PcdGetBool (PcdCpuStackGuard)) {
   GuidHob2 = GetFirstGuidHob (&gDxeMemoryProtectionSettingsGuid);
   if ((GuidHob2 != NULL) &&
       (((DXE_MEMORY_PROTECTION_SETTINGS *)GET_GUID_HOB_DATA (GuidHob2))->StructVersion == (UINT8)DXE_MEMORY_PROTECTION_SETTINGS_CURRENT_VERSION) &&
       ((DXE_MEMORY_PROTECTION_SETTINGS *)GET_GUID_HOB_DATA (GuidHob2))->CpuStackGuard)
   {
     Status = InitializeSeparateExceptionStacks (NULL, NULL);
-    ASSERT_EFI_ERROR (Status);
+    // ASSERT_EFI_ERROR (Status);
```
os-d: This would be a major error on an x86 platform, right? Is it better to change the NULL lib to return EFI_SUCCESS, instead of removing this assert?

TaylorBeebe: It's a major issue if stack overflows or stack corruption issues are occurring. The InitializeSeparateExceptionStacks() routine is also used in the CPU Multiprocessor PEI and DXE instances where it definitely needs to know when the exception stack wasn't successfully created.

makubacki: If `EFI_UNSUPPORTED` is an expected failure. Can it still assert on other (unexpected) failures?

```suggestion
    ASSERT (Status == EFI_UNSUPPORTED || !EFI_ERROR (Status));
```

makubacki (approved):

MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c
TaylorBeebe: Yes, I'll leave the print and use your version of the assert

os-d (approved):

PR #675, opened 2024-01-15
Fix flatten PDBs

### LINKED ISSUE #673: [Bug]: FlattenPdbs not copying any files

### Current Behavior

In #511, FlattenPdbs was modified to only include PDBs from the OUTPUT folder. However my build for whatever reason puts the PDB in the folder outside of this (plus the DEBUG folder), but not the OUTPUT folder, so no PDBs are getting copied to the PDB flat directory.

ie, my directory structure looks like this:
```
 DxeMain
 ├╴DEBUG
 │ ├╴AutoGen.c
 │ ├╴AutoGen.h
 │ ├╴DxeCore.dll
 │ ├╴DxeCore.efi
 │ ╰╴DxeCore.map
 ├╴OUTPUT
 │ ├╴AutoGen.c.deps
 │ ├╴AutoGen.obj
 │ ├╴DxeCore.depex
 │ ├╴DxeCore.efi
 │ ├╴DxeCore.inf
 │ ├╴DxeCore.lib
 │ ├╴DxeCore.map
 │ ├╴DxeCore.txt
 │ ├╴DxeCore.uni
 │ ├╴DxeCoreExtra.uni
 │ ├╴object_files.lst
 │ ╰╴static_library_files.lst
 ├╴AutoGenTimeStamp
 ├╴dependency
 ├╴deps.txt
 ├╴deps_target
 ├╴DxeCore.makefile
 ├╴DxeCore.pdb   <-- This should be copied out
 ╰╴Makefile
 ```
 
 Suggested fix:
 ```python
        logging.critical("Copying PDBs to flat directory")

        symbols = set()
        for file in Path(build_path).rglob("*.pdb"):
            pdb_out = Path(pdb_path / file.name)

            # Skip MSVC symbols
            if "vc1" in file.name.lower():
                continue

            # Deduplicate
            if not file.name in symbols:
                symbols.add(file.name)

                # If it exists and has the same file identifier, skip it.
                if pdb_out.exists() and file.stat().st_ino == pdb_out.stat().st_ino:
                    continue

                # Hard link it, which is slightly faster, but mainly allows us to tell
                # if the file has changed (st_ino is different)
                pdb_out.unlink(missing_ok=True)
                pdb_out.hardlink_to(file)

        if not symbols:
            logging.error('No symbols were copied')
            return -1

        return 0
 ```

### Expected Behavior

PDBs should be copied under $BuildOutput\PDB.

### Steps To Reproduce

Build platform, observe no symbols copied to PDB folder.

### Build Environment

```markdown
- OS(s): Windows 11
- Tool Chain(s): VS2022
- Targets Impacted: All
```

### Version Information

```text
5658e8a1f82238bfa2059f82f58699f7f3dcb8f3
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

jorticus-msft:
## Description

Fixes #673 - PDBs were not being copied out into the flat PDB directory in the build output, as something else in the build system moved the location of the PDB output file, and the original code was not finding it there.

This change expands the search scope, whilst still preserving the de-duplication logic and performance improvement that was there before.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran through our build system, observed that PDB files are emitted.
Plugin also tests whether symbols are generated, and fails the build if they were not.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Plugin/FlattenPdbs/FlattenPdbs.py
+++ b/BaseTools/Plugin/FlattenPdbs/FlattenPdbs.py
@@ -48,18 +48,30 @@ class FlattenPdbs(IUefiBuildPlugin):
             logging.critical("Error making PDB directory")
 
         logging.critical("Copying PDBs to flat directory")
+
+        symbols = set()
         for file in Path(build_path).rglob("*.pdb"):
-            # pdb exists in DEBUG and OUTPUT directory. Same file.
             pdb_out = Path(pdb_path / file.name)
-            if file.parent.name != "OUTPUT":
-                continue
-            # If it exists and has the same file identifier, skip it.
-            if pdb_out.exists() and file.stat().st_ino == pdb_out.stat().st_ino:
-                continue
+
+            # Skip MSVC symbols
             if "vc1" in file.name.lower():
                 continue
-            # Hard link it, which is slightly faster, but mainly allows us to tell
-            # if the file has changed (st_ino is different)
-            pdb_out.unlink(missing_ok=True)
-            pdb_out.hardlink_to(file)
+
+            # Deduplicate
+            if not file.name in symbols:
+                symbols.add(file.name)
+
+                # If it exists and has the same file identifier, skip it.
+                if pdb_out.exists() and file.stat().st_ino == pdb_out.stat().st_ino:
+                    continue
+
+                # Hard link it, which is slightly faster, but mainly allows us to tell
+                # if the file has changed (st_ino is different)
+                pdb_out.unlink(missing_ok=True)
+                pdb_out.hardlink_to(file)
+
+        if not symbols:
+            logging.error('No symbols were copied')
+            return -1
+
         return 0


jorticus-msft:
It looks like the Linux build is failing because it doesn't actually generate PDBs:

> CRITICAL - Copying PDBs to flat directory
ERROR - No symbols were copied
ERROR - Plugin Failed: Flatten Pdbs Uefi Build Plugin returned -1

Can someone confirm if this is expected?
If so, I'll modify this check to only run for Windows/MSVC.

makubacki:
> It looks like the Linux build is failing because it doesn't actually generate PDBs:
> 
> > CRITICAL - Copying PDBs to flat directory
> > ERROR - No symbols were copied
> > ERROR - Plugin Failed: Flatten Pdbs Uefi Build Plugin returned -1
> 
> Can someone confirm if this is expected? If so, I'll modify this check to only run for Windows/MSVC.

Correct. GCC does not generate PDB files.

Javagedes:
@jorticus-msft  I see this plugin as an automation tool to move the map files from the default generated location (DEBUG - see screenshot for build rule) to the flattened directory. This is to prevent the need of overriding the `/PDB` and `/PDBALTPATH` DLINK flags.

image

Since the platform is already overriding the defaults by changing the name and location of the PDB with `MSFT:*_*_*_DLINK_FLAGS = /PDB: <...>.pdb /PDBALTPATH: <...>.pdb`, It seems more appropriate for the platform to update the path that the PDB is placed, rather than to update this plugin.

I've noted that the primary reason `/PDB` and `/PDBALTPATH` is updated, is to change the name of the final PDB. Reviewing the DSC, it seems more like a bug in how the platform is setting the `/PDB` and `/PDBALTPATH` because the values being set is the pdbname only, but it should be folder_path + pdbname.

I see this as a platform bug, not a plugin bug, with the following two solutions:

1. Update your current BuildOption rule override for DLINK_FLAGS to place it directly in PDB (Must create the PDB folder manually in the PlatformBuild.py):
`MSFT:*_*_*_DLINK_FLAGS = /PDB:$(BUILD_DIR)/PDB/<name>.pdb /PDBALTPATH:$(BUILD_DIR)/PDB/<name>.pdb`

2. Update your current BuildOption rule override for DLINK_FLAGS to place it back into the DEBUG dir:
`MSFT:*_*_*_DLINK_FLAGS = /PDB:$(DEBUG_DIR)/<name>.pdb /PDBALTPATH:$(DEBUG_DIR)/<name>.pdb`

3. A third option, if there is a consensus that FlattenPdbs should changed, is that I can glob the entire output dir moving any *pdb files into the PDB folder.

jorticus-msft:
It looks like the change to the /PDB path on our side was intentional in order to remove path information from the PDB, which is beneficial for a number of reasons (including a reduction in flash space and for security).

Therefore I won't be modifying this flag for our platforms and would highly recommend that this plugin be updated to be more flexible. My proposed solution implements suggestion 3 already.

Javagedes:
You can 

> It looks like the change to the /PDB path on our side was intentional in order to remove path information from the PDB, which is beneficial for a number of reasons (including a reduction in flash space and for security).
> 
> Therefore I won't be modifying this flag for our platforms and would highly recommend that this plugin be updated to be more flexible. My proposed solution implements suggestion 3 already.

I know. And I gave you the solution that keeps the renamed PDBs but places them in the correct location. This is a platform bug.

Your DSC has this: 
`MSFT:*_*_*_DLINK_FLAGS    = /PDB:$(MODULE_NAME_GUID).pdb /PDBALTPATH:$(MODULE_NAME_GUID).pdb`

As I stated. This is a platform bug. `PDB` and `PDBALTPATH` should be an absolute path whereas it is currently just provided as a filename.

My suggested solution is to change it to this:
`MSFT:*_*_*_DLINK_FLAGS = /PDB:$(DEBUG_DIR)/<name>.pdb /PDBALTPATH:$(DEBUG_DIR)/<name>.pdb`

i.e.
`MSFT:*_*_*_DLINK_FLAGS = /PDB:$(DEBUG_DIR)/$(MODULE_NAME_GUID).pdb /PDBALTPATH:$(DEBUG_DIR)/$(MODULE_NAME_GUID).pdb`

This keeps the renamed file, but places it in the correct filepath.

Javagedes:
> It looks like the change to the /PDB path on our side was intentional in order to remove path information from the PDB, which is beneficial for a number of reasons (including a reduction in flash space and for security).
> 
> Therefore I won't be modifying this flag for our platforms and would highly recommend that this plugin be updated to be more flexible. My proposed solution implements suggestion 3 already.

@jorticus-msft 

I now understand what you mean in regard to stripping path information from the PDB. Thanks for that extra bit of information.

However, to accomplish this, you only need to override [/PDBALTPATH](https://learn.microsoft.com/en-us/cpp/build/reference/pdbaltpath-use-alternate-pdb-path?view=msvc-170): ` /PDBALTPATH does not change the location or name of the actual .pdb file; it changes the information that the linker writes in the binary file. `

The only thing that [/PDB](https://learn.microsoft.com/en-us/cpp/build/reference/pdb-use-program-database?view=msvc-170) does is change the location of the file. `Use /PDB:filename to specify the name of the PDB file.`

Therefore the solution for your situation would be to change your BuildOption Override

`MSFT:*_*_*_DLINK_FLAGS    = /PDBALTPATH:$(MODULE_NAME_GUID).pdb`

I'm still happy to make the plugin more flexible if that is the route decided, but I still lean towards ensuring platforms are correct, over adapting core logic to support misuse.

jorticus-msft:
I see, yes it looks like using separate /PDB and /PDBALTPATH definitions will work.

It would still be useful if the flatten PDB plugin caused a build failure if it fails to find any PDBs when PDBs are expected... this issue went unnoticed for quite some time as the tool wasn't copying anything out.

Javagedes:
> I see, yes it looks like using separate /PDB and /PDBALTPATH definitions will work.
> 
> It would still be useful if the flatten PDB plugin caused a build failure if it fails to find any PDBs when PDBs are expected... this issue went unnoticed for quite some time as the tool wasn't copying anything out.

In my opinion, it is outside of the scope of the plugin to know when (and when not) PDBs are expected to exist, particularly because the tool chain is generic to support all platforms. Many things can influence whether or not PDBs will be generated including the toolchain (as Michael mentioned, GCC does not produce pdbs, I'm not sure about others), the build target (By default, RELEASE builds do not generate pdbs), and pdbs can be added or removed by any DLINK_FLAGS override, whether at the INF or DSC level.

I think a better approach would be to add a step to your platform's CI pipeline that ensures PDBs exist in the PDB folder when your particular platform expects it to exist.

Let me know your thoughts.

Thanks,
Joey

jorticus-msft:
That sounds reasonable. I'll add a separate plugin on our side for the validation step and work on updating the linker flags to put PDBs in the correct location.

PR #714, opened 2024-02-01
[REBASE && FF] Add Stack Cookie Support for IA32, ARM, and AARCH64

TaylorBeebe:
This update changes the following:

### **Basetools update**
CI builds would use the basetools installed via pip if available. Because we have Project Mu edits to basetools and the project has little support in EDK2, this update causes all builds to use our in-tree basetools.

### **Fix NULL lib parsing**
When parsing INF files, Basetools treats both libraries and modules the same. When the library dependencies are being collected for a module/library, libraries linked via:

`NULL|Path/To/Library`

would be included in the list of dependencies for libraries which does not match how these expressions are expected to be interpreted. This update changes the evaluation loop to skip NULL links when collecting dependencies for libraries.

### **Add GCC ARM, AARCH64, IA32 support and MSVC IA32 support for StackCheckLibNull**
MSVC IA32 requires the __security_check_cookie function to specify byte size (@__security_check_cookie@4). This change also declares __stack_chk_fail() in StackCheckLibNull.c to support GCC.

### **Implement Stack Cookie Support for MSVC IA32 and GCC IA32, ARM, and AARCH64**
This update replaces StackCheckLib with StackCheckLibStaticInit and StackCheckLibDynamicInit. The new libraries have GCC support for ARM, AARCH64, IA32 and X64 builds. The libraries have MSVC support for IA32 and X64 builds.

StackCheckLibStaticInit does not have a library constructor and should be used whenever the stack cookie value cannot be updated during driver execution (i.e. when the stack cookie is not in a writable or no RNG library is available). The value of the stack cookie is generated at buildtime via a Basetools update.

StackCheckLibDynamicInit has a library constructor and should be used whenever the stack cookie value can be updated at runtime (i.e. for DXE modules and shadowed PEIMs).

### **Update StackCheckFailureLib to StackCheckFailureHookLib**
To clarify the purpose of StackCheckFailureLib, this PR renames it to StackCheckFailureHookLib. Also, the failure address is passed as an argument to the hook function to allow the hook to trace the fault. An interrupt will still be called after the hook returns.

### **How This Was Tested**
Tested on a Q35 GCC and MSVC builds by purposefully performing a stack overflow and verifying the stack check failure hook was called followed by the stack check failure interrupt.

### **Integration Instructions**
Platforms will need to explicitly declare the StackCheckLib and StackCheckFailureLib instances for their platforms.

EXAMPLE:

```
StackCheckFailureHookLib|MdePkg/Library/StackCheckFailureHookLibNull/StackCheckFailureHookLibNull.inf
[LibraryClasses.common.SEC, LibraryClasses.common.PEI_CORE]
  NULL|MdePkg/Library/StackCheckLibNull/StackCheckLibNull.inf

[LibraryClasses.common.PEIM, LibraryClasses.common.MM_CORE_STANDALONE, LibraryClasses.common.MM_STANDALONE]
  NULL|MdePkg/Library/StackCheckLib/StackCheckLibStaticInit.inf

[LibraryClasses.common.DXE_CORE, LibraryClasses.common.SMM_CORE, LibraryClasses.common.DXE_SMM_DRIVER, LibraryClasses.common.DXE_DRIVER, LibraryClasses.common.DXE_RUNTIME_DRIVER, LibraryClasses.common.DXE_SAL_DRIVER, LibraryClasses.common.UEFI_DRIVER, LibraryClasses.common.UEFI_APPLICATION]
  NULL|MdePkg/Library/StackCheckLib/StackCheckLibDynamicInit.inf
```


apop5:
nit: The StackCheckLibRuntimeInit, the "Runtime" in the name creates confusion with Runtime drivers (DXE_RUNTIME_DRIVER).

Maybe "DynamicInit" would be a better name?

MdePkg/MdePkg.dec
```diff
@@ -104,7 +104,7 @@
 
 ## MU_CHANGE [BEGIN]
   ##  @libraryclass  Provides a hook called when a stack cookie check fails.
-  StackCheckFailureLib|Include/Library/StackCheckFailureLib.h
+  StackCheckFailureHookLibNull|Include/Library/StackCheckFailureHook.h
```
Javagedes: This should be the library class name, not the library class instance name - i.e.  `StackCheckFailureHookLib`

TaylorBeebe:
> nit: The StackCheckLibRuntimeInit, the "Runtime" in the name creates confusion with Runtime drivers (DXE_RUNTIME_DRIVER).
> 
> Maybe "DynamicInit" would be a better name?

Yes, I can rename to DynamicInit and StaticInit.

TaylorBeebe:
This PR is on hold while some CI issues are resolved.

1. The build-time stack cookie logic requires edits to base tools. In CI, we use the EDK2 basetools from Tianocore. The EDK2 basetools can either be dropped from pip-requirements, or the setting in CISettings should be updated to specify that the local basetools should be used.

2. Sometimes, more than one stack cookie library can be included for a module. This is inconsistent and only seems to happen in CI which indicates there may be a bug in basetools. This is being investigated.

TaylorBeebe:
> This PR is on hold while some CI issues are resolved.
> 
> 1. The build-time stack cookie logic requires edits to base tools. In CI, we use the EDK2 basetools from Tianocore. The EDK2 basetools can either be dropped from pip-requirements, or the setting in CISettings should be updated to specify that the local basetools should be used.
> 2. Sometimes, more than one stack cookie library can be included for a module. This is inconsistent and only seems to happen in CI which indicates there may be a bug in basetools. This is being investigated.

These two issues should be fixed with the second and third commits of this PR

BaseTools/Source/Python/AutoGen/GenC.py
```diff
@@ -2042,6 +2047,23 @@ def CreateFooterCode(Info, AutoGenC, AutoGenH):
 def CreateCode(Info, AutoGenC, AutoGenH, StringH, UniGenCFlag, UniGenBinBuffer, StringIdf, IdfGenCFlag, IdfGenBinBuffer):
     CreateHeaderCode(Info, AutoGenC, AutoGenH)
 
+    # MU_CHANGE [START]: Add Stack Cookie Support
+    if Info.ModuleType != SUP_MODULE_HOST_APPLICATION:
```
Javagedes, line 2051: I don't think there is any harm in applying the STACK_COOKIE_VALUE define to all header files instead of this logic to only apply it under special circumstances. It reduces the complexity and removes the need of a special conditional on a specific library. @makubacki do you have any differing opinions?

BaseTools/Source/Python/Workspace/WorkspaceCommon.py
```diff
@@ -122,7 +122,12 @@ def GetModuleLibInstances(Module, Platform, BuildDatabase, Arch, Target, Toolcha
 
     while len(LibraryConsumerList) > 0:
         M = LibraryConsumerList.pop()
+        # MU_CHANGE [START]: Fix NULL LibraryClass Inclusion Issue
+        ModuleIsLibrary = True if len(M.LibraryClass) > 0 else False
```
Javagedes: You should be able to just use `bool(M.LibraryClass)` following the example [here](https://github.com/microsoft/mu_basecore/blob/d5379bdd63dc290d209b43746e0ab83108661ce4/BaseTools/Source/Python/AutoGen/ModuleAutoGen.py#L405) since `bool("") == False`

BaseTools/Source/Python/AutoGen/GenC.py
makubacki: I agree. I think the upside - consistency for debug, scaling to new archs, etc. is worth it. This is already done for some things like `EDKII_DSC_PLATFORM_GUID` which is rarely used.

BaseTools/Source/Python/Workspace/WorkspaceCommon.py
TaylorBeebe: Updated to bool(M.LibraryClass). I'll also update the EDK2 patch

BaseTools/Source/Python/AutoGen/GenC.py
TaylorBeebe: Updated to define STACK_COOKIE_VALUE for all modules and libraries which are not host-based unit tests

apop5 (approved):

Javagedes (approved):

MdePkg/Library/StackCheckFailureHookLibNull/StackCheckFailureHookLibNull.inf
```diff
@@ -1,24 +1,20 @@
...
 #
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  SPDX-License-Identifier: BSD-2-Clause-Patent
 ##
 
 [Defines]
   INF_VERSION                    = 0x00010005
-  BASE_NAME                      = StackCheckFailureLibNull
+  BASE_NAME                      = StackCheckFailureHookLibNull
   FILE_GUID                      = 0a4f7cde-a567-485d-8c9a-98ea271ba322
```
os-d: nit: I tend to change the file guid when renaming a file, to avoid duplicate in existing build dirs, but probably doesn't matter too  much

os-d (approved):

MdePkg/Library/StackCheckFailureHookLibNull/StackCheckFailureHookLibNull.inf
TaylorBeebe: Good call -- updated the guid

.pytool/CISettings.py
makubacki, line 1: It looks like the first two commits are essentially cherry picked from https://github.com/microsoft/mu_basecore/commit/e3a139fa066c52271ac490a19b9d5bad6f9c7a18 in release/202311.

Can you make that a litte more obvious for tracking? With [CHERRY-PICK] tags and/or cherry picked from commit xyz messages in the commits?

BaseTools/Source/Python/Workspace/WorkspaceCommon.py
```diff
@@ -123,6 +123,10 @@ def GetModuleLibInstances(Module, Platform, BuildDatabase, Arch, Target, Toolcha
     while len(LibraryConsumerList) > 0:
         M = LibraryConsumerList.pop()
         for LibraryClassName in M.LibraryClasses:
+        # MU_CHANGE [START]: Fix NULL LibraryClass Inclusion Issue
+            if LibraryClassName.startswith("NULL") and bool(M.LibraryClass):
```
makubacki, line 127: Although unlikely, should this check for an exact NULL match? For example, it shouldn't match a name like `NULLGpioLib`, right?

makubacki (approved):

BaseTools/Source/Python/Workspace/WorkspaceCommon.py
TaylorBeebe: The NULL instances are each given a unique library class name 'NULL<n>'. So if there is a single NULL instance defined for DXE_DRIVER, DXE_CORE, and UEFI_DRIVER then 3 NULL library names will be generated labelled NULL0, NULL1, and NULL2. We could split the NULL off and check that the remaining part is an integer, but that's probably not necessary here. The startswith("NULL") pattern is used to idetify NULL linked libraries in several places in Basetools.

.pytool/CISettings.py
TaylorBeebe: Ah yes Joey committed those to 202311 while this was in flight. I'll cherry pick that commit in this series.

TaylorBeebe: Updated

PR #730, opened 2024-02-09
Remove static initialization of gMmst for MM Core Lib

cfernald:
## Description

Removes static initialization for gMmst in the MM Core implementation of MmServicesTableLib. Static initialization of this structure is problematic as callers will use the initialized value as a indication services are ready for use. This results in premature calls to memory allocations causing early faults in the core.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested boot with change.

## Integration Instructions

N/A


kuqin12 (approved):

os-d (approved):
Is this something that should go upstream?

PR #752, opened 2024-02-28
Fix section data length always 4 bytes larger than real data

kuqin12:
## Description

This change fixed an issue where the returned section data length is always 4 bytes larger than the real section. This would cause an issue where the caller could read into the final 4 bytes which is invalid data region.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is verified on QEMU Q35 platform and booted to UEFI shell.

## Integration Instructions

N/A


StandaloneMmPkg/Library/FvLib/FvLib.c
```diff
@@ -380,10 +380,10 @@ FfsFindSectionData (
     if (Section->Type == SectionType) {
       if (IS_SECTION2 (Section)) {
         *SectionData     = (VOID *)((EFI_COMMON_SECTION_HEADER2 *)Section + 1);
-        *SectionDataSize = SECTION2_SIZE (Section);
+        *SectionDataSize = SECTION2_SIZE (Section) - sizeof (EFI_COMMON_SECTION_HEADER2);
```
makubacki, line 383: Matches similar code like [MinPlatformPkg/Library/PeiLib/PeiLib.c](https://github.com/microsoft/mu_common_intel_min_platform/blob/6d75b45170e1f0f4cd5b90a0540246844c1a6571/MinPlatformPkg/Library/PeiLib/PeiLib.c#L348). However, I think the `SectionDataSize` parameter should be clarified to mention the header is not included (especially for people who may come across this change versus old behavior).

kuqin12: Agreed. I think the variable name is self-explanatory, which should reflect the size of the data, instead of the section length. But I added a short explanation to cover such change.

makubacki (approved):

Javagedes (approved):

PR #764, opened 2024-03-09
[REBASE & FF] Several Fixes for 64k Runtime Page Allocation Granularity

os-d:
## Description

This PR is a collection of patches to edk2 (some in slightly different forms due to changes in Mu vs the upstream). They are intended to be pulled into Mu before the upstream accepts them, to fix active bugs before the upstream cadence allows for them to be accepted.

These commits fix a series of issues that have prevented Project Mu from using 64k runtime page allocation granularity for ARM64. The commit messages each have greater detail on the change in question, but in summary:

A CodeQL error is fixed that was uncovered by making a change to Page.c
A UEFI spec 2.10 violation is fixed by changing EfiACPIReclaimMemory to EfiReservedMemoryType for memory types that should have runtime page allocation granularity
Page and pool guards are not set for EfiACPIMemoryNVS, EfiReservedMemoryType, EfiRuntimeServicesCode, and EfiRuntimeServicesData for systems with a runtime page allocation granularity greater than EFI_PAGE_SIZE as the heap guard system does not support this
The HOB based memory protections are updated to ensure the above point is consistent in platform provided HOB
Image Records are fixed to correctly report the size of the images as aligned to the section alignment

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [x] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on multiple physical and virtual platforms, both x86 and ARM64.

## Integration Instructions

Page or Pool guards that are being set on EfiACPIMemoryNVS, EfiReservedMemoryType, EfiRuntimeServicesCode, or EfiRuntimeServicesData for systems that do not have a runtime page allocation granularity equal to the EFI_PAGE_SIZE (ARM64 is the main example, after the revert PR goes in) will need to be removed. A system will not break as a result, the FW will print a warning and remove these memory protections.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Mem/Page.c
+++ b/MdeModulePkg/Core/Dxe/Mem/Page.c
@@ -1608,9 +1608,17 @@ CoreInternalFreePages (
     goto Done;
   }
 
+  // MU_CHANGE: 64k exposed CodeQL issue
+  if (Entry == NULL) {
+    ASSERT (Entry != NULL);
+    Status = EFI_NOT_FOUND;
+    goto Done;
+  }
+
+  // MU_CHANGE END
+
   Alignment = DEFAULT_PAGE_ALLOCATION_GRANULARITY;
 
-  ASSERT (Entry != NULL);
   if ((Entry->Type == EfiACPIReclaimMemory) ||
       (Entry->Type == EfiACPIMemoryNVS) ||
       (Entry->Type == EfiRuntimeServicesCode) ||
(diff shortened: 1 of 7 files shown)


TaylorBeebe (approved):

makubacki (approved):
Approving. We talked about cleaning up the commit messages offline.

apop5 (approved):

PR #768, opened 2024-03-13
[REBASE && FF] Apply EFI_MEMORY_RP on Free Memory and Fix Unsafe Allocation Checking

TaylorBeebe:
This PR:
1. Adds the ability to apply EFI_MEMORY_RP on free memory
2. Fixes an unsafe allocation check so the logic branches if the allocation failed instead of just ASSERTing

### DIFF
--- a/MdeModulePkg/Core/Dxe/DxeMain.inf
+++ b/MdeModulePkg/Core/Dxe/DxeMain.inf
@@ -183,6 +183,7 @@
   gMemoryProtectionDebugProtocolGuid            ## SOMETIMES_PRODUCES        ## MS_CHANGE
   gEfiMemoryAttributeProtocolGuid               ## CONSUMES                  ## MS_CHANGE
   gMemoryProtectionSpecialRegionProtocolGuid    ## PRODUCES                  ## MU_CHANGE
+  gEdkiiGcdSyncCompleteProtocolGuid             ## CONSUMES                  ## MU_CHANGE
 
 [Pcd]
   gEfiMdeModulePkgTokenSpaceGuid.PcdLoadFixAddressBootTimeCodePageNumber    ## SOMETIMES_CONSUMES
--- a/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
+++ b/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
@@ -252,8 +252,14 @@ FindGuardedMemoryMap (
                  &MapMemory,
                  FALSE
                  );
-      ASSERT_EFI_ERROR (Status);
-      ASSERT (MapMemory != 0);
+      // MU_CHANGE START: Check if memory was successfully allocated
+      if (EFI_ERROR (Status) || (MapMemory == 0)) {
+        ASSERT_EFI_ERROR (Status);
+        ASSERT (MapMemory != 0);
+        return 0;
+      }
+
+      // MU_CHANGE END
       // MU_CHANGE START: Apply Protection policy to the allocated memory
       ApplyMemoryProtectionPolicy (
         EfiConventionalMemory,
@@ -291,8 +297,14 @@ FindGuardedMemoryMap (
                  &MapMemory,
                  FALSE
                  );
-      ASSERT_EFI_ERROR (Status);
-      ASSERT (MapMemory != 0);
+      // MU_CHANGE START: Check if memory was successfully allocated
+      if (EFI_ERROR (Status) || (MapMemory == 0)) {
+        ASSERT_EFI_ERROR (Status);
+        ASSERT (MapMemory != 0);
+        return 0;
+      }
+
+      // MU_CHANGE END
       // MU_CHANGE START: Apply Protection policy to the allocated memory
       ApplyMemoryProtectionPolicy (
         EfiConventionalMemory,
--- a/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
+++ b/MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
@@ -16,6 +16,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 //
 GLOBAL_REMOVE_IF_UNREFERENCED BOOLEAN  mOnGuarding = FALSE;
 
+extern BOOLEAN  mPageAttributesInitialized; // MU_CHANGE
+
 //
 // Pointer to table tracking the Guarded memory with bitmap, in which  '1'
 // is used to indicate memory guarded. '0' might be free memory or Guard
@@ -252,6 +254,14 @@ FindGuardedMemoryMap (
                  );
       ASSERT_EFI_ERROR (Status);
       ASSERT (MapMemory != 0);
+      // MU_CHANGE START: Apply Protection policy to the allocated memory
+      ApplyMemoryProtectionPolicy (
+        EfiConventionalMemory,
+        EfiBootServicesData,
+        MapMemory,
+        ALIGN_VALUE (Size, EFI_PAGE_SIZE)
+        );
+      // MU_CHANGE END
 
       SetMem ((VOID *)(UINTN)MapMemory, Size, 0);
 
@@ -283,6 +293,14 @@ FindGuardedMemoryMap (
                  );
       ASSERT_EFI_ERROR (Status);
       ASSERT (MapMemory != 0);
+      // MU_CHANGE START: Apply Protection policy to the allocated memory
+      ApplyMemoryProtectionPolicy (
+        EfiConventionalMemory,
+        EfiBootServicesData,
+        MapMemory,
+        ALIGN_VALUE (Size, EFI_PAGE_SIZE)
+        );
+      // MU_CHANGE END
 
       SetMem ((VOID *)(UINTN)MapMemory, Size, 0);
       *GuardMap = MapMemory;
@@ -560,6 +578,13 @@ UnsetGuardPage (
     Attributes |= EFI_MEMORY_XP;
   }
 
+  // MU_CHANGE START: Add support for RP on free mem
+  if (gDxeMps.FreeMemoryReadProtected) {
+    Attributes |= EFI_MEMORY_RP;
+  }
+
+  // MU_CHANGE END
+
   //
   // Set flag to make sure allocating memory without GUARD for page table
   // operation; otherwise infinite loops could be caused.
@@ -690,10 +715,10 @@ IsHeapGuardEnabled (
   UINT8  GuardType
   )
 {
-  // MU_CHANGE START Update to work with memory protection settings HOB
+  // MU_CHANGE START: Update to work with memory protection settings HOB,
+  //                  remove freed memory guard.
   if ((GuardType & GUARD_HEAP_TYPE_PAGE && gDxeMps.HeapGuardPolicy.Fields.UefiPageGuard) ||
-      (GuardType & GUARD_HEAP_TYPE_POOL && gDxeMps.HeapGuardPolicy.Fields.UefiPoolGuard) ||
-      (GuardType & GUARD_HEAP_TYPE_FREED && gDxeMps.HeapGuardPolicy.Fields.UefiFreedMemoryGuard))
+      (GuardType & GUARD_HEAP_TYPE_POOL && gDxeMps.HeapGuardPolicy.Fields.UefiPoolGuard))
   {
     return TRUE;
   }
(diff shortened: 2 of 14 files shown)


Docs/feature_memory_protection.md
```diff
@@ -284,7 +284,6 @@ Overhead is same as [Page Guards](#page-guards) and [Pool Guards](#pool-guards).
 
 - UefiPageGuard - Enable UEFI page guard
 - UefiPoolGuard - Enable UEFI pool guard
-- UefiFreedMemoryGuard - Enable Use-After-Free memory detection
```
os-d: Should we update this doc to indicate that freed memory is RP by default?

MdeModulePkg/Core/Dxe/DxeMain.inf
```diff
@@ -182,6 +182,7 @@
   gMemoryProtectionDebugProtocolGuid            ## SOMETIMES_PRODUCES        ## MS_CHANGE
   gEfiMemoryAttributeProtocolGuid               ## CONSUMES                  ## MS_CHANGE
   gMemoryProtectionSpecialRegionProtocolGuid    ## PRODUCES                  ## MU_CHANGE
+  gEdkiiGcdSyncComplete                         ## CONSUMES                  ## MS_CHANGE
```
os-d: nit: MU_CHANGE

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -252,6 +254,14 @@ FindGuardedMemoryMap (
                  );
       ASSERT_EFI_ERROR (Status);
       ASSERT (MapMemory != 0);
+      // MU_CHANGE START: Apply Protection policy to the allocated memory
+      ApplyMemoryProtectionPolicy (
+        EfiConventionalMemory,
+        EfiBootServicesData,
+        MapMemory,
+        ALIGN_VALUE (Size, EFI_PAGE_SIZE)
```
os-d, line 268: nit: using same `EFI_SIZE_TO_PAGES` as above.

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
```diff
@@ -83,6 +82,10 @@ typedef struct {
   //  FALSE - UEFI Stack Guard will be disabled.
   BOOLEAN                                   CpuStackGuard;
 
+  // If enabled, all EfiConventionalMemory will be marked with EFI_MEMORY_RO. This can
+  // be used in conjunction with the NX setting for EfiConventionalMemory.
+  BOOLEAN                                   MarkFreeMemoryReadOnly;
```
os-d: Two things:
- This should be `MarkFreeMemoryReadProtected` not ReadOnly.
- I thought we were not having this configurable via the HOB? I.e. you just got a system with RP on free memory because we at the core level decided that was best for you? It wouldn't surprise me if then in compatibility mode we need to disable it. What are your thoughts here?

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -727,18 +730,25 @@ GetPermissionAttributeForMemoryType (
...
   }
 
-  return 0;
+  if (GetDxeMemoryTypeSettingFromBitfield (MemoryType, gDxeMps.NxProtectionPolicy)) {
+    Attributes |= EFI_MEMORY_XP;
+  }
 
-  // MU_CHANGE END
+  if ((MemoryType == EfiConventionalMemory) && gDxeMps.MarkFreeMemoryReadOnly && mPageAttributesInitialized) {
+    Attributes |= EFI_MEMORY_RP;
```
os-d, line 541: I was thinking that we would do this in Page.c itself, on AllocatePages we would remove RP, on FreePages we would set RP. This gets called out from a level of AllocatePages, but is this just to check the HOB? Or to ensure it is happening after the GCD sync? My concern of doing it at this level is that, as earlier in the PR shows, we have code that calls around it and any future code that calls around it would have to do the same thing or it will break.

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
TaylorBeebe: - Ahhh yes the first time I tackled this we were marking it read only. I'll update the refs to be read protected.
- At some point we will remove these configuration options but we aren't quite there yet. This change in particular may cause some friction when integrated into platforms so it's probably better to make it toggleable.

TaylorBeebe:
The following PRs should go in before this one:
1. https://github.com/microsoft/mu_silicon_arm_tiano/pull/220
2. https://github.com/microsoft/mu_tiano_platforms/pull/876
3. https://github.com/microsoft/mu_plus/pull/450

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
TaylorBeebe: I could do EFI_PAGES_TO_SIZE(EFI_SIZE_TO_PAGES(Size)) but it's much easier to align the value here

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
TaylorBeebe: AllocateMemoryProtection policy is how all attributes are set outside of image memory. There were some cases where attributes weren't being applied, but all were contained within DxeCore where the internal functions are accessible outside of a protocol or MemoryAllocationLib. I think we should be out of the woods on those missed allocations -- especially with this update.

What you're suggesting can be done in a future update but I prefer to stick with the current pattern for this.

MdeModulePkg/Core/Dxe/DxeMain.inf
TaylorBeebe: Updated

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
TaylorBeebe: Updated to FreeMemoryReadProtected

Docs/feature_memory_protection.md
TaylorBeebe: I'm going to do a separate PR for the full doc update but it's almost done

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1120,37 +1131,38 @@ CoreInitializeMemoryProtection (
...
-  //
-  Status = CoreRegisterProtocolNotify (
-             &gEfiCpuArchProtocolGuid,
-             Event,
-             &Registration
-             );
-  ASSERT_EFI_ERROR (Status);
+  // Status = CoreCreateEvent (
+  //            EVT_NOTIFY_SIGNAL,
+  //            // MU_CHANGE START: Use Project Mu Arch Protocol Notify
```
os-d: This is commenting out another MU_CHANGE, can we either revert that change if it is no longer applicable in entirety or drop the MU_CHANGE parts of these comments in a separate commit that we can mark as something to fix up on rebase?

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1300,8 +1319,10 @@ ApplyMemoryProtectionPolicy (
   UINT64  OldAttributes;
   UINT64  NewAttributes;
 
+  // DEBUG ((DEBUG_INFO, "ApplyMemoryProtectionPolicy: 0x%016lx - 0x%016lx\n", Memory, Length));
```
os-d: Is this commented out now because the call occurs so much? Can we add a MU_CHANGE here?

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
```diff
@@ -456,7 +456,8 @@ extern GUID  gDxeMemoryProtectionSettingsGuid;
 #define DXE_MEMORY_PROTECTION_SETTINGS_OFF                      \
           {                                                     \
             DXE_MEMORY_PROTECTION_SETTINGS_CURRENT_VERSION,     \
-            FALSE,   /* Stack Guard On */                       \
+            FALSE,   /* Stack Guard Off */                      \
+            FALSE,   /* Free Memory Guard Off */                \
```
os-d, line 462: I'm honestly tempted to not make it a choice. This will blow platforms up to be sure. But this is also something that they really need to fix. I understand this is a major change, so I'm ok with having this configurable for now, but I feel like in the next iteration or two we should drop the configurability, unless we find a hard block in a bootloader or something.

os-d (approved):

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
TaylorBeebe: Agreed on all points. I think this may cause issues in the short term for platforms which are misusing free memory so it's good to have a policy configuration but hopefully they will be ironed out soon and we can remove this option.

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
TaylorBeebe: I must have added this print when testing and forgotten about it -- removed

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
TaylorBeebe: Reverted this code group to what it is in EDK2 and commented it out.

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
```diff
@@ -252,6 +254,14 @@ FindGuardedMemoryMap (
                  );
       ASSERT_EFI_ERROR (Status);
```
makubacki: Not entirely related to these changes, but this is the kind of place something other than just asserting should be done if the allocation fails.

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
makubacki, line 1: Might be helpful to link to the memory protection markdown file at the top of this file for visibility to the reference.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -833,6 +833,10 @@
   ## Include/Protocol/UsbEthernetProtocol.h
   gEdkIIUsbEthProtocolGuid = { 0x8d8969cc, 0xfeb0, 0x4303, { 0xb2, 0x1a, 0x1f, 0x11, 0x6f, 0x38, 0x56, 0x43 } }
 
+  ## MU_CHANGE START: Add protocol to indicate the GCD sync has completed
+  ## Include/Protocol/Cpu.h
+  gEdkiiGcdSyncComplete                 = { 0x650B7F40, 0x6564, 0x4FA9, { 0x93, 0x75, 0xCD, 0x6B, 0x52, 0x1B, 0x6E, 0x50 }}
```
makubacki: Suggest to follow the naming convention with a `ProtocolGuid` suffix.

```suggestion
  gEdkiiGcdSyncCompleteProtocolGuid = { 0x650B7F40, 0x6564, 0x4FA9, { 0x93, 0x75, 0xCD, 0x6B, 0x52, 0x1B, 0x6E, 0x50 }}
```

makubacki (approved):

MdeModulePkg/Core/Dxe/Mem/HeapGuard.c
TaylorBeebe: Converted this to a REBASE && FF PR. The second commit updates this section and another similar section to branch execution if the allocation fails in addition to ASSERTing.

MdeModulePkg/Include/Guid/DxeMemoryProtectionSettings.h
TaylorBeebe: Added a link to https://github.com/microsoft/mu_basecore/tree/HEAD/Docs/feature_memory_protection.md in both the DXE and MM settings headers.

MdeModulePkg/MdeModulePkg.dec
TaylorBeebe: Updated all references to gEdkiiGcdSyncCompleteProtocolGuid

PR #770, opened 2024-03-14
EccCheck update to later tools. Change EccCheck plugin scope. Add EccCheck 

apop5:
## Description

EccCheck is a tool that verifies additional Efi Code Compatibility for files that have been modified. 

EccCheck relies on the basetool/Source/Python/Ecc.
Ecc contained files that were generated for parsing C files, but these were tied antlr4-runtime 4.7.1.

Documented instructions for generating new Ecc parser files/

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Ran tests locally on repo after adding scope `cibuild-edk2`
Modified different files within packages to see if CI error would be tripped.

## Integration Instructions
If wanting to use the EccCheck plugin, add `cibuild-edk2` to the scope of the build files.


.pytool/Plugin/EccCheck/EccCheck.py
```diff
@@ -138,11 +138,23 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
             return 1
 
     def GetDiff(self, pkg: str, temp_diff_output: str) -> List[str]:
+        # MU_CHANGE [BEGIN]
+        #
+        # Run git command to determine repos default branch
+        #
+        result = StringIO()
+        params = "symbolic-ref refs/remotes/origin/HEAD --short"
```
Javagedes, line 146: I agree the current way of hardcoding to origin/master is bad for us, but I think this way will also cause issues if we are targeting the non-default branch on CI. Such as cherry-picking changes back or if you have a PR targeting a feature branch.

apop5: I agree that this tool is not optimal. Thankfully, this is not being enabled as part of CI, its scope is just being changed to `cibuild-edk2` to enable mu based projects to be able to run the edk2 tool locally. 

Comparing against the default branch may seem problematic, but the way the tool works is it uses the lines changed between the default branch and the current branch and filters the ecc check log to only report the errors corresponding to the changed lines. This means that the tool will only ever check a subset of the entire files for errors, and that a lot of things which do not conform to the coding standards will be able to slip through repeatedly until someone changes something in those lines. 

As for cherry-picking changes, our most common cherry-pick source is from upstream, which runs this tool as part of their CI process. So cherry-picking most change from upstream should, for the most part, already have passed this check.

makubacki: I think Joey's saying that we often work on non-default branches. For example, release/202311 (default) and release/202302. In that case, cherry-picking into release/202302 will have the files compared to their equivalents in release/202311 which might be in a different base state which would result in an invalid delta.

The patchset only diff is okay I think but it needs to be correct.

PR #773, opened 2024-03-19
Rework Buildtime Random Stack Cookie Values to Improve Incremental Build Times and Ensure Binary Reproducibility

TaylorBeebe:
## Description

If the stack cookie value is randomized in the AutoGen.h file each build, the build system will determine the module/library must be rebuilt causing effectively a clean build every time. This also makes binary reproducibility impossible.

This PR updates the early build scripts to create 32 and 64-bit JSON files in the build output directory which each contain 100 randomized stack cookie values for each bitwidth. If the JSON files are already present, then they are not recreated which allows them to be stored and moved to other builds for binary reproducibility. Because they are in the build directory, a clean build will cause the values to be regenerated.

The logic which creates AutoGen.h will read these JSON files and use a hash of the module GUID (the hash seed is fixed in Basetools) to index into the array of stack cookie values for the module bitwidth. This model is necessary because there isn't thread-consistent data so we cannot use a locking mechanism to ensure only one thread is writing to the stack cookie files at a time. With this model, the build threads only need to read from the files.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by building and confirming the reduced build time for debug builds after the initial build.

## Integration Instructions

N/A


BaseTools/Source/Python/AutoGen/GenC.py
makubacki, line 1: This still regresses incremental RELEASE builds which is going to be an issue at the scale of builds supported here and impacts binary reproducibility.

I think this either needs to be very simple - e.g. only generate a new value on clean builds otherwise it remains the same, in which case the instructions to generate a new value are easy to understand and its operation is easy to follow. Or allow flexibility with something more complicated but abstract - e.g. static values by default; can associate cookie refresh with abstract signal from platform (which can use SHIP_MODE or whatever).

The implementation details could of course take several directions.

TaylorBeebe: Discussed offline -- changes incoming

os-d (approved):

makubacki (approved):

Javagedes (approved):

PR #782, opened 2024-03-29
Add CRC16 CCITT False Implementation

kuqin12:
# Preface

## Description

This change is added to incorporate basic implementation for CRC16-CCITT-FALSE algorithm.

This function is useful for providing CRC16 value used in other data structures that requires CRC16 value that complies with JEDEC SPD requirements, i.e. BDAT table.

The lookup table is inherited from `https://crccalc.com/` and the result values are also compared against this site.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

The function output is compared and matches with the results of 3rd party online CRC calculators.

## Integration Instructions

N/A


MdePkg/Library/BaseLib/CheckSum.c
```diff
@@ -762,3 +762,71 @@ CalculateCrc32c (
...
+CalculateCrc16CcittF (
+  IN CONST VOID  *Buffer,
+  IN UINTN       Length,
+  IN UINT16      InitialValue
+  )
+{
+  CONST UINT8  *Buf;
+  UINT16       Crc;
+
+  Buf = Buffer;
```
apop5, line 825: Buffer needs validated before it's used

apop5 (approved):

MdePkg/Library/BaseLib/CheckSum.c
kuqin12: I thought about adding it, but any idea on what i should return if validation fails?

apop5: CalculateCrc32 (in the same file) does the following:

  ASSERT (Buffer != NULL);
  ASSERT (Length <= (MAX_ADDRESS - ((UINTN)Buffer) + 1));
  
  I think we could just do the same here, and then return 0 if it is NULL.

kuqin12: Will update shortly, thanks!

kuqin12: Updated~

cfernald (approved):

PR #792, opened 2024-04-06
Refine UncrustifyCheck

yangrongwei:
## Description

1. Report Newlines only diff.
2. Keep the formatted file in filesystem if unable to diff by code.
3. Refine error report, so we can at least know the filename in build log. 
4. Let -v same as TestCase StdError, so we can check error directly in local build.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [X ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

I tested it on my private Pkg

## Integration Instructions

The breaking change is I added a `_keep_formatted_file` attribute in Python script.

```
If OutputFileDiffs == True AND 'we are unable to diff file by code'
    _keep_formatted_file  was set  AND 
   _cleanup_temporary_formatted_files(self) will NOT delete formatted file
```

This is reasonable. The old behavior continue reporting formatting errors, but I am unable to understand the reason. From the `TestSuites.xml` file, I can only know the filename.

Why NOT,
1. Report the filename directly ?
2. **Keep the formatted file** and let user find the reason if code cannot find diff?
3. Dump the error report for `TestSuites.xml` directly when `-v` is used? 

Finally, after changing the `UncrustifyCheck.py` here and there and using other diff tools, I found my `HelloWorld.c` is Newlines only diff! That is the story. It took me two days.


yangrongwei:
@microsoft-github-policy-service agree [company="{Raymond}"]

yangrongwei:
@microsoft-github-policy-service agree [company="Raymond"]

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
makubacki, line 1: Hi @yangrongwei, thank you for the contribution. I'll begin by saying various aspects of UncrustifyCheck have been contentious in the past, in part because formatting tends to incite a range of varying opinions that extend into how formatting is enforced and displayed in various interfaces such as the developer's terminal, build log, and test case output. Also, because the UncrustifyCheck plugin is not necessarily meant to encapsulate all the complexity of using Uncrustify. It is meant to verify that Uncrustify was run successfully such that no errors exist before code is merged. A developer should understand how to use Uncrustify in their code development process outside the plugin.

**Errors in Build Log**

The original UncrustifyCheck displayed file diffs in the build output. That was modified upon request to limit output to the test case file so it could be viewed in CI and prevent "overcrowding" build output. The link to this page [`EDK II Code Formatting`](https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-Formatting) was added in the UncrustifyCheck output to explain:

- How to run Uncrustify locally in an IDE like VS Code where the diffs are obvious and immediate
- Where to find diff results in an Azure Pipeline CI environment
- What to do when diff results are empty - indicating a line ending issue

So, file diffs in the build log were rejected earlier, they may be accepted now. I don't have a strong opinion. I think at least a filename in build output is reasonable in any case.

**Newlines Only Diff**

Line ending checks were explicitly left out of the plugin because they're covered by two other plugins either run in Project Mu ([Line Ending Check](https://github.com/microsoft/mu_basecore/blob/822b984a66116b1e4430980175ea497180bad85f/.pytool/Plugin/LineEndingCheck/Readme.md)) or edk2 ([PatchCheck](https://github.com/tianocore/edk2/blob/b7f8779fe1f60113fdaab3b2f3f17c9f900b0456/BaseTools/Scripts/PatchCheck.py)). The idea was to simply run Uncrustify in the UncrustifyCheck plugin and leave line ending checks to other plugins to consolidate line ending logic and results in CI, etc. To keep in line with those prior decisions, I suggest dropping this.

**Keep Formatted Files in Filesystem**

The intermediate/temporary files are artifacts of how diffs are calculated, they are not meant to impact or be retained on a user's filesystem unless the deletion operation fails. This expectation in behavior should not be modified. There are a variety of other ways users can get diffs:

- Run Uncrustify directly following instructions on the wiki page on a per package or per file basis
- Run Uncrustify in VS Code on a per file basis
- Run the UncrustifyCheck plugin with `UNCRUSTIFY_IN_PLACE=TRUE` to modify files in place

For all of these, git can be used to manage what to do with file diffs produced. Uncrustify directly also allows file diffs to be saved off to the side without impacting the original file. Therefore, I suggest dropping this.

**Upstream Integration**

This plugin is [in edk2](https://github.com/tianocore/edk2/tree/master/.pytool/Plugin/UncrustifyCheck), so I recommend modifications go there first and then cherry-picked back here. If made here first, we'd have to start a new review cycle of the patch that is open to wider community feedback in edk2 and some of the changes might not be accepted causing volatily in behavior for Project Mu users.

**General Change Feedback**

- Any changes retained should update the [`UncrustifyCheck Plugin Readme`](https://github.com/microsoft/mu_basecore/blob/release/202311/.pytool/Plugin/UncrustifyCheck/Readme.md).
- Some of the changes are in style only. Those should be made in a separate commit from functional changes.

---

This was overall feedback. I will supplement it with some specific comments in the file.

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
```diff
@@ -213,12 +247,14 @@ def _check_for_preexisting_formatted_files(self) -> None:
...
+        pre_existing_formatted_files = [str(path.resolve()) for path in pathlib.Path(self._abs_package_path).rglob(f'*{UncrustifyCheck.FORMATTED_FILE_EXTENSION}')]
+        pre_existing_formatted_file_count = len(pre_existing_formatted_files)
 
         if pre_existing_formatted_file_count > 0:
-            raise UncrustifyStalePluginFormattedFilesException(
-                f"{pre_existing_formatted_file_count} formatted files already exist. To prevent overwriting these files, please remove them before running this plugin.")
+            err_str  = f"{pre_existing_formatted_file_count} formatted files already exist. \n" \
+                       f"To prevent overwriting these files, please remove them before running this plugin. \n" \
+                       f"{'\n'.join(pre_existing_formatted_files) } \n"
+            raise UncrustifyStalePluginFormattedFilesException(err_str)
```
makubacki, line 258: I think this change is worth keeping regardless of dropping `_keep_formatted_file`. In the rare case files remain a list of those files is valuable.

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
```diff
@@ -236,6 +272,11 @@ def _cleanup_temporary_formatted_files(self) -> None:
         This will recursively remove all formatted files generated by Uncrustify
         during this execution instance.
         """
+        # Something wrong with our OutputFileDiffs function, check _process_uncrustify_results(self)
+        # OR turn off OutputFileDiffs
+        if hasattr(self, '_keep_formatted_file'):
```
makubacki, line 277: As noted, please drop this feature.

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
```diff
@@ -25,6 +25,40 @@
 from edk2toollib.utility_functions import  RunCmd
 from io import StringIO
 from typing import Any, Dict, List, Tuple
+from enum import Enum
+
+class NewLineStyle(Enum):
```
makubacki, line 30: As noted, please make changes to the existing line ending check plugins if you determine improvements are needed.

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
```diff
@@ -599,20 +653,33 @@ def _process_uncrustify_results(self) -> None:
...
+                        
+                        content_diff = False
                         for line in difflib.unified_diff(pre_formatted_file_text.split('\n'), formatted_file_text.split('\n'), fromfile=pre_formatted_file, tofile=formatted_file, n=3):
-                            self._tc.LogStdError(line)
-
-                        self._tc.LogStdError('\n')
+                            content_diff = True
+                            self._dual_out(line)
+                        
+                        if not content_diff:
```
yangrongwei, line 671: @makubacki I think we can give a prompt here for (a) line ending check, (b) the way to diff by ourselves

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
yangrongwei: > This plugin is [in edk2](https://github.com/tianocore/edk2/tree/master/.pytool/Plugin/UncrustifyCheck), so I recommend modifications go there first and then cherry-picked back here.

I see. I will split my changes in different PRs for review.

.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
yangrongwei: I see.

PR #794, opened 2024-04-08
Map Legacy BIOS Memory RWX When Entering Compatibility Mode

TaylorBeebe:
## Description

The Linux distro used in Mariner 2 has logic [directly carves memory from the legacy BIOS region](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2FCBL-Mariner-Linux-Kernel%2Fblob%2Fedab6ad780cfa0be041a08a79b600443fde10c7f%2Farch%2Fx86%2Fboot%2Fcompressed%2Fpgtable_64.c%23L38&data=05%7C02%7CTaylor.Beebe%40microsoft.com%7Cd6a0f39431794d716fc608dc55c9f96b%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638479573168984897%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=2YR1egvp4z%2BCmMu5aeQnZvZegQTRhdVO0vTzIF5cBAc%3D&reserved=0) and copies the trampoline code into it for execution. Mariner 2 will trigger memory protection compatibility mode due to the absence of the NX_COMPAT flag in Shim, but because the UEFI allocator is not used for this allocation, the memory attributes won't be updated to make the region executable. To resolve this, compatibility mode will now map the writable legacy BIOS region (0x0-0xa0000) as RWX to resolve this issue with Mariner 2 and other Linux distros older than a couple of years.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 by booting with a recent a Mariner 2 image.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -3559,7 +3559,7 @@ UninstallMemoryAttributeProtocol (
 }
 
 /**
-  Maps memory below 1MB (legacy BIOS memory) as readable, writeable, and executable.
+  Maps memory below 640K (legacy BIOS write-back memory) as readable, writeable, and executable.
 **/
 STATIC
 VOID
@@ -3569,11 +3569,14 @@ MapLegacyBiosMemoryRWX (
 {
   EFI_STATUS  Status;
 
+  // https://wiki.osdev.org/Memory_Map_(x86)
+  //
+  // Map the legacy BIOS write-back memory as RWX.
   if (gCpu != NULL) {
     Status = gCpu->SetMemoryAttributes (
                      gCpu,
                      0x0,
-                     0x100000,
+                     0xa0000,
                      0
                      );
   }
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -3559,6 +3559,29 @@ UninstallMemoryAttributeProtocol (
 }
 
 /**
+  Maps memory below 1MB (legacy BIOS memory) as readable, writeable, and executable.
+**/
+STATIC
+VOID
+MapLegacyBiosMemoryRWX (
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+
+  if (gCpu != NULL) {
+    Status = gCpu->SetMemoryAttributes (
+                     gCpu,
+                     0x0,
+                     0x100000,
+                     0
+                     );
+  }
+
+  ASSERT_EFI_ERROR (Status);
+}
+
+/**
   Sets the NX compatibility global to FALSE so future checks to
   IsEnhancedMemoryProtectionActive() will return FALSE.
 **/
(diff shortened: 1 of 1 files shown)


apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -3558,6 +3558,29 @@ UninstallMemoryAttributeProtocol (
...
+  VOID
+  )
+{
+  EFI_STATUS  Status;
+
+  if (gCpu != NULL) {
+    Status = gCpu->SetMemoryAttributes (
+                     gCpu,
+                     0x0,
+                     0x100000,
```
makubacki: To my understanding we should effectively be able to limit this from `0x0` to `0xA0000`. Where `0xA0000` to `0x100000` are hardware mapped ranges. In most cases, fixed MTRRs are used to set memory types for the Legacy BIOS range (detectable via the `FE` flag in the `IA32_MTRR_DEF_TYPE` MSR) and would produce ranges like:

```
WB:0000000000000000-000000000009FFFF
UC:00000000000A0000-00000000000BFFFF
WP:00000000000C0000-00000000000FFFF
```

Where the area below `0xA0000` would be outside the uncached video memory and write-protected BIOS flash areas. Fixed MTRRs aside though, I wouldn't expect code to be written and executed from that upper range. I mentioned this since part of the memory protection effort is minimizing privilege and this seems like an opportunity to do so.

TaylorBeebe: The EFI stub does indeed limit its search for usable memory for the trampoline code to 0x0-0xa0000, updating this PR to only set that range to match the MTRR WB range.

TaylorBeebe: Updated. I also included the link to the OS wiki page outlining the memory layout of legacy bios.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -3558,6 +3558,32 @@ UninstallMemoryAttributeProtocol (
...
+  if (gCpu != NULL) {
+    Status = gCpu->SetMemoryAttributes (
+                     gCpu,
+                     0x0,
+                     0xa0000,
+                     0
+                     );
+  }
+
+  ASSERT_EFI_ERROR (Status);
```
makubacki, line 3584: You'll need to initialize `Status` for the case when `gCpu` is null.

makubacki (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Good catch -- fixed

PR #797, opened 2024-04-09
CryptoPkg/BaseCryptLib: add DigestLen to RsaOaepEncrypt(), RsaOaepDecrypt()

cmruffin:
## Description

Add DigestLen parameter to RsaOaepEncrypt(), RsaOaepDecrypt() so that SHA256, SAH384, SHA512 hash and MGF can be used with the API.

- [ ] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Unit tests, wycheproof test vectors, end-to-end test with 3rd party application.

## Integration Instructions

Build crypto release on private branch, update Crypto Driver with this PR.

(contains changes from edk2 commit 89ff5da9f9787be29c173ad5d5c0beb0c262e3cc)
(contains changes from edk2 commit 503344cdbd289e6f360f900551d1cc7c23f4b27a)
(contains changes from edk2 commit c98c14576f1a93519402a8e2f4736b479ad6007c)


makubacki (approved):
Approving since the changes look complete to me with a comment left in file feedback to block in response to it.

CryptoPkg/Include/Protocol/Crypto.h
makubacki, line 1: We had discussed recently trying to limit changes to the crypto protocol that are not backward compatible. Would like to get feedback from @spbrogan on how he would like to treat this particular change.

makubacki: I do think bumping `EDKII_CRYPTO_VERSION` would help indicate the change to the protocol API.

That could allow an abstraction layer that depends on it (like the `OnProtocolPpi` BaseCryptLib instances) to be able to call the protocol correctly based on a produced version.

cmruffin: There are no consumers yet of this api.  The function only went in a couple of weeks ago.  This change aligns with what ended up in edk2.

makubacki: I approved because in practice, I agree. This is an API change in respect to how the API is called (which impacts API callers) but, unlike the previous protocol change which introduced new functions in the middle of the protocol, it does not impact the callers to other functions in the protocol.

Given the time of the API, it is likely you are the only impacted user, so I believe this is acceptable. However, I know `EDKII_CRYPTO_VERSION` only states that it is updated when new APIs are added at the end of the structure, I still think this is a case where it should be updated to reflect an API change.

Flickdm (approved):

PR #798, opened 2024-04-09
Add a Deprecated Build Option To Allow Incompatible AARCH64 Platforms To Opt Out of 64k Runtime Granularity

os-d:
## Description

Some AARCH64 platforms are incompatible with 64k runtime granularity. This PR adds a build option that is consumed to enable a deprecated option to set the runtime page allocation granularity to 4k. This option should not be used except for old platforms that cannot be updated to comply with 64k runtime granularity. Using this option breaks compatibility with the UEFI spec.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with setting flag and unsetting.

## Integration Instructions

Do not use this option without an old platform that cannot support 64k. For those platforms, set the following in the platform DSC:

```c
MSFT:*_*_*_CC_FLAGS = /D __DEPRECATED_AARCH64_4K_RUNTIME_GRANULARITY
GCC:*_*_*_CC_FLAGS = -D __DEPRECATED_AARCH64_4K_RUNTIME_GRANULARITY
```

### DIFF
--- a/MdePkg/Include/AArch64/ProcessorBind.h
+++ b/MdePkg/Include/AArch64/ProcessorBind.h
@@ -159,9 +159,17 @@ typedef INT64 INTN;
 
 ///
 /// Page allocation granularity for AARCH64
+/// MU_CHANGE BEGIN: Enable Deprecated 4k Granularity Mode for Platforms
+/// that do not support 64k runtime allocation
 ///
 #define DEFAULT_PAGE_ALLOCATION_GRANULARITY  (0x1000)
+#ifdef __DEPRECATED_AARCH64_4K_RUNTIME_GRANULARITY
+#define RUNTIME_PAGE_ALLOCATION_GRANULARITY  (0x1000)
+#else
 #define RUNTIME_PAGE_ALLOCATION_GRANULARITY  (0x10000)
+#endif
+
+// MU_CHANGE END
 
 //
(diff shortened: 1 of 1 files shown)


MdePkg/Include/AArch64/ProcessorBind.h
```diff
@@ -159,9 +159,17 @@ typedef INT64 INTN;
 
 ///
 /// Page allocation granularity for AARCH64
+/// MU_CHANGE BEGIN: Enable Deprecated 4k Granularity Mode for Platforms
+/// that do not support 64k runtime allocation
 ///
 #define DEFAULT_PAGE_ALLOCATION_GRANULARITY  (0x1000)
```
apop5, line 165: do you need this define still?

I think its covered by the #ifdef #else

apop5 (approved):

kuqin12 (approved):

MdePkg/Include/AArch64/ProcessorBind.h
os-d: Yes, without this we won't know what size to allocate boot memory :). This is the DEFAULT vs RUNTIME granularity.

TaylorBeebe (approved):

PR #814, opened 2024-04-18
Fix Possible Allocation of Pages While They Are Being Freed

TaylorBeebe:
## Description

FreeMemoryMapStack() within CoreConvertPagesEx() may allocate pages which, if the converted memory is being freed (implied by the new type being EfiConventionalMemory), could cause the memory currently being freed to be allocated during the convert process it if CoreFreeMemoryMapStack() is called after AddRange(). So, if the convert is being done due to a free pages call, free the memory map stack before adding memory to the free list.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Mem/Page.c
+++ b/MdeModulePkg/Core/Dxe/Mem/Page.c
@@ -949,6 +949,21 @@ CoreConvertPagesEx (
       Entry = NULL;
     }
 
+    // MU_CHANGE [BEGIN]
+    // The below call may allocate pages which, if we're freeing memory (implied by
+    // the new type being EfiConventionalMemory), could cause the memory we're currently
+    // freeing to be allocated before we're done freeing it if CoreFreeMemoryMapStack()
+    // is called after AddRange(). So, if we are freeing, let's free the memory map
+    // stack before adding memory we're converting to the free list.
+    if (NewType == EfiConventionalMemory) {
+      //
+      // Move any map descriptor stack to general pool
+      //
+      CoreFreeMemoryMapStack ();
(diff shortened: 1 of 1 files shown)


MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -949,6 +949,21 @@ CoreConvertPagesEx (
       Entry = NULL;
     }
 
+    // MU_CHANGE [BEGIN]
+    // The below call may allocate pages which, if we're freeing memory (implied by
+    // the new type being EfiConventionalMemory), could cause the memory we're currently
+    // freeing to be allocated before we're done freeing it if CoreFreeMemoryMapStack()
+    // is called after AddRange(). So, if we are freeing, let's free the memory map
+    // stack before adding memory we're converting to the free list.
+    if (NewType == EfiConventionalMemory) {
```
os-d: Should we also check if ChangingType is true here?

os-d (approved):

MdeModulePkg/Core/Dxe/Mem/Page.c
TaylorBeebe: I think so, yes. Updated the behavior to only change the FreeMemoryMapStack() ordering if we are changing the type.

kuqin12 (approved):

PR #825, opened 2024-04-23
Generate Random Stack Cookie Values in Module Build Path

TaylorBeebe:
## Description

If only a single module was being built instead of a platform DSC, the build scripts would not generate the random stack cookie values. On clean builds, this causes an error to be reported during the autogen process.

This PR updates build.py so the code path taken on module builds will generate the random stack cookie values.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by specifying a build module on Q35.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/Python/build/build.py
+++ b/BaseTools/Source/Python/build/build.py
@@ -1974,6 +1974,7 @@ class Build():
                         self.Progress,
                         self.ModuleFile
                         )
+                GenerateStackCookieValues() # MU_CHANGE [BEGIN]: Add build-time random stack cookie support
                 self.Fdf = Wa.FdfFile
                 self.LoadFixAddress = Wa.Platform.LoadFixAddress
                 Wa.CreateMakeFile(False)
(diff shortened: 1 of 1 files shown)


apop5 (approved):

BaseTools/Source/Python/build/build.py
```diff
@@ -1974,6 +1974,7 @@ def _BuildModule(self):
                         self.Progress,
                         self.ModuleFile
                         )
+                GenerateStackCookieValues() # MU_CHANGE [BEGIN]: Add build-time random stack cookie support
```
makubacki: You can drop `[BEGIN]` if it is not matching an end tag. Afaict it is not.

makubacki (approved):

BaseTools/Source/Python/build/build.py
TaylorBeebe: Thanks -- I also removed the [BEGIN] from a couple of other spots where it's not necessary.

PR #829, opened 2024-04-25
[REBASE && FF] Add StackCheckLib Readme and Add Function Address Printing on Stack Check Failure

TaylorBeebe:
This PR has two changes:
1. Print the return address when a stack check failure occurs to make debugging easier.
2. Add a readme for StackCheckLib


os-d (approved):
nit: can you update the PR title to something like: `Add StackCheckLib Readme and Cleanup`

apop5 (approved):

PR #835, opened 2024-04-30
[REBASE && FF] Consolidate MU_CHANGEs and Fix RP on Free Mem Bug

TaylorBeebe:
Commit breakdown:
1. Many changes have been made to CoreInitializeMemoryProtection() to improve the security state. It's gotten to the point that the code has been essentially rewritten so to improve readability and maintainability, consolidate the changes to MemoryProtectionSupport.c (a Project Mu file). Creating a CoreInitializeMemoryProtectionMu() function also allows us to move other MU_CHANGEs to MemoryProtectionSupport.c .
2. mPageAttributesInitialized was used to signal that the memory protection initialization routine was complete so EFI_MEMORY_RP can be safely set on free memory. However, the blocker to setting EFI_MEMORY_RP on free memory is not the initialization of the page attributes, but the completion of the GCD sync. This change renames mPageAttributesInitialized to mGcdSyncComplete to better reflect the intent of the variable. Additionally, it makes more sense to not change what GetPermissionAttributeForMemoryType() returns based on the value of mGcdSyncComplete. Instead, it should always return the permission attributes which should be applied to the memory type but ApplyMemoryProtectionPolicy() should bail if the GCD sync has not been completed. 
3. There's a window between when the CPU Arch Protocol is published and the GCD sync is complete where guard pages could be set on allocated memory and unset on freed memory. These operations should not be done until the GCD sync is complete or some free memory may have EFI_MEMORY_RP applied before the memory protection logic expects.

**This PR was tested by running the DxePagingAuditTestApp on Q35 and SBSA platforms. Additionally tested by booting to Windows on Q35.**

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -39,7 +39,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #include <Protocol/FirmwareVolume2.h>
 #include <Protocol/SimpleFileSystem.h>
-#include <Protocol/MemoryProtectionDebug.h> // MU_CHANGE
 #include <Protocol/MemoryAttribute.h>       // MU_CHANGE
 
 #include "DxeMain.h"
@@ -66,23 +65,14 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define PREVIOUS_MEMORY_DESCRIPTOR(MemoryDescriptor, Size) \
   ((EFI_MEMORY_DESCRIPTOR *)((UINT8 *)(MemoryDescriptor) - (Size)))
 
-// UINT32   mImageProtectionPolicy; // MU_CHANGE
+UINT32  mImageProtectionPolicy;
 
 extern LIST_ENTRY  mGcdMemorySpaceMap;
 
 EFI_MEMORY_ATTRIBUTE_PROTOCOL  *mMemoryAttribute = NULL;          // MU_CHANGE
+extern BOOLEAN                 mPageAttributesInitialized;        // MU_CHANGE
 
-STATIC LIST_ENTRY  mProtectedImageRecordList;
-// MS_CHANGE - START
-STATIC MEMORY_PROTECTION_DEBUG_PROTOCOL  mMemoryProtectionDebug =
-{
-  IsGuardPage,
-  GetImageList
-};
-
-BOOLEAN  mPageAttributesInitialized = FALSE;
-
-// MS_CHANGE - END
+STATIC LIST_ENTRY  mProtectedImageRecordList = INITIALIZE_LIST_HEAD_VARIABLE (mProtectedImageRecordList); // MU_CHANGE: Initialize at compile time
 
 /**
   Get the image type.
@@ -992,21 +982,25 @@ MemoryProtectionExitBootServicesCallback (
   }
 }
 
-// MU_CHANGE [START]: Move disable NULL detection to separate routine so it can be called
-//                    outside of the event context.
-
 /**
-  Disable NULL pointer detection.
+  Disable NULL pointer detection after EndOfDxe. This is a workaround resort in
+  order to skip unfixable NULL pointer access issues detected in OptionROM or
+  boot loaders.
+
+  @param[in]  Event     The Event this notify function registered to.
+  @param[in]  Context   Pointer to the context data registered to the Event.
 **/
 VOID
-DisableNullDetection (
-  VOID
+EFIAPI
+DisableNullDetectionAtTheEndOfDxe (
+  EFI_EVENT  Event,
+  VOID       *Context
   )
 {
   EFI_STATUS                       Status;
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  Desc;
 
-  DEBUG ((DEBUG_INFO, "%a - start\r\n", __FUNCTION__));
+  DEBUG ((DEBUG_INFO, "DisableNullDetectionAtTheEndOfDxe(): start\r\n"));
   //
   // Disable NULL pointer detection by enabling first 4K page
   //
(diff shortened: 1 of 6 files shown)


MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -992,21 +982,25 @@ MemoryProtectionExitBootServicesCallback (
...
+DisableNullDetectionAtTheEndOfDxe (
+  EFI_EVENT  Event,
+  VOID       *Context
   )
 {
   EFI_STATUS                       Status;
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR  Desc;
 
-  DEBUG ((DEBUG_INFO, "%a - start\r\n", __FUNCTION__));
+  DEBUG ((DEBUG_INFO, "DisableNullDetectionAtTheEndOfDxe(): start\r\n"));
```
apop5, line 1003:
```suggestion
  DEBUG ((DEBUG_INFO, "%a(): start\r\n", __func__));

```

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1029,249 +1023,91 @@ DisableNullDetection (
...
-{
-  EFI_STATUS                       Status;
-  EFI_GCD_MEMORY_SPACE_DESCRIPTOR  Desc;
-
-  Status = CoreGetMemorySpaceDescriptor (0, &Desc);
-
-  if (EFI_ERROR (Status)) {
-    return;
-  }
+  DEBUG ((DEBUG_INFO, "DisableNullDetectionAtTheEndOfDxe(): end\r\n"));
```
apop5, line 1032:
```suggestion
  DEBUG ((DEBUG_INFO, "%a(): end\r\n", __func__));

```

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
TaylorBeebe: This function is being restored to what it is in EDK2, and this function is unused in Project Mu. Because the purpose of restoring it was to remove mu_change tags, I'd prefer to leave it as is.

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
TaylorBeebe: Same comment as above, this function is being restored to what it is in the upstream

apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2792,6 +2787,120 @@ GetDxeMemoryProtectionSettings (
...
+**/
+VOID
+DisableNullDetection (
+  VOID
+  )
+{
+  EFI_STATUS                       Status;
+  EFI_GCD_MEMORY_SPACE_DESCRIPTOR  Desc;
+
+  DEBUG ((DEBUG_INFO, "%a - Enter\n", __FUNCTION__));
```
os-d: nit: __func__ throughout

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2792,6 +2787,120 @@ GetDxeMemoryProtectionSettings (
...
+    return;
+  }
+
+  if ((Desc.Attributes & EFI_MEMORY_RP) == 0) {
+    DEBUG ((DEBUG_INFO, "%a - Nothing to do... Exit\n", __FUNCTION__));
+    return;
+  }
+
+  if ((Desc.Capabilities & EFI_MEMORY_RP) == 0) {
+    Status = CoreSetMemorySpaceCapabilities (
```
os-d: Why do we set the capability here? Does the set attribute call fail to remove this attribute if the capability is not set?

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2910,49 +3019,114 @@ InitializePageAttributesCallback (
...
+
+  // Initialize paging attributes
   InitializePageAttributesForMemoryProtectionPolicy ();
 
+  // Set all the guard pages
   HeapGuardCpuArchProtocolNotify ();
 
-  mPageAttributesInitialized = TRUE;
+  // Enable NULL pointer detection
+  EnableNullDetection ();
```
os-d: This will unconditionally enable null detection, right? Is this no longer back by gDxeMps HOB?

os-d (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Good catch! Will update

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: I wrote this to mirror the upstream DisableNullDetectionAtTheEndOfDxe() function but looking at the ConvertSpace() logic it doesn't seem like it's necessary to set the capabilities since this is clearing the attribute.

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: fixed

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
TaylorBeebe: Removed

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c

PR #841, opened 2024-05-06
CryptoPkg: Add Shared Crypto FDF include file

makubacki:
## Description

Adds a file that can be included by platforms to satisfy the build
rules needed to construct FFS files from the .efi files provided by
shared crypto.

It is also provided as a reference for the rules needed to use the
files in a platform.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- `QemuQ35Pkg` and `QemuSbsaPkg` build with the build rules directly in their
  FDF file removed and this file included.

## Integration Instructions

Users of Mu Shared Crypto may find this file useful to directly include in a
FDF file or as a reference for the build rules needed to use Shared Crypto.

This file can be included using `!include CryptoPkg/SharedCrypto.BuildRules.fdf.inc`
in the platform FDF file after the last `FV` section and before the other build rules
defined in the platform FDF file begin.

**IMPORTANT** - You should review these rules and ensure they meet your platform's
requirements. You should also ensure they do not conflict with any other bulid rules
that may already be defined in the platform FDF file. This file is not required to
use Shared Crypto. It is provided for convenience and reference.

### DIFF
--- a/CryptoPkg/SharedCrypto.BuildRules.fdf.inc
+++ b/CryptoPkg/SharedCrypto.BuildRules.fdf.inc
@@ -1,0 +1,74 @@
+## @file
+# Shared Crypto FDF include file.
+#
+# These rules are used to define how an FFS file is constructed.
+#
+# The Mu Shared Crypto project (https://github.com/microsoft/mu_crypto_release) provides binary images in the form
+# of .efi binaries. These are used for the PE32 section in FFS files. This FDF include file can be included in a FDF
+# file to define how FFS files should be created for Shared Crypto.
+#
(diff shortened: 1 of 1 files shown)


os-d (approved):

kuqin12:
This seems to be more than just a shared crypto inclusion. And the entries are applicable to the entire platform. Do we really want to provide this inc file, and in this package?

apop5 (approved):

makubacki:
Closing in favor of placing the FDF file in the release from mu_crypto_release in https://github.com/microsoft/mu_crypto_release/pull/85.

PR #861, opened 2024-05-24
MdeModulePkg\DxeCore: BS memory allocated from SMM Entry points is still RP

apop5:
## Description

With the completion of #768, memory is by default read protected.

If an SMM driver, in its entry point, allocates boot services memory, the existing code would not attempt to apply memory protections to the memory. This resulted in the original permissions continuing to exist. (RP)

Removing the check that prevents memory attributes to be applied to memory allocated from within a SMM driver.

- [X] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Verified that after this change, silicon vendor code was still able to allocate boot services memory and access the memory

## Integration Instructions
N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -43,7 +43,7 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 #include "DxeMain.h"
 #include "Mem/HeapGuard.h"
-#include "MemoryProtectionSupport.h"
+#include "MemoryProtectionSupport.h"        // MU_CHANGE
 
 //
 // Image type definitions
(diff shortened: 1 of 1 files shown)


os-d (approved):
Needs MU_CHANGE, unless this was originally a MU_CHANGE (in which case we can revert instead?)

apop5:
> Needs MU_CHANGE, unless this was originally a MU_CHANGE (in which case we can revert instead?)

The whole file is a MU specific file, it doesn't exist in edk2.

os-d:
> > Needs MU_CHANGE, unless this was originally a MU_CHANGE (in which case we can revert instead?)
> 
> The whole file is a MU specific file, it doesn't exist in edk2.

MemoryProtections.c does exist in edk2: https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c

MemoryProtectionSupport.c is a Mu specific file.

apop5:
> > > Needs MU_CHANGE, unless this was originally a MU_CHANGE (in which case we can revert instead?)
> > 
> > 
> > The whole file is a MU specific file, it doesn't exist in edk2.
> 
> MemoryProtections.c does exist in edk2: https://github.com/tianocore/edk2/blob/master/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
> 
> MemoryProtectionSupport.c is a Mu specific file.

I'll fix it.

makubacki (approved):

PR #881, opened 2024-05-31
Add RNG PPI Support

makubacki:
## Description

**MdePkg: Add Random Number Generator (RNG) PPI**

Adds a new PPI that serves the same purpose as EFI_RNG_PROTOCOL in
DXE. This PPI can be produced by a PEIM to provide a dynamic interface
to RNG services in PEI.

This PPI is called `EFI_RNG_PPI` because it shares the exact same
interface with `EFI_RNG_PROTOCOL` which is described in the UEFI
Speficiation.

---

**MdePkg: Add PeiRngLib**

Adds a new PEI library instance for RngLib that uses the RNG services
provided by the RNG PPI.

This library instance will add a DEPEX on gEfiRngPpiGuid on modules
it links against. It can be used to allow PEIMs to get RNG support
over a dynamic interface.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- MdePkg CI
- Verify the RNG PPI can be successfully produced and consumed

## Integration Instructions

If a platform needs to share RNG support across a dynamic interface
between PEIMs, the RNG PPI can be used. `PeiRngLib` provides a `RngLib`
instance that use the RNG PPI. It will include a dependency on `gEfiRngPpiGuid`.


MdePkg/Library/PeiRngLib/PeiRngLib.c
```diff
@@ -0,0 +1,227 @@
...
+  @retval EFI_UNSUPPORTED         Not supported.
+  @retval EFI_INVALID_PARAMETER   Invalid parameter.
+**/
+EFI_STATUS
+EFIAPI
+GetRngGuid (
+  GUID  *RngGuid
+  )
+{
+  return EFI_UNSUPPORTED;
```
os-d, line 229: Is this unsupported because we might be one of four algorithms/we don't want people to directly call the GUID?

os-d (approved):

MdePkg/Library/PeiRngLib/PeiRngLib.c
makubacki: It is mostly for symmetry in behavior to the same API in [`DxeRngLib`](https://github.com/tianocore/edk2/blob/b0930e3f4e6de1ce1c480bca687b44875e071f74/MdePkg/Library/DxeRngLib/DxeRngLib.c#L224C1-L224C12). Like there, we don't know exactly which algorithm might be used. I can add a similar comment as well.

Regarding the two instances (PEI and DXE), I considered moving common code to its own file with two INFs, but changing the `DxeRngLib` would be breaking, the majority of code is in the phase-specific function, and I think the instances might eventually differ in implementation anyway, so I just created a new instance.

makubacki: I added a comment. Please feel free to reopen and let me know if you have any issue with the change pushed.

MdePkg/Include/Ppi/Rng.h
```diff
@@ -0,0 +1,28 @@
+/** @file
+  The Random Number Generator (RNG) PPI is used to provide random bits for use
+  in PEIMs, or entropy for seeding other random number generators.
+
+  Copyright (c) Microsoft Corporation. All rights reserved.
```
kuqin12:
```suggestion
  Copyright (c) Microsoft Corporation.
```

MdePkg/Library/PeiRngLib/PeiRngLib.c
```diff
@@ -0,0 +1,230 @@
+/** @file
+  RNG library instance that uses the Random Number Generator (RNG) PPI to provide
+  random numbers.
+
+  Copyright (C) Microsoft Corporation. All rights reserved.
```
kuqin12:
```suggestion
  Copyright (c) Microsoft Corporation.
```

MdePkg/Library/PeiRngLib/PeiRngLib.inf
```diff
@@ -0,0 +1,43 @@
+## @file
+# PPI-based Instance of RNG (Random Number Generator) Library.
+#
+# This library instance requires a RNG PPI to be produced so that the module may use
+# it for RNG operations. A RNG PPI DEPEX will be placed on the module.
+#
+# Copyright (C) Microsoft Corporation. All rights reserved.
```
kuqin12:
```suggestion
# Copyright (c) Microsoft Corporation.
```

MdePkg/Library/PeiRngLib/PeiRngLib.inf
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# MU_CHANGE: WHOLE FILE
+##
+
+[Defines]
+  INF_VERSION           = 0x0001001B
+  BASE_NAME             = PeiRngLib
+  MODULE_UNI_FILE       = RngLib.uni
```
kuqin12: Do we have this file?

kuqin12 (approved):

MdePkg/Library/PeiRngLib/PeiRngLib.inf
makubacki: Good catch. I added it.

PR #882, opened 2024-06-03
MuCodeQlQueries.qls: Pin to the 0.9.12 codeq/cpp-queries pack

makubacki:
## Description

The `codeql/cpp-queries` pack used in MuCodeQlQueries.qls was versioned
0.9.12 for the CodeQL CLI v2.17.3 release currently used.

https://github.com/github/codeql/blob/codeql-cli/v2.17.3/cpp/ql/src/qlpack.yml

This change pins that pack version to prevent the CodeQL CLI and
pack from getting out of sync until explicitly updated.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified the CodeQL query pack version listed is pulled.

## Integration Instructions

- N/A - No change to queries used. Should prevent breaks in the future where
  the latest queries are no longer compatible with the current CodeQL CLI used.

### DIFF
--- a/BaseTools/Plugin/CodeQL/MuCodeQlQueries.qls
+++ b/BaseTools/Plugin/CodeQL/MuCodeQlQueries.qls
@@ -2,7 +2,7 @@
 - description: Project Mu UEFI (C++) queries
 
 - queries: '.'
-  from: codeql/cpp-queries
+  from: codeql/cpp-queries@0.9.12
 
 ##########################################################################################
 # "Core" Queries - Part of Core SDL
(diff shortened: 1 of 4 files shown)


makubacki:
Please review the 2302 version of this PR as well - https://github.com/microsoft/mu_basecore/pull/883.

apop5 (approved):

os-d (approved):

PR #891, opened 2024-06-05
[CHERRY-PICK] [Release/202311] UnitTestFrameworkPkg: Fix Google Test components with multiple files

Flickdm:
# Preface

REF: https://bugzilla.tianocore.org/show_bug.cgi?id=4610

Google Test hides test registration in global constructors on global objects. Global constructors are traditionally implemented by placing references to the global constructor's symbol in special sections (traditionally named .ctors or .init_array). These sections are not explicitly referenced by the linker, and libc only looks at special start and end symbols (and calls them).

This works fine if you're linking a program manually using

    gcc a.o b.o c.o -o test_suite

but fails miserably when using static libraries (such as what EDK2 does), because traditional static archive symbol resolution rules don't allow for object files to be pulled in to the link if there isn't an undefined symbol reference to that .o elsewhere.

Fix it by passing --whole-archive (GCC) and /WHOLEARCHIVE (MSVC). These options force the linker to pull in the entire static library, thus including previously-unreferenced constructors and making sure multi-file gtest EDK2 components work.

Cc: Michael D Kinney <michael.d.kinney@intel.com>
Cc: Michael Kubacki <mikuback@linux.microsoft.com>
Cc: Sean Brogan <sean.brogan@microsoft.com>

Reviewed-by: Cc: Michael D Kinney <michael.d.kinney@intel.com>

## Description

This corrects an issue with MSVC where unit tests are not being linked correctly - thus creating unit tests with zero tests

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
  - This will cause unit tests to actually run with MSVC
- [ ] Includes documentation?

## How This Was Tested

Local builds with Host Based Unit Tests

## Integration Instructions

N/A

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -43,8 +43,8 @@
   #
   # MSFT
   #
-  MSFT:*_*_*_CC_FLAGS               = /EHs -D GOOGLETEST_HOST_UNIT_TEST_BUILD=1
-  MSFT:*_*_*_DLINK_FLAGS            == /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /IGNORE:4001 /NOLOGO /SUBSYSTEM:CONSOLE /DEBUG /STACK:0x40000,0x40000 /NODEFAULTLIB:libcmt.lib libcmtd.lib
+  MSFT:*_*_*_CC_FLAGS               = /EHsc
+  MSFT:*_*_*_DLINK_FLAGS            == /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /IGNORE:4001 /NOLOGO /SUBSYSTEM:CONSOLE /DEBUG /STACK:0x40000,0x40000 /WHOLEARCHIVE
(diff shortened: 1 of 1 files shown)


UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
```diff
@@ -43,8 +43,8 @@
   #
   # MSFT
   #
-  MSFT:*_*_*_CC_FLAGS               = /EHs -D GOOGLETEST_HOST_UNIT_TEST_BUILD=1
```
VivianNK, line 46: Why remove the  `GOOGLETEST_HOST_UNIT_TEST_BUILD=1` ? I thought that was done for defining the static macro 
https://github.com/microsoft/mu_basecore/commit/6e76157d919202791a3d25cf9e606a72ee1af6a3

apop5 (approved):

Javagedes (approved):

UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
VivianNK: @os-d

os-d: Yeah, we want to keep this.

Flickdm: I went ahead and re-cherrypicked your commit because this commit from EDK2 wants to delete it

VivianNK (approved):

PR #893, opened 2024-06-06
[CHERRY-PICK] [Release/202302] UnitTestFrameworkPkg: Fix Google Test components with multiple files

Flickdm:
# Preface
REF: https://github.com/microsoft/mu_basecore/pull/891
  - Dropping GOOGLETEST_HOST_UNIT_BUILD option as release/202302 does
    not have any expectation to support it.

REF: https://bugzilla.tianocore.org/show_bug.cgi?id=4610

Google Test hides test registration in global constructors on global objects. Global constructors are traditionally implemented by placing references to the global constructor's symbol in special sections (traditionally named .ctors or .init_array). These sections are not explicitly referenced by the linker, and libc only looks at special start and end symbols (and calls them).

This works fine if you're linking a program manually using

    gcc a.o b.o c.o -o test_suite

but fails miserably when using static libraries (such as what EDK2 does), because traditional static archive symbol resolution rules don't allow for object files to be pulled in to the link if there isn't an undefined symbol reference to that .o elsewhere.

Fix it by passing --whole-archive (GCC) and /WHOLEARCHIVE (MSVC). These options force the linker to pull in the entire static library, thus including previously-unreferenced constructors and making sure multi-file gtest EDK2 components work.

Cc: Michael D Kinney <michael.d.kinney@intel.com>
Cc: Michael Kubacki <mikuback@linux.microsoft.com>
Cc: Sean Brogan <sean.brogan@microsoft.com>

Reviewed-by: Cc: Michael D Kinney <michael.d.kinney@intel.com>

## Description

This corrects an issue with MSVC where unit tests are not being linked correctly - thus creating unit tests with zero tests

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
  - This will cause unit tests to actually run with MSVC
- [ ] Includes documentation?

## How This Was Tested

Local builds with Host Based Unit Tests

## Integration Instructions

N/A

### DIFF
--- a/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
+++ b/UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
@@ -42,7 +42,7 @@
   # MSFT
   #
   MSFT:*_*_*_CC_FLAGS               = /EHsc
-  MSFT:*_*_*_DLINK_FLAGS            == /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /IGNORE:4001 /NOLOGO /SUBSYSTEM:CONSOLE /DEBUG /STACK:0x40000,0x40000 /NODEFAULTLIB:libcmt.lib libcmtd.lib
+  MSFT:*_*_*_DLINK_FLAGS            == /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /IGNORE:4001 /NOLOGO /SUBSYSTEM:CONSOLE /DEBUG /STACK:0x40000,0x40000 /WHOLEARCHIVE
   MSFT:*_*_IA32_DLINK_FLAGS         = /MACHINE:I386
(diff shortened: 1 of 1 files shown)


UnitTestFrameworkPkg/UnitTestFrameworkPkgHost.dsc.inc
```diff
@@ -41,8 +41,10 @@
   #
   # MSFT
   #
-  MSFT:*_*_*_CC_FLAGS               = /EHsc
-  MSFT:*_*_*_DLINK_FLAGS            == /out:"$(BIN_DIR)\$(MODULE_NAME_GUID).exe" /pdb:"$(BIN_DIR)\$(MODULE_NAME_GUID).pdb" /IGNORE:4001 /NOLOGO /SUBSYSTEM:CONSOLE /DEBUG /STACK:0x40000,0x40000 /NODEFAULTLIB:libcmt.lib libcmtd.lib
+  # MU_CHANGE [BEGIN] - Allow GoogleTest to Test STATIC Functions
+  MSFT:*_*_*_CC_FLAGS               = /EHsc -D GOOGLETEST_HOST_UNIT_TEST_BUILD=1
```
os-d: I think this doesn't exist in 202302. This won't break anything without the Base.h change, but if you are anticipating needing to test static functions, you need that change as well.

os-d (approved):

VivianNK (approved):

PR #903, opened 2024-06-11
[Rebase & FF] 202405: BaseTools/WinRcPath: Improve Performance.

apop5:
## Description

WinRcPath generally takes about 2 seconds to run, due to calling multiple .bat files behind the scenes. This change reduces this time to ~0 seconds due to the following changes:

1. It will attempt to load the path from the cache, which is located a $(WORKSPACE)/Conf/.rc_path. If the loading is a success and the rc_path still exists, it will use it.

2. If the cache did not exist, or the path provided by the cache does not exist, it will find the rc path via the .bat files. If that succeeds, it will write the path to the cache.

## Cherry-Pick and squash the following commits:
[fe9fc525d9](https://github.com/microsoft/mu_basecore/commit/fe9fc525d9)

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. Confirmed the plugin will create a cache file on first run and use itin later runs
2. Confirmed updating the path in the cache to a wrong value will invoke the typical means to find the cache and save the correct path into the cache.

## Integration Instructions

N/A


BaseTools/Plugin/WindowsResourceCompiler/WinRcPath.py
makubacki, line 1: To have history be more valuable, we should preserve the original author of commits unless it is much more complicated to do so. In this case, the commit author should be Joey with a co-author of Matt Carlson since it is essentially a squash of https://github.com/microsoft/mu_basecore/commit/5c597f57bfc59b73a95730af3d4782ddf91eac4a and https://github.com/microsoft/mu_basecore/commit/fe9fc525d998cc59210fec7ae487da958ddaad8c.

Then, we will have primary author Joey, co-author Matt, committer Aaron making the change much easier to trace and track to original authors for clarifying info.

makubacki: The author currently shown is:

```
commit abbb272a32d7846581d3bfb69bce33c2a34ccd53 (apop5/personal/apop5/winrcimprovement)
Author: Aaron <105021049+apop5@users.noreply.github.com>
Date:   Tue Jun 11 10:13:59 2024 -0700
```

I recommend changing the author to a full name and email in applicable commits in the future, so it transfers better to edk2 and is more consistent for author searches in the codebase.

makubacki (approved):

Javagedes:
@apop5 you accidently added a step 3 in your PR description

Javagedes (approved):

PR #946, opened 2024-06-25
Add extra RestoreTpl() call in DiskIo and StartImage() to maintain TPL raise/restore symmetry [Rebase & FF]

kenlautner:
## Description

Adds a call to RestoreTpl() in DiskIo2ReadWriteDisk(). While the current implementation does not technically violate spec on raise/restore TPL, this extra call ensures symmetry between RaiseTpl and RestoreTpl calls, which makes analysis of TPL correctness simpler and permits certain non-standard TPL usages that some platforms require.

Comments out a redundant call to RestoreTpl(). While this does not technically violate spec on raise/restore TPL, TPL should already be at the specified level. This extra call introduces an asymmetry between RaiseTpl and RestoreTpl calls, which makes analysis of TPL correctness more difficult and hampers certain non-standard TPL usages that some platforms require.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Booted and observed no un-intended side effects w.r.t. TPL with this modification. Added test instrumentation to verify that TPL is always already at the desired state prior to this call being executed.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Image/Image.c
+++ b/MdeModulePkg/Core/Dxe/Image/Image.c
@@ -1725,7 +1725,10 @@ CoreStartImage (
   // Image has completed.  Verify the tpl is the same
   //
   ASSERT (Image->Tpl == gEfiCurrentTpl);
-  CoreRestoreTpl (Image->Tpl);
+  // MU_CHANGE - reduce TPL restore
+  if (Image->Tpl != gEfiCurrentTpl) {
+    CoreRestoreTpl (Image->Tpl);
+  } // MU_CHANGE - reduce TPL restore
 
   CoreFreePool (Image->JumpBuffer);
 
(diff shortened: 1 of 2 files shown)


MdeModulePkg/Core/Dxe/Image/Image.c
```diff
@@ -1725,7 +1725,10 @@ CoreStartImage (
   // Image has completed.  Verify the tpl is the same
   //
   ASSERT (Image->Tpl == gEfiCurrentTpl);
-  CoreRestoreTpl (Image->Tpl);
+  if (Image->Tpl != gEfiCurrentTpl) {
+    // MU_CHANGE - reduce TPL restore
```
makubacki: The `MU_CHANGE` tag here should be outside the new if statement introduced.

kenlautner: Fixed.

makubacki (approved):

os-d (approved):

PR #954, opened 2024-06-26
Improve log messages for DependencyCheck, UncrustifyCheck, and SpellCheck

kenlautner:
## Description

Improve CI log output with more actionable messages, and downgraded some errors/warnings that should not be treated as such.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Checked CI plugin results.

## Integration Instructions

N/A

### DIFF
--- a/.pytool/Plugin/DependencyCheck/DependencyCheck.py
+++ b/.pytool/Plugin/DependencyCheck/DependencyCheck.py
@@ -108,11 +108,9 @@ class DependencyCheck(ICiBuildPlugin):
                     if mod_specific_key in pkgconfig and p in pkgconfig[mod_specific_key]:
                         continue
 
-                    # MU_CHANGE [BEGIN] - Provide more DependencyCheck error info
                     logging.error(f"Dependency Check: {file} depends on pkg {p} but pkg is not listed in AcceptableDependencies")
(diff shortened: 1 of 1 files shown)


.pytool/Plugin/DependencyCheck/DependencyCheck.py
```diff
@@ -94,7 +94,7 @@ def RunBuildPlugin(self, packagename, Edk2pathObj, pkgconfig, environment, PLM,
         for file in INFFiles:
             ip = InfParser()
             logging.debug("Parsing " + file)
-            ip.SetEdk2Path(Edk2pathObj).ParseFile(file)     # MU_CHANGE
+            ip.SetBaseAbsPath(Edk2pathObj.WorkspacePath).SetPackagePaths(Edk2pathObj.PackagePathList).ParseFile(file)
```
os-d: This would seem to already be in 2405, looking at the original commit. It also does not do what the PR message says. I suspect this can be dropped, @makubacki?

kenlautner: Looking at the history you're correct that this change isn't necessary.  I'll fix it now.

kenlautner:
Dropping this PR as all the changes basically went in when CI was added.

PR #958, opened 2024-06-26
Add support for HTTP error 429

kenlautner:
## Description
Add support for HTTP error 429.  This includes two commits from 202311 that are listed below:

30695f9
f237312

Below are descriptions for the commits:

Add HTTP Error code 429.  This is a temporary fix pending the outcome of BZ 1685

---

HTTP status codes have to be explicitly specified in DxeHttpLib in order to be supported. Intune may return 429 to throttle requests.

---

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Compiled.

## Integration Instructions

N/A


makubacki (approved):

MdePkg/Include/Protocol/Http.h
makubacki, line 1: You can squash the two commits locally, push the single commit, and then Rb&Ff that to commit the same commit info.

kenlautner: Wouldn't we want them separate for upstreaming purposes?

makubacki: For something this simple, I think we can just do that when they're upstreamed and keep our history simpler. It should make it more obvious both need to go and edk2 package split rules would dictate to break it out into two commits at that time.

os-d (approved):

PR #959, opened 2024-06-26
UefiBootManagerLib: Update assert condition in BmFindBootOptionInVariable

kenlautner:
## Description

Currently the following assertion is made in
`BmFindBootOptionInVariable()`:

```c
  if (OptionNumber == LoadOptionNumberUnassigned) {
    BootOptions = EfiBootManagerGetLoadOptions (&BootOptionCount, LoadOptionTypeBoot);

    if (BootOptions == NULL) {
      ASSERT (BootOptions != NULL);
      return LoadOptionNumberUnassigned;
    }

    Index = EfiBootManagerFindLoadOption (OptionToFind, BootOptions, BootOptionCount);
```

The reason behind this is to prevent passing a null pointer to `EfiBootManagerFindLoadOption()`.

However, `EfiBootManagerFindLoadOption()` accepts a null pointer argument to be passed as the second argument ('Array') as long as the `Count` argument is `0`.

In that case `LoadOptionNumberUnassigned` will still be returned from the function but an assert is not necessary.

This change updates the condition to only assert if the pointer is null and the boot option count is non-zero.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified build and boot on Qemu35Pkg.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
+++ b/MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c
@@ -147,6 +147,14 @@ BmFindBootOptionInVariable (
   if (OptionNumber == LoadOptionNumberUnassigned) {
     BootOptions = EfiBootManagerGetLoadOptions (&BootOptionCount, LoadOptionTypeBoot);
 
+    // MU_CHANGE [BEGIN] - Only assert if BootOptionCount is non-zero
+    if ((BootOptions == NULL) && (BootOptionCount > 0)) {
+      ASSERT (BootOptions != NULL);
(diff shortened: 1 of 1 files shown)


makubacki (approved):

makubacki:
I didn't add `MU_CHANGE` tags to the original change so I pushed that update to your PR branch.

os-d (approved):

PR #994, opened 2024-06-27
MdePkg/Include/Library/BaseLib: Add CRC16 CCITT False Implementation. [Rebase & FF]

apop5:
## Description

This change is added to incorporate basic implementation for CRC16-CCITT-FALSE algorithm.

This function is useful for providing CRC16 value used in other data structures that requires CRC16 value that complies with JEDEC SPD requirements, i.e. BDAT table.

The lookup table is inherited from `https://crccalc.com/` and the result values are also compared against this site.

## Cherry-Pick the following commits:
[dcdd08f1f0](https://github.com/microsoft/mu_basecore/commit/dcdd08f1f0)

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

The function output is compared and matches with the results of 3rd
party online CRC calculators.

## Integration Instructions

N/A


makubacki (approved):

MdePkg/Library/BaseLib/CheckSum.c
```diff
@@ -762,3 +762,74 @@ CalculateCrc32c (
 
   return ~Crc;
 }
+
+// MU_CHANGE [START] - Add CRC16-CCITT-FALSE implementation
```
kuqin12: Since we are here, can we please add the source of this table into the comment? I put it in the commit message originally, but I think in the code would be better reflecting the state of the implementation:

The lookup table is inherited from `[https://crccalc.com/`](https://crccalc.com/%60)

kuqin12 (approved):

MdePkg/Library/BaseLib/CheckSum.c
apop5: Yes, I added it, and it will be in the next push.

PR #1010, opened 2024-06-28
TCBZ2442: Change S3 PCD default

kenlautner:
## Description

Change S3 PCD default to FALSE.

- [x] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Passed CI.

## Integration Instructions

If a platform was depending on PcdAcpiS3Enable being TRUE they'll need to update the value in their platform dsc file.

### DIFF
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -2160,7 +2160,8 @@
   #   TRUE  - ACPI S3 will be enabled.<BR>
   #   FALSE - ACPI S3 will be disabled.<BR>
   # @Prompt ACPI S3 Enable.
-  gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiS3Enable|TRUE|BOOLEAN|0x01100000
+  # MU_CHANGE - Default should be FALSE. We don't support S3. TCBZ2442
+  gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiS3Enable|FALSE|BOOLEAN|0x01100000
 
   ## Specify memory size for boot script executor stack usage in S3 phase.
   #  The default size 32K. When changing the value make sure the memory size is large enough


MdeModulePkg/MdeModulePkg.dec
```diff
@@ -2155,7 +2155,8 @@
   #   TRUE  - ACPI S3 will be enabled.<BR>
   #   FALSE - ACPI S3 will be disabled.<BR>
   # @Prompt ACPI S3 Enable.
-  gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiS3Enable|TRUE|BOOLEAN|0x01100000
+  # MU_CHANGE - Default should be FALSE. We don't support S3. TCBZ2442
+  gEfiMdeModulePkgTokenSpaceGuid.PcdAcpiS3Enable|FALSE|BOOLEAN|0x01100000
```
makubacki, line 2164: This does match modern platform behavior, but I think it's debatable if it is really needed as an override in the DEC since it is so easy to change at the platform level. I'm fine with moving it forward though.

makubacki (approved):

MdeModulePkg/MdeModulePkg.dec
kenlautner: I think we should keep the change because all platforms consuming project MU could be relying on the new default value.

makubacki: Yes, I'm speaking long term with the understanding that moving to a new release branch is the best time to make breaking changes. That same problem is inverted for the entire edk2 community to take this change, so they're unlikely to, making it an unreconcilable override. However, I don't think the debt it brings justifies a change here right now.

os-d: I agree, I think we should drop this and add it to the 2405 release notes in the breaking change section.

PR #1018, opened 2024-07-02
Removed openssl and mbedtls submodules

kenlautner:
## Description

In the PR that got CryptoPkg up to parity with the release/202311 branch the openssl and mbedtls submodules were accidentally left in the codebase.  This PR fixes that by removing them.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "CryptoPkg/Library/OpensslLib/openssl"]
-	path = CryptoPkg/Library/OpensslLib/openssl
-	url = https://github.com/openssl/openssl
 [submodule "UnitTestFrameworkPkg/Library/CmockaLib/cmocka"]
 	path = UnitTestFrameworkPkg/Library/CmockaLib/cmocka
 	url = https://github.com/tianocore/edk2-cmocka.git
@@ -26,6 +23,3 @@
 [submodule "MdePkg/Library/MipiSysTLib/mipisyst"]
 	path = MdePkg/Library/MipiSysTLib/mipisyst
 	url = https://github.com/MIPI-Alliance/public-mipi-sys-t.git
-[submodule "CryptoPkg/Library/MbedTlsLib/mbedtls"]
-	path = CryptoPkg/Library/MbedTlsLib/mbedtls
-	url = https://github.com/ARMmbed/mbedtls
--- a/.pytool/CISettings.py
+++ b/.pytool/CISettings.py
@@ -174,8 +174,6 @@ class Settings(CiBuildSettingsManager, UpdateSettingsManager, SetupSettingsManag
         '''
         rs = []
         rs.append(RequiredSubmodule(
-            "CryptoPkg/Library/OpensslLib/openssl", False))
-        rs.append(RequiredSubmodule(
             "UnitTestFrameworkPkg/Library/CmockaLib/cmocka", False))
         rs.append(RequiredSubmodule(
             "UnitTestFrameworkPkg/Library/GoogleTestLib/googletest", False))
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -1,31 +1,28 @@
...
+	url = https://github.com/Zeex/subhook.git
+[submodule "MdePkg/Library/BaseFdtLib/libfdt"]
+	path = MdePkg/Library/BaseFdtLib/libfdt
+	url = https://github.com/devicetree-org/pylibfdt.git
+[submodule "MdePkg/Library/MipiSysTLib/mipisyst"]
+	path = MdePkg/Library/MipiSysTLib/mipisyst
+	url = https://github.com/MIPI-Alliance/public-mipi-sys-t.git
+[submodule "CryptoPkg/Library/MbedTlsLib/mbedtls"]
+	path = CryptoPkg/Library/MbedTlsLib/mbedtls
+	url = https://github.com/ARMmbed/mbedtls
```
makubacki: mbedtls can also be removed since the idea is that crypto implementation will only be shared through binaries in Mu Crypto Release.

.gitmodules
```diff
@@ -1,31 +1,28 @@
...
-[submodule "MdePkg/Library/BaseFdtLib/libfdt"]
-	path = MdePkg/Library/BaseFdtLib/libfdt
-	url = https://github.com/devicetree-org/pylibfdt.git
-[submodule "MdePkg/Library/MipiSysTLib/mipisyst"]
-	path = MdePkg/Library/MipiSysTLib/mipisyst
-	url = https://github.com/MIPI-Alliance/public-mipi-sys-t.git
-[submodule "CryptoPkg/Library/MbedTlsLib/mbedtls"]
-	path = CryptoPkg/Library/MbedTlsLib/mbedtls
-	url = https://github.com/ARMmbed/mbedtls
+[submodule "UnitTestFrameworkPkg/Library/CmockaLib/cmocka"]
```
makubacki: It looks like the line endings are changing in this file. It is allowed to be LF since it is a git file not in a package.

.pytool/CISettings.py
```diff
@@ -173,8 +173,6 @@ def GetRequiredSubmodules(self):
         If no RequiredSubmodules return an empty iterable
         '''
         rs = []
-        rs.append(RequiredSubmodule(
-            "CryptoPkg/Library/OpensslLib/openssl", False))
```
os-d, line 177: I think this can be squashed into the first commit.

os-d (approved):
Approved, with Michael's comments incorporated

.gitmodules
kenlautner: True.  I didn't check if all the corresponding MbedTLS code was already removed and if it wasn't I thought it removing it would be with that commit.  I see now that it has already been removed so it makes sense to get rid of the submodule here as well.

.gitmodules
kenlautner: This was automatically done when removing the submodule and I didn't notice.  It's fixed now.

.pytool/CISettings.py
kenlautner: Done.

.gitmodules
makubacki: Just a reminder to please squash this one or update the commit message to also mention removal of the mbedtls submodule.

makubacki (approved):

PR #1030, opened 2024-07-05
MdeModulePkg: Compatibility Mode: Only Remap System Memory Regions

os-d:
## Description

When we enter memory protections compatibility mode, we attempt to disable null protection and remap 0 - 0xA0000 as RWX. This was done for x86 systems with broken shim/grubs on Linux that would attempt to use those regions. This resolved that issue and we could boot non-memory protection safe Linux images on x86 HW. However, this approach did not take into account systems that do not have that range marked as system memory, for example ARM64 systems do not have this requirement. As such, this would inappropriately map these regions as RWX when they were not system memory.

This patch updates the remapping to only remap and disable null protection if these ranges are marked as system memory, otherwise it will leave them alone.

- [x] Impacts functionality?
- [x] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an ARM64 platform that does not have 0 - 0xA0000 as system memory, as well as an X86 system that does have that range as system memory, booting a Linux image on both that forces us to enter compatibility mode.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
@@ -5,6 +5,7 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 **/
 
+#include <Library/SafeIntLib.h>
 #include "MemoryProtectionSupport.h"
 
 IMAGE_PROPERTIES_PRIVATE_DATA  mImagePropertiesPrivate = {
@@ -65,6 +66,8 @@ BOOLEAN                        mGcdSyncComplete                = FALSE;
 
 // TRUE if A is a bitwise subset of B
 #define CHECK_SUBSET(A, B)  ((A | B) == B)
+
+#define LEGACY_BIOS_WB_LENGTH  0xA0000
 
 /**
   Return the section alignment requirement for the PE image section type.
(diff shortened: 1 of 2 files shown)


apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
```diff
@@ -2941,21 +2954,80 @@ MapLegacyBiosMemoryRWX (
...
+        "%a - Failed to get memory space descriptor for address 0x%llx! Status = %r\n",
+        __func__,
+        Start,
+        Status
+        ));
+      return;
+    }
+
+    if (Desc.BaseAddress + Desc.Length > LEGACY_BIOS_WB_LENGTH) {
+      Length = LEGACY_BIOS_WB_LENGTH - Desc.BaseAddress;
```
VivianNK: Can you explain how this Length variable differs and why it's not used on line 3003 or 3029?

VivianNK (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtectionSupport.c
os-d: `Length` here is the range of memory that we are marking as RWX. `Start` is being used to look through the different memory descriptors (which may be one for multiple and may extend past the end of the range we need to mark RWX). So in 3003 and 3029 we are navigating the descriptors, not the length we are changing attributes for.

makubacki (approved):

spbrogan (dismissed):
reviewing but i have some concerns about under/overflow and checking consistency between different fields

os-d:
@spbrogan, refactored to use SafeIntLib and to only use my own fields as the basis for further checks.

PR #1034, opened 2024-07-11
Update BaseCryptLib tests to reference the PCDs before running

kenlautner:
## Description

The BaseCryptLibUnitTestApp tests the linked BaseCryptLib instance's crypto to make sure all functions are performing as expected.  With the move to the Crypto binary and the BaseCryptLibOnProtocol instances we disable certain crypto functionality on purpose which causes the test to fail (and also the BaseCryptLibOnProtocol lib to assert).  The changes made here use the already existing crypto PCDs to check if the tested cryptography is enabled with the current Crypto binary and if not to skip the test.  This will allow the test to show if the enabled crypto is working correctly instead of failing for crypto we don't care about.

- [ ] Impacts functionality?
- [ ] Impacts security?
    flow, or firmware?
    validation improvement, ...
- [ ] Breaking change?
    in build or boot behavior?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Qemu and intel physical platforms with various crypto binary layouts.  The relevant tests pass and disabled crypto skips their tests.  Furthermore when the PCDs are configured to run tests for crypto we don't support with the selected crypto binary, the test fails as expected.

## Integration Instructions

N/A.  Using the crypto binaries should automatically configure the correct PCDs and BaseCryptLib library for the test to work correctly.

### DIFF
--- a/CryptoPkg/Test/UnitTest/Library/BaseCryptLib/AeadAesGcmTests.c
+++ b/CryptoPkg/Test/UnitTest/Library/BaseCryptLib/AeadAesGcmTests.c
@@ -52,6 +52,10 @@ TestVerifyAeadAesGcm (
   UINT8    OutTag[1024];
   UINTN    OutTagSize;
 
+  if (!PcdGetBool (PcdCryptoServiceAeadAesGcmEncrypt) || !PcdGetBool (PcdCryptoServiceAeadAesGcmDecrypt)) {
+    return UNIT_TEST_ERROR_PREREQUISITE_NOT_MET;
+  }
+
   OutBufferSize = sizeof (OutBuffer);
   OutTagSize    = sizeof (gcm_tag);
   ZeroMem (OutBuffer, sizeof (OutBuffer));
--- a/CryptoPkg/Test/UnitTest/Library/BaseCryptLib/AeadAesGcmTests.c
+++ b/CryptoPkg/Test/UnitTest/Library/BaseCryptLib/AeadAesGcmTests.c
@@ -52,9 +52,12 @@ TestVerifyAeadAesGcm (
   UINT8    OutTag[1024];
   UINTN    OutTagSize;
 
+  // MU_CHANGE [START]
   if (!PcdGetBool (PcdCryptoServiceAeadAesGcmEncrypt) || !PcdGetBool (PcdCryptoServiceAeadAesGcmDecrypt)) {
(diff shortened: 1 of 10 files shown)


CryptoPkg/Test/UnitTest/Library/BaseCryptLib/AeadAesGcmTests.c
```diff
@@ -52,6 +52,10 @@ TestVerifyAeadAesGcm (
   UINT8    OutTag[1024];
   UINTN    OutTagSize;
 
+  if (!PcdGetBool (PcdCryptoServiceAeadAesGcmEncrypt) || !PcdGetBool (PcdCryptoServiceAeadAesGcmDecrypt)) {
```
VivianNK, line 56: Do these not need MU_CHANGE tags?

kenlautner: You're right that they should be included.  I've updated the PR.

CryptoPkg/Test/UnitTest/Library/BaseCryptLib/BaseCryptLibUnitTestApp.inf
```diff
@@ -60,8 +60,14 @@
   gEfiCryptoPkgTokenSpaceGuid.PcdCryptoServiceHmacSha256SetKey            ## CONSUMES
   gEfiCryptoPkgTokenSpaceGuid.PcdCryptoServiceHmacSha256Update            ## CONSUMES
   gEfiCryptoPkgTokenSpaceGuid.PcdCryptoServiceHmacSha256Final             ## CONSUMES
+  gEfiCryptoPkgTokenSpaceGuid.PcdCryptoServiceHmacSha384New               ## CONSUMES
```
VivianNK: nit: mu change tags missing from this file.

VivianNK (approved):

CryptoPkg/Test/UnitTest/Library/BaseCryptLib/BaseCryptLibUnitTestApp.inf
kenlautner: Updated.

kuqin12 (approved):

PR #1045, opened 2024-07-18
Fix buffer overrun in FTWDxe driver

sureshkumarpMSFT:
## Description

There is buffer overrun in FtwGetLastWriteHeader function when the buffer examined is malformed. In this case the code is trying to access beyond FtwWorkSpaceSize in any of the code prior to (Offset >= FtwWorkSpaceSize) check

Changed the code to anticipate this early and terminate the loop earlier when offset + EFI_FAULT_TOLERANT_WRITE_HEADER is greater than FtwWorkSpaceSize

- [x ] Impacts functionality?
  -  Terminates the loops without doing a buffer overrun
- [ x] Impacts security?
    flow, or firmware?
  - Terminates the loops without doing a buffer overrun

- [ ] Breaking change?
    in build or boot behavior?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested this in platform which has a SPI chip with improper firmware blocks.  Without this fix boot crashed and with this change system continued to boot.

## Integration Instructions
N/A


os-d (approved):
Looks like the PR template got munged here, can you use our standard PR template?

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
```diff
@@ -815,7 +815,10 @@ FtwGetLastWriteHeader (
     //
     // If Offset exceed the FTW work space boudary, return error.
     //
-    if (Offset >= FtwWorkSpaceSize) {
+     // MU_CHANGE [BEGIN] - Prevent buffer overrun
+    //if (Offset >= FtwWorkSpaceSize) {
+    if ((Offset + sizeof (EFI_FAULT_TOLERANT_WRITE_HEADER)) >= FtwWorkSpaceSize)
```
kuqin12: Is it really illegitimate for the "=" condition? Can it not just be an empty header?

VivianNK (approved):

apop5 (approved):

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: Hi @kuqin12 

In overall perspective, an empty header at the end of the block is not useful at all.

Within the scope of function, yes, I agree with you on the fact that an empty header sitting at end of the block should be considered `legitimate` and should return success and it is up to the caller to make sure if there is a space to write the record data.

let me wait for a while and check if other folks have any other comments. If others have no comments, then I will change the condition to   `_if ((Offset + sizeof (EFI_FAULT_TOLERANT_WRITE_HEADER)) > FtwWorkSpaceSize)_`

cc: @os-d  @apop5 @VivianNK @makubacki

VivianNK: Without including the `EFI_FAULT_TOLERANT_WRITE_HEADER` the condition used to be `>=`, was this inaccurate?

os-d: @sureshkumarpMSFT, I'm good to go with Kun's suggestion. @VivianNK, yeah, the assumption here is that this was incorrect.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki, line 1: You mentioned "reverse polarity" in the PR description. I just want to mention that is invalid for this driver as it assumes [`1` is the erase polarity](https://github.com/microsoft/mu_basecore/blob/release/202311/MdeModulePkg/Universal/FaultTolerantWriteDxe/FaultTolerantWrite.h#L35). So, if inverse polarity actually is being used, additional work is needed.

The firmware volume header specifies the polarity (as set in the FDF file) and this driver depends on the firmware volume block protocol, so technically it could verify the polarity at entry. Nothing to fix in the PR just a comment.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
```diff
@@ -815,7 +815,10 @@ FtwGetLastWriteHeader (
     //
     // If Offset exceed the FTW work space boudary, return error.
     //
-    if (Offset >= FtwWorkSpaceSize) {
+    // MU_CHANGE [BEGIN] - Prevent buffer overrun
```
makubacki, line 823: If this function is receiving `FtwWorkSpaceHeader` as input, I think it should not make any assumptions about previous validation of the buffer contents and check the [`Signature` at the beginning of the structure](https://github.com/microsoft/mu_basecore/blob/release/202311/MdeModulePkg/Include/Guid/SystemNvDataGuid.h#L37), that's why it is there.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: Hi @makubacki 

Yes,  that is true. To fix the reverse polarity issue we need to do more work.

By default it assumes that polarity is 1 https://github.com/microsoft/mu_basecore/blob/a9612223c13b05682855801c2593c44ff347c0f4/MdeModulePkg/Include/Guid/SystemNvDataGuid.h#L64

if the default case it would exit as soon as `FtwHeader->Complete == FTW_VALID_STATE`

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: @makubacki 

as per the comment of this function the last written header may be an empty one

The last write header is the header whose 'complete' state hasn't been set.
  **After all, this header may be a EMPTY header entry for next Allocate.**

makubacki: It is checking the `Complete` field in a `EFI_FAULT_TOLERANT_WRITE_HEADER` which should follow a `EFI_FAULT_TOLERANT_WORKING_BLOCK_HEADER`.

image

In the above image:

- `FtwWorkSpaceHeader` is `WORKING_BLOCK_HEADER`
- `FtwHeader` is `WRITE_HEADER`
- `FtwHeader->NumberOfWrites` specifies number of `WRITE_RECORD`s

That is saying "**this header**" may be an "**EMPTY header entry**". I interpret that to mean the header entry would be present and the signature should be valid.

Have you inspected the code flow to know that the `Signature` of `FtwWorkSpaceHeader` would not be valid? I have not looked in detail at that aspect but it seems strange to start processing write headers and records following an invalid working block header.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki: My point is that I think that should be removed from the PR description as it could mislead someone that reads it to believe that is a valid use case the driver failed to handle when it is not valid.

I mentioned the polarity in the FV header because this is an easy check after the driver locates the proper firmware volume block protocol for the FTW area, it could simply call `GetAttributes()` and check the polarity.

That is getting outside the scope of this change though. I mainly didn't want the invalid case to be mentioned as a possibility the driver should encounter. Other things than just this will not work as expected.

sureshkumarpMSFT: Sure, I agree. I will remove those statements from PR description

makubacki: Thanks. Since this is depended on for fault tolerance I just want to make sure it serves the role as expected.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: I have not yet looked the entire flow in detail. I will check on the `Signature of FtwWorkSpaceHeader ` and then we can make the appropriate changes

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki: This would reintroduce a bug fixed 12 years ago - https://github.com/tianocore/edk2/commit/fe92f438ac860a09e57a2a6a37c1a1d8b6539321.

sureshkumarpMSFT: Thank you @makubacki for finding this history. Much appreciated

makubacki: I understand this is changing the calculation, but this logic has been around long enough that we need a stronger assertion to make a change. @sureshkumarpMSFT, please double check this when you look at the flow in more detail for the other conversation thread.

sureshkumarpMSFT: Hi @makubacki @kuqin12 @VivianNK @os-d 

I would like to take a step back in making the change to remove `=` condition.
since this logic is there for long time and served well. There was bug reported with having `=` condition removed.

For testing, tried to do

_*FtwWriteHeader = (EFI_FAULT_TOLERANT_WRITE_HEADER*)((UINTN) (FtwWorkSpaceHeader) + FtwWorkSpaceSize - sizeof (EFI_FAULT_TOLERANT_WRITE_HEADER));_

and I do not see any hang or exception.

 But still, I am leaning towards not removing the `=` condition. We do not know exact scenario for the change https://github.com/tianocore/edk2/commit/fe92f438ac860a09e57a2a6a37c1a1d8b6539321 .  So, it's better to have that code as it is.

If we find a compelling case and if needed, we can work it as a separate issue later.  Please let me know your thoughts.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: I have checked the flow and it looks like FtwGetLastWriteHeader() is called after InitializeLocalWorkSpaceHeader(). 
This InitializeLocalWorkSpaceHeader() does write 0xFF (FTW_ERASED_BYTE) to EFI_FAULT_TOLERANT_WORKING_BLOCK_HEADER and copies the guid gEdkiiWorkingBlockSignatureGuid

So even in negative case (corrupted or malformed block) we will have valid guid in EFI_FAULT_TOLERANT_WORKING_BLOCK_HEADER and signature verification would PASS
hence I believe we cannot make use the signature check here. Please let me know if you think otherwise.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki: I agree. Any changes need to have a detailed assessment of the problematic scenario (including diagrams with correlation to the overall flow) and subsequent cause and effect described by the proposed change.

This code can cause silent bugs to not be caught until late in the field.

Really, a unit test demonstrating the problem and showing the fix would be ideal but a heavy lift given no existing FTW unit test code.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki: If I understand correctly, it is valuable in the case of hardware failure such as wear out of the containing flash sector where a successful write and readback is expected.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: 👍

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: Yes, this change will be useful in case such as wear, and tear causes the entire block to be 0s instead of 1s as the current code expects.

However, in any case, the code should not access the beyond `FtwWorkSpaceHeader+FtwWorkSpaceSize`

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
os-d: Agreed, thanks for the due diligence.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
ddaney-fungible: The code must be robust if corrupt data is encountered. It does not really matter the source of the corruption, it could be a hardware failure, or from defective or malicious software.

makubacki: @sureshkumarpMSFT, were you planning to add a signature check? Since this change is building robustness against malformed contents, that could potentially find the issue sooner before having to rely on this code.

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki, line 1: Have you verified this flow for a full variable store that performs reclaim?

sureshkumarpMSFT: Yes, I have verified the flow with variable store full case.
During reclaim process also, when there is no corruption, the loop exits early once FtwHeader->Complete != FTW_VALID_STATE.
And when there is a corruption, the change helps to exit the loop without having to access beyond FtwWorkSpaceSize

makubacki: Thanks. Final request is, can you please post this is a PR to the file in the [edk2 repo](https://github.com/tianocore/edk2)?

I want to make sure the maintainers there approve of this as well before we take a delta on it.

sureshkumarpMSFT: Sure

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
sureshkumarpMSFT: Hi @makubacki,   I have added the signature check. Can you please check now.

I will send the changes to EDK2 forum with all these changes

sureshkumarpMSFT: created a PR here https://github.com/tianocore/edk2/pull/6045

kuqin12 (approved):

MdeModulePkg/Universal/FaultTolerantWriteDxe/FtwMisc.c
makubacki, line 1: Please wait to complete this PR until https://github.com/tianocore/edk2/pull/6045 is completed and the edk2 commit can be cherry-picked back.

os-d:
Closing in favor of: https://github.com/microsoft/mu_basecore/pull/1131.

PR #978, opened 2024-06-26
[202405][Rebase&&FF] Variable Policy Changes

Flickdm:
# Preface
[!NOTE]  Depends on https://github.com/microsoft/mu_basecore/pull/961

## Description

Cherry Picks:

1. 1a63f67364  - Picked
2. aeb50c2       - Picked
3. 521a898       - Picked
4. 43530fa937  - Dropped
5. 5b46552d23 - Picked
6. 7a8dba5bfd  - Picked
7. dac933023a - Picked
8. 0549e53278 - Picked & splited
9. 28ec22916a - picked
10. 68670a0cef - picked
11. a8c4697c22 - picked
12. 28789f8177 - picked
13. 1fc9ff0859 - picked
14. 8c63697118 - picked
15. 383ec13340 - picked

This pull request primarily focuses on modifying the UEFI boot manager library and the variable policy library to improve variable policy enforcement and boot option handling. The most significant changes include the addition of new load option types, the implementation of variable policy locking, and the modification of several methods to ensure better policy enforcement and error handling.

Changes to UEFI Boot Manager Library:

1. `MdeModulePkg/Include/Library/UefiBootManagerLib.h`: Expanded the `Create` method to handle additional load option types like `SysPrep####` and `PlatformRecovery####`. Also, added a lock now variable policy for the `PlatformRecovery####` variable if the `SetVariable` call was successful. [[1]](diffhunk://#diff-51420d64292debad5769d7c521ea067d7f493c26534c3dbb5b7483eb49dd04d2L158-R165) [[2]](diffhunk://#diff-473f029e0700daf8817b1238114f00f886f50a311304c3b7048e9eb31f6c5934R163-R168)
2. `MdeModulePkg/Library/UefiBootManagerLib/BmBoot.c`: Added a constructor for `UefiBootManagerLib` that locates the `VariablePolicy` protocol and registers a basic variable policy.
3. `MdeModulePkg/Library/UefiBootManagerLib/BmLoadOption.c`: Modified the `BmGetFreeOptionNumber` method to register a lock now variable policy for the `PlatformRecovery####` variable if the `SetVariable` call was successful. Also, moved `RegisterBasicVariablePolicy` after `SetVariable` for `OptionName`.
[...]

PR #1102, opened 2024-08-09
[Release/202405] Moves placement of variable policy registration to after the last variable that is "Read-Only"

Flickdm:
## Description

Attach to https://github.com/microsoft/mu_basecore/pull/978

This change moves the locking of the variable policy to below the last Read-Only variable. This is necessary to ensure that the variable policy is not locked until after read only variables are set. Otherwise, the SetVariable On a Read-Only variable will fail with "WRITE_PROTECTED" error.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

QC Surface Platform

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/BdsDxe/BdsEntry.c
+++ b/MdeModulePkg/Universal/BdsDxe/BdsEntry.c
@@ -751,27 +751,10 @@ BdsEntry (
   //
   BdsFormalizeEfiGlobalVariable ();
 
+  // MU_CHANGE
   //
-  // Mark the read-only variables if the Variable Lock protocol exists
+  // Initialize Hardware Error Recording Support.
   //
-  Status = gBS->LocateProtocol (&gEdkiiVariablePolicyProtocolGuid, NULL, (VOID **)&VariablePolicy);
-  DEBUG ((DEBUG_INFO, "[BdsDxe] Locate Variable Policy protocol - %r\n", Status));
-  if (!EFI_ERROR (Status)) {
-    for (Index = 0; Index < ARRAY_SIZE (mReadOnlyVariables); Index++) {
-      Status = RegisterBasicVariablePolicy (
-                 VariablePolicy,
-                 &gEfiGlobalVariableGuid,
-                 mReadOnlyVariables[Index],
-                 VARIABLE_POLICY_NO_MIN_SIZE,
-                 VARIABLE_POLICY_NO_MAX_SIZE,
-                 VARIABLE_POLICY_NO_MUST_ATTR,
-                 VARIABLE_POLICY_NO_CANT_ATTR,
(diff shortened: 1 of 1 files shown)


kenlautner (approved):

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -832,6 +815,34 @@ BdsEntry (
...
+
+  //
+  // READ ONLY Variable (registered in mReadOnlyVariables) MUST BE be set before this point.
+  //
+
+  //
+  // Mark the read-only variables if the Variable Lock protocol exists
+  //
+  Status = gBS->LocateProtocol (&gEdkiiVariablePolicyProtocolGuid, NULL, (VOID **)&VariablePolicy);
+  DEBUG ((DEBUG_INFO, "[BdsDxe] Locate Variable Policy protocol - %r\n", Status));
```
os-d: We can probably just print if we fail to find the protocol.

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
```diff
@@ -832,6 +815,34 @@ BdsEntry (
...
+                 VariablePolicy,
+                 &gEfiGlobalVariableGuid,
+                 mReadOnlyVariables[Index],
+                 VARIABLE_POLICY_NO_MIN_SIZE,
+                 VARIABLE_POLICY_NO_MAX_SIZE,
+                 VARIABLE_POLICY_NO_MUST_ATTR,
+                 VARIABLE_POLICY_NO_CANT_ATTR,
+                 VARIABLE_POLICY_TYPE_LOCK_NOW
+                 );
+      ASSERT_EFI_ERROR (Status);
```
os-d: I think worth adding a print if we fail to register the policy.

os-d (approved):
The PR description says we do this to ensure policy is locked before any variable is accessed. Can you change this to indicate we are locking after we finishing setting the variables?

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
makubacki: Probably should even assert if you fail to find the protocol.

makubacki (approved):

MdeModulePkg/Universal/BdsDxe/BdsEntry.c
Flickdm: Do we expect every edk2 based platform to produce the protocol at this point?

makubacki: I'd expect every Mu project to not every edk2 platform.

PR #1127, opened 2024-08-29
[202405] ImageValidation.py: Don't parse entire image

Javagedes:
## Description

This commit modifies the PE parsing functionality to only parse the headers of the image, rather than the entire image. This change is made to improve performance and also the probability of failing to parse the entire image. This comes after this commit (https://github.com/erocarrera/pefile/pull/365) in pefile resulted in efi image parsing failures, breaking the build.

This commit also wraps the parsing of the image in a try-except block to catch any exceptions that may be raised during parsing, to cleanly exit.

See: https://github.com/microsoft/mu_tiano_platforms/pull/1025 and https://github.com/erocarrera/pefile/issues/421

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Validated pipelines build on mu_tiano_platforms

## Integration Instructions

N/A

### DIFF
--- a/.pytool/Plugin/ImageValidation/ImageValidation.py
+++ b/.pytool/Plugin/ImageValidation/ImageValidation.py
@@ -93,7 +93,6 @@ class ImageValidation(IUefiBuildPlugin):
 
         # Load Configuration Data
         config_path = thebuilder.env.GetValue("PE_VALIDATION_PATH", None)
-        tool_chain_tag = thebuilder.env.GetValue("TOOL_CHAIN_TAG")
         if config_path is None:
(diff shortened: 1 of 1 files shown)


os-d (approved):
Straight to edk2? I'm good to take this here too, so we aren't broken, of course. Or has this plugin not gone upstream?

Javagedes:
> Straight to edk2? I'm good to take this here too, so we aren't broken, of course. Or has this plugin not gone upstream?

Not upstreamed yet

apop5 (approved):

PR #1174, opened 2024-10-03
MdeModulePkg: Add InsecureDeviceState PCD

VivianNK:
## Description

Add PCD to provide a FixedAtBuild value to define what an insecure Device state is.

For example, the default value defines an insecure device state as a DEVICE_STATE with DEVICE_STATE_SOURCE_DEBUG_ENABLED, or DEVICE_STATE_UNIT_TEST_MODE.

Add a function to the DeviceStateLib, GetInsecureDeviceStateSettings, that retrieves the PCD. Expected use cases include comparing this PCD against current device state to check for an insecure device state, then measuring into the TPM.

This PR has a test that's dependent on a mock for PcdLib being added in MdePkg https://github.com/microsoft/mu_basecore/pull/1178

- [ ] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

DeviceStateLibGoogleTest.inf

## Integration Instructions

Update the FixedAtBuild PCD InsecureDeviceState to include any device states that the platform considers insecure. Call this API to take action if an insecure device state has been entered.

Expected use cases include comparing this bitmask against the current device state to check
for an insecure device state, then measuring into the TPM. 

Make any modifications needed to accommodate all OEM-defined device states. 

To add the measurement of this PCD in Tcg2Pei, update your mu_tiano_plus submodule to the release/202405 branch or release tag once this change is merged (https://github.com/microsoft/mu_tiano_plus/pull/342).

### DIFF
--- a/MdeModulePkg/Include/Library/DeviceStateLib.h
+++ b/MdeModulePkg/Include/Library/DeviceStateLib.h
@@ -21,24 +21,27 @@ MU_CHANGE: new file
 #define DEVICE_STATE_UNDEFINED                  (1 << (4))
 #define DEVICE_STATE_UNIT_TEST_MODE             (1 << (5))
 
-#define DEVICE_STATE_PLATFORM_MODE_0  (1 << (20))
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  20
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    27
+
+#define DEVICE_STATE_PLATFORM_MODE_0  (1 << (DEVICE_STATE_PLATFORM_MODE_INDEX_START))
 #define DEVICE_STATE_PLATFORM_MODE_1  (1 << (21))
 #define DEVICE_STATE_PLATFORM_MODE_2  (1 << (22))
 #define DEVICE_STATE_PLATFORM_MODE_3  (1 << (23))
 #define DEVICE_STATE_PLATFORM_MODE_4  (1 << (24))
 #define DEVICE_STATE_PLATFORM_MODE_5  (1 << (25))
 #define DEVICE_STATE_PLATFORM_MODE_6  (1 << (26))
-#define DEVICE_STATE_PLATFORM_MODE_7  (1 << (27))
+#define DEVICE_STATE_PLATFORM_MODE_7  (1 << (DEVICE_STATE_PLATFORM_MODE_INDEX_END))
 
 #define DEVICE_STATE_MAX  (1 << (31))
 
-#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  20
-#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    27
-
 #define MAX_INSECURE_DEVICE_STATE_STRING_SIZE  512
 
+// The number of library-defined device states, not including OEM-defined states.
+#define NUMBER_OF_DEVICE_STATE_STRINGS  6
+
 //
-// DEFINE the device state strings here. These are used for logging and measuring purposes.
+// These strings are used to represent device state for logging and measurements.
 //
 #define DEVICE_STATE_SECUREBOOT_OFF_STRING             "SECURE_BOOT_OFF "
 #define DEVICE_STATE_MANUFACTURING_MODE_STRING         "MANUFACTURING_MODE "
@@ -51,48 +54,68 @@ MU_CHANGE: new file
 typedef UINT32 DEVICE_STATE;
 
 /**
- * Get String that represents the device's current insecure states, using the platform defined set of
- * insecure device states. This can be used to measure insecure device states into the TPM or perform
- * other required platform actions when the device enters an insecure state.
- *
- * @param[out] Buffer - Buffer to store the string
- * @param[in] MaxSize - Maximum size of the buffer
- * @retval The The number of characters in the string, including the null character.
- */
-UINTN
+  Get String that represents the device's current insecure states, using the platform defined set of
+  insecure device states. This can be used to measure insecure device states into the TPM or perform
+  other required platform actions when the device enters an insecure state.
+
+  The string convention to be returned is described as follows:
+  A concatenation of all the insecure device states that are currently active, represented in
+  MACRO_CASE, separated by a space. The string, if not empty, will end in a space followed by a Null terminator.
+
+  Examples:
+    - "SECUREBOOT_OFF MANUFACTURING_MODE "
+    - "DEVELOPMENT_BUILD_ENABLED SOURCE_DEBUG_ENABLED UNIT_TEST_MODE "
+
+  @param[out] Destination Buffer to write the string
+  @param[in]  DestMax     Maximum size of the destination buffer
+  @param[out] StringSize  Size of the Null-terminated Ascii string in bytes, including the Null terminator.
+
+  @retval EFI_SUCCESS            String is written to buffer.
+  @retval EFI_BUFFER_TOO_SMALL   If DestMax is NOT greater than StrLen to be written.
+  @retval EFI_INVALID_PARAMETER  If Destination is NULL.
+                                    If StringSize is NULL.
+                                    If DestMax is 0.
+                                    If the bitmask is out of range.
+ **/
+EFI_STATUS
 EFIAPI
 GetInsecureDeviceStateString (
-  IN OUT CHAR8  *Buffer,
-  IN UINTN      MaxSize
+  IN OUT CHAR8  *Destination,
+  IN UINTN      DestMax,
+  IN OUT UINTN  *StringSize
   );
 
 /**
- * Get the platform defined set of insecure device states. This can be used to measure insecure device
- * states into the TPM or perform other required platform actions when the device enters an insecure state.
- *
- * @retval The bitmask of insecure device states as defined by the platform.
- */
+  Get the platform defined set of insecure device states. This can be used to measure insecure device
+  states into the TPM or perform other required platform actions when the device enters an insecure state.
+
+  @return The bitmask of insecure device states as defined by the platform.
+**/
 DEVICE_STATE
 EFIAPI
 GetInsecureDeviceStateSetting (
   );
 
 /**
- * Function to Get current device state
- *
- * @retval the current DEVICE_STATE
- */
+  Get current device state
+
+  @return the current DEVICE_STATE
+**/
 DEVICE_STATE
 EFIAPI
 GetDeviceState (
+  VOID
   );
 
 /**
- * Function to Add additional bits to the device state
- *
- * @param[in] AdditionalState - additional state to set active
- * @retval Status of operation.  EFI_SUCCESS on successful update.
- */
+  Add an additional device state
+
+  @param[in] AdditionalState  The additional state to add.
+
+  @retval EFI_SUCCESS             The device state was added successfully.
+  @retval RETURN_OUT_OF_RESOURCES The device state could not be added due to lack of resources.
+  @retval RETURN_DEVICE_ERROR     The device state could not be added.
+**/
 RETURN_STATUS
 EFIAPI
 AddDeviceState (
--- a/MdeModulePkg/Include/Library/DeviceStateLib.h
+++ b/MdeModulePkg/Include/Library/DeviceStateLib.h
@@ -1,4 +1,4 @@
-/** @file
+/** @file DeviceStateLib.h
 Functions used to support Getting and Setting Device States.
 
 Copyright (C) Microsoft Corporation.
@@ -32,27 +32,71 @@ MU_CHANGE: new file
 
 #define DEVICE_STATE_MAX  (1 << (31))
 
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  20
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    27
+
+#define MAX_INSECURE_DEVICE_STATE_STRING_SIZE  512
+
+//
+// DEFINE the device state strings here. These are used for logging and measuring purposes.
+//
+#define DEVICE_STATE_SECUREBOOT_OFF_STRING             "SECURE_BOOT_OFF "
+#define DEVICE_STATE_MANUFACTURING_MODE_STRING         "MANUFACTURING_MODE "
+#define DEVICE_STATE_DEVELOPMENT_BUILD_ENABLED_STRING  "DEVELOPMENT_BUILD_ENABLED "
+#define DEVICE_STATE_SOURCE_DEBUG_ENABLED_STRING       "SOURCE_DEBUG_ENABLED "
+#define DEVICE_STATE_UNDEFINED_STRING                  "UNDEFINED "
+#define DEVICE_STATE_UNIT_TEST_MODE_STRING             "UNIT_TEST_MODE "
+#define DEVICE_STATE_PLATFORM_MODE_STRING              "OEM_DEFINED "
+
 typedef UINT32 DEVICE_STATE;
 
 /**
-Function to Get current device state
-@retval the current DEVICE_STATE
-**/
+ * Get String that represents the device's current insecure states, using the platform defined set of
+ * insecure device states. This can be used to measure insecure device states into the TPM or perform
+ * other required platform actions when the device enters an insecure state.
+ *
+ * @param[out] Buffer - Buffer to store the string
+ * @param[in] MaxSize - Maximum size of the buffer
+ * @retval The The number of characters in the string, including the null character.
+ */
+UINTN
+EFIAPI
+GetInsecureDeviceStateString (
+  IN OUT CHAR8  *Buffer,
+  IN UINTN      MaxSize
+  );
+
+/**
+ * Get the platform defined set of insecure device states. This can be used to measure insecure device
+ * states into the TPM or perform other required platform actions when the device enters an insecure state.
+ *
+ * @retval The bitmask of insecure device states as defined by the platform.
+ */
+DEVICE_STATE
+EFIAPI
+GetInsecureDeviceStateSetting (
+  );
+
+/**
+ * Function to Get current device state
+ *
+ * @retval the current DEVICE_STATE
+ */
 DEVICE_STATE
 EFIAPI
 GetDeviceState (
   );
 
 /**
-Function to Add additional bits to the device state
-
-@param AdditionalState - additional state to set active
-@retval Status of operation.  EFI_SUCCESS on successful update.
-**/
+ * Function to Add additional bits to the device state
+ *
+ * @param[in] AdditionalState - additional state to set active
+ * @retval Status of operation.  EFI_SUCCESS on successful update.
+ */
 RETURN_STATUS
 EFIAPI
 AddDeviceState (
-  DEVICE_STATE  AdditionalState
+  IN DEVICE_STATE  AdditionalState
   );
 
 #endif
--- a/MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
+++ b/MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
@@ -1,6 +1,7 @@
 /** @file DeviceStateLib.c
 Functions used to support Getting and Setting Device States.
-This library uses the PcdDeviceStateBitmask dynamic PCD to store the device state
+This library uses the PcdDeviceStateBitmask dynamic PCD to store the device state.
+It also uses the PcdInsecureDeviceState fixed PCD to retrieve the platform defined set of insecure device states.
 
 Copyright (C) Microsoft Corporation.
 SPDX-License-Identifier: BSD-2-Clause-Patent
@@ -13,7 +14,7 @@ MU_CHANGE: new file
 #include <Library/DeviceStateLib.h>
 #include <Library/DebugLib.h>
 
-CHAR8  *DEVICE_STATE_STRINGS[6] = {
+CHAR8  *DEVICE_STATE_STRINGS[NUMBER_OF_DEVICE_STATE_STRINGS] = {
   DEVICE_STATE_SECUREBOOT_OFF_STRING,
   DEVICE_STATE_MANUFACTURING_MODE_STRING,
   DEVICE_STATE_DEVELOPMENT_BUILD_ENABLED_STRING,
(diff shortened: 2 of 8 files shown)


MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -43,6 +45,14 @@ EFIAPI
 GetDeviceState (
   );
 
+/**
+Function to get InsecureDeviceState PCD
+**/
+INSECURE_DEVICE_STATE
```
os-d: I think this can just return a `DEVICE_STATE` since it is being used to compare against a `DEVICE_STATE`.

MdeModulePkg/Library/DeviceStateLib/Readme.md
```diff
@@ -26,6 +26,10 @@ states or to define any of the unused bits:
 * BIT 26: DEVICE_STATE_PLATFORM_MODE_2
 * BIT 27: DEVICE_STATE_PLATFORM_MODE_3
 
+### Insecure Device State
+This library has a function, GetInsecureDeviceState, that retrieves the FixedAtBuild PCD,
+GetInsecureDeviceState. This PCD, by default, ORs DEVICE_STATE_SOURCE_DEBUG_ENABLED with DEVICE_STATE_UNIT_TEST_MODE, but can be customized to include other states.
```
os-d: Can you add some more details here on the expected use case for this (i.e. measuring in to the TPM)

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1386,6 +1386,12 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdMaxMemoryTypeInfoPages            |0x60000|UINT32|0x40000150
   # MU_CHANGE TCBZ1086 [END]
 
+  ## MU_CHANGE [BEGIN] - Add InsecureDeviceState to MdeModulePkg
+  # @Prompt Describes an insecure device state. 
+  # Default value is bits 3 and 5 as defined in Include/Library/DeviceStateLib.h
+  gEfiMdeModulePkgTokenSpaceGuid.PcdInsecureDeviceState|0x00000000|UINT32|0x000000028
```
os-d: This is the wrong format. It should be `Name|Value|Type|UniqueIdentifier`

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,6 +13,20 @@ MU_CHANGE: new file
...
+/**
+Function to get InsecureDeviceState
+@retval the InsecureDeviceState PCD value
+**/
+INSECURE_DEVICE_STATE
+EFIAPI
+GetInsecureDeviceState (
+  )
+{
+  INSECURE_DEVICE_STATE  InsecureDevState = PcdGet32 (PcdInsecureDeviceState);
```
os-d: Let's use `FixedPcdGet32`

MdeModulePkg/Library/DeviceStateLib/Readme.md
os-d: Can you similarly add the intended use case for this to the PR description?

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -43,6 +45,14 @@ EFIAPI
 GetDeviceState (
   );
 
+/**
+Function to get InsecureDeviceState PCD
+**/
+INSECURE_DEVICE_STATE
+EFIAPI
+GetInsecureDeviceState (
```
cfernald: Maybe GetInsecureDeviceStateSettings or something that conveys what this returns a bit better.

MdeModulePkg/Library/DeviceStateLib/Readme.md
VivianNK: Updated ReadMe

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -43,6 +43,14 @@ EFIAPI
 GetDeviceState (
   );
 
+/**
+Function to get InsecureDeviceState PCD
```
os-d: I missed this before. Can we also not call out the PCD here? This should just describe the interface, not implementation details, so that if the library is overridden, it can change the implementation as it needs. And then this comment should be exactly the same as in the c file.

So something like

```c
/**
  Get the platform defined set of insecure device states. This can be used to measure insecure device
  states into the TPM or perform other required platform actions when the device enters an insecure state.

  @retval  The bitmask of insecure device states as defined by the platform.
**/
```

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,6 +13,20 @@ MU_CHANGE: new file
...
+/**
+Function to get InsecureDeviceState
+@retval the DEVICE_STATE PCD value
+**/
+DEVICE_STATE
+EFIAPI
+GetInsecureDeviceStateSetting (
+  )
+{
+  DEVICE_STATE  InsecureDevState = FixedPcdGet32 (PcdInsecureDeviceState);
```
os-d: nit: You can just return this directly.

MdeModulePkg/Library/DeviceStateLib/Readme.md
```diff
@@ -26,6 +26,12 @@ states or to define any of the unused bits:
 * BIT 26: DEVICE_STATE_PLATFORM_MODE_2
 * BIT 27: DEVICE_STATE_PLATFORM_MODE_3
 
+### Insecure Device State
+This library has a function, GetInsecureDeviceState, that retrieves the FixedAtBuild PCD,
+InsecureDeviceState. This PCD, by default, ORs DEVICE_STATE_SOURCE_DEBUG_ENABLED with DEVICE_STATE_UNIT_TEST_MODE, but can be customized to include other states.
```
os-d:
```suggestion
gMdeModulePkgTokenSpaceGuid.InsecureDeviceState. This PCD, by default, ORs DEVICE_STATE_SOURCE_DEBUG_ENABLED with DEVICE_STATE_UNIT_TEST_MODE, but can be customized to include other states.
```

MdeModulePkg/Library/DeviceStateLib/Readme.md
```diff
@@ -26,6 +26,12 @@ states or to define any of the unused bits:
...
+
+Expected use cases include comparing this PCD against current device state to check for an insecure device state, then measuring into the TPM.
```
os-d:
```suggestion
Expected use cases include comparing this bitmask against the current device state to check for an insecure device state, then measuring into the TPM. A platform can update this PCD with any other states it considers to be insecure and can call this API to take action if an insecure device state has been entered.
```

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -1386,6 +1386,12 @@
   gEfiMdeModulePkgTokenSpaceGuid.PcdMaxMemoryTypeInfoPages            |0x60000|UINT32|0x40000150
   # MU_CHANGE TCBZ1086 [END]
 
+  ## MU_CHANGE [BEGIN] - Add InsecureDeviceState to MdeModulePkg
+  # @Prompt Describes an insecure device state. 
+  # Default value is bits 3 and 5 as defined in Include/Library/DeviceStateLib.h
```
os-d:
```suggestion
  # Default value is DEVICE_STATE_SOURCE_DEBUG_ENABLED | DEVICE_STATE_UNIT_TEST_MODE as defined in Include/Library/DeviceStateLib.h

```

os-d (approved):

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,85 @@ MU_CHANGE: new file
 
 #define DEVICE_STATE_MAX  (1 << (31))
 
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  20
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    27
+
+#define MAX_INSECURE_DEVICE_STATE_STRING_SIZE  512
```
os-d, line 38: Is this from the TCG spec?

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,129 @@ MU_CHANGE: new file
...
+  OUT CHAR8  *Buffer,
+  IN UINTN   MaxSize
+  )
+{
+  CHAR8         *String;
+  DEVICE_STATE  BitmaskFromIndex;
+  DEVICE_STATE  CurrentDeviceState         = GetDeviceState ();
+  DEVICE_STATE  InsecureDeviceStateSetting = GetInsecureDeviceStateSetting ();
+
+  UINTN  NumberOfDeviceStates = sizeof (DEVICE_STATE_STRINGS) / sizeof (DEVICE_STATE_STRINGS[0]);
```
os-d: Where possible, we should avoid UINTN. It should only be used in a case where we actually need a different size on different architectures (like for computing integers as pointers.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,129 @@ MU_CHANGE: new file
...
+
+  //
+  // Check standard device states
+  //
+  for (UINTN i = 0; i < NumberOfDeviceStates; i++) {
+    BitmaskFromIndex = (1 << (i));
```
os-d: nit: none of the parentheses are needed

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,129 @@ MU_CHANGE: new file
...
+    }
+  }
+
+  //
+  // Check OEM defined insecure device states
+  //
+  for (UINTN i = DEVICE_STATE_PLATFORM_MODE_INDEX_START; i <= DEVICE_STATE_PLATFORM_MODE_INDEX_END; i++) {
+    BitmaskFromIndex = (1 << (i));
+    if ((DeviceStateBitmaskIsSet (InsecureDeviceStateSetting, BitmaskFromIndex)) &&
+        (DeviceStateBitmaskIsSet (CurrentDeviceState, BitmaskFromIndex)))
```
os-d: This API doesn't seem to add value. We can drop it and just directly do the bitwise & here.

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,85 @@ MU_CHANGE: new file
...
+
+#define MAX_INSECURE_DEVICE_STATE_STRING_SIZE  512
+
+//
+// DEFINE the device state strings here. These are used for logging and measuring purposes.
+//
+#define DEVICE_STATE_SECUREBOOT_OFF_STRING             "SECURE_BOOT_OFF "
+#define DEVICE_STATE_MANUFACTURING_MODE_STRING         "MANUFACTURING_MODE "
+#define DEVICE_STATE_DEVELOPMENT_BUILD_ENABLED_STRING  "DEVELOPMENT_BUILD_ENABLED "
+#define DEVICE_STATE_SOURCE_DEBUG_ENABLED_STRING       "SOURCE_DEBUG_ENABLED "
```
os-d, line 49: I assume in Tcg2Pei you will also measure the specific TCG spec defined debugger string?

os-d (approved):

MdeModulePkg/Include/Library/DeviceStateLib.h
VivianNK: https://github.com/microsoft/mu_tiano_plus/pull/342/files#diff-dd252bb953ec11695bb497d99150fe3c48e069945f08a98630c4a2e47baef3b3R660
yes

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
VivianNK: I added it for readability, specifically for consumption in Tcg2Pei. (https://github.com/microsoft/mu_tiano_plus/pull/342/files#diff-dd252bb953ec11695bb497d99150fe3c48e069945f08a98630c4a2e47baef3b3R1169)
But I can remove it if it's not achieving that goal/it's not much better than bitwise operation. I can add to the function descriptions that use it instead

MdeModulePkg/Include/Library/DeviceStateLib.h
VivianNK: Discussed offline, there's no TCG spec defined max

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
VivianNK: dropped

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
+ * insecure device states. This can be used to measure insecure device states into the TPM or perform
+ * other required platform actions when the device enters an insecure state.
+ *
+ * @param[out] Buffer - Buffer to store the string
+ * @param[in] MaxSize - Maximum size of the buffer
+ * @retval The The number of characters in the string, including the null character.
+ */
+UINTN
+EFIAPI
+GetInsecureDeviceStateString (
```
makubacki, line 82: Since this function is at the interface level, it should state the string format returned as a contract to the caller. For example, the string convention ("macro case"), the separation character (e.g. space), etc.

Although someone could postulate that the macros above which are "macro case" and contain space separators are concatenated together based on looking at the entirety of this file that's different from an explicitly stated interface requirement.

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
 /**
-Function to Get current device state
-@retval the current DEVICE_STATE
-**/
+ * Get String that represents the device's current insecure states, using the platform defined set of
+ * insecure device states. This can be used to measure insecure device states into the TPM or perform
+ * other required platform actions when the device enters an insecure state.
+ *
+ * @param[out] Buffer - Buffer to store the string
+ * @param[in] MaxSize - Maximum size of the buffer
```
makubacki: The interface should explicitly describe the behavior taken when `MaxSize` insufficient to hold the string.

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
+ * @retval The The number of characters in the string, including the null character.
+ */
```
makubacki: You can drop the excess asterisks. For example:

https://github.com/microsoft/mu_basecore/blob/a45519aa3d4e033caed186eff927557381282bf4/MdePkg/Include/Library/BaseMemoryLib.h#L16-L32

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
-Function to Get current device state
-@retval the current DEVICE_STATE
-**/
+ * Get String that represents the device's current insecure states, using the platform defined set of
+ * insecure device states. This can be used to measure insecure device states into the TPM or perform
+ * other required platform actions when the device enters an insecure state.
+ *
+ * @param[out] Buffer - Buffer to store the string
+ * @param[in] MaxSize - Maximum size of the buffer
+ * @retval The The number of characters in the string, including the null character.
```
makubacki: The function behavior when an invalid parameter is given should be defined.

For example:

- What is the expected behavior if `Buffer` is `NULL`?
- What is the expected behavior if `MaxSize` is `0`?

makubacki: I suggest using `@return` here versus `@retval`.

From EDK II C Coding Standards Specification:
image

An example that interestingly uses both, but the character count is `@return`:

https://github.com/microsoft/mu_basecore/blob/a45519aa3d4e033caed186eff927557381282bf4/MdePkg/Include/Library/BaseLib.h#L1027-L1037

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
 EFIAPI
-GetDeviceState (
+GetInsecureDeviceStateSetting (
   );
 
 /**
-Function to Add additional bits to the device state
+ * Function to Get current device state
+ *
+ * @retval the current DEVICE_STATE
```
makubacki: Another case where I think `@return` makes sense.

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
+ */
 DEVICE_STATE
 EFIAPI
-GetDeviceState (
+GetInsecureDeviceStateSetting (
   );
 
 /**
-Function to Add additional bits to the device state
+ * Function to Get current device state
```
makubacki: No need to precede with "Function to..".

```suggestion
 Get the current device state.
```

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
+ *
+ * @retval the current DEVICE_STATE
+ */
+DEVICE_STATE
+EFIAPI
+GetDeviceState (
+  );
```
makubacki: Per EDK II C Coding Standards Specification.

image

```suggestion
GetDeviceState (
  VOID
  );
```

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
...
 
-@param AdditionalState - additional state to set active
-@retval Status of operation.  EFI_SUCCESS on successful update.
-**/
+/**
+ * Function to Add additional bits to the device state
+ *
+ * @param[in] AdditionalState - additional state to set active
+ * @retval Status of operation.  EFI_SUCCESS on successful update.
+ */
```
makubacki:
```suggestion
/**
  Add an additional device state.
  
  @param[in]  AdditionalState    The additional device to state to add.
  
  @retval  EFI_SUCCESS           The device state was added successfully.
**/
```

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
 #include <Library/DeviceStateLib.h>
 #include <Library/DebugLib.h>
 
+CHAR8  *DEVICE_STATE_STRINGS[6] = {
```
makubacki: I suggest making `6` a macro defined value for clarity. For example, the earlier code defined `7` strings:

```c
#define DEVICE_STATE_SECUREBOOT_OFF_STRING             "SECURE_BOOT_OFF "
#define DEVICE_STATE_MANUFACTURING_MODE_STRING         "MANUFACTURING_MODE "
#define DEVICE_STATE_DEVELOPMENT_BUILD_ENABLED_STRING  "DEVELOPMENT_BUILD_ENABLED "
#define DEVICE_STATE_SOURCE_DEBUG_ENABLED_STRING       "SOURCE_DEBUG_ENABLED "
#define DEVICE_STATE_UNDEFINED_STRING                  "UNDEFINED "
#define DEVICE_STATE_UNIT_TEST_MODE_STRING             "UNIT_TEST_MODE "
#define DEVICE_STATE_PLATFORM_MODE_STRING              "OEM_DEFINED "
```

So, if the macro name indicated why these 6 were chosen and `DEVICE_STATE_PLATFORM_MODE_STRING` was excluded that would make it clear it was intentional and how to determine whether future strings should be added here or not.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+ *
+ * @param[out] Buffer - Buffer to store the string
+ * @param[in] MaxSize - Maximum size of the buffer
+ * @retval The The number of characters in the string, including the null character.
+ */
+UINTN
 EFIAPI
-GetDeviceState (
+GetInsecureDeviceStateString (
+  OUT CHAR8  *Buffer,
```
makubacki: Need to handle the case where `Buffer` is `NULL` to prevent null pointer dereference.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+  DEVICE_STATE  BitmaskFromIndex;
+  DEVICE_STATE  CurrentDeviceState         = GetDeviceState ();
+  DEVICE_STATE  InsecureDeviceStateSetting = GetInsecureDeviceStateSetting ();
+
+  UINT32  NumberOfDeviceStates = sizeof (DEVICE_STATE_STRINGS) / sizeof (DEVICE_STATE_STRINGS[0]);
+
+  //
+  // Check standard device states
+  //
+  for (UINT32 i = 0; i < NumberOfDeviceStates; i++) {
```
makubacki: Typically, loop indices are named `Index`. I personally prefer that to `i` for consistency with the rest of the codebase.

EDK II C Coding Standards Specification example:

image

makubacki: I'm not aware of an actual requirement here though. If you prefer `i`, please keep it.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+  IN UINTN   MaxSize
   )
 {
-  DEVICE_STATE  DevState = PcdGet32 (PcdDeviceStateBitmask);
+  CHAR8         *String;
+  DEVICE_STATE  BitmaskFromIndex;
+  DEVICE_STATE  CurrentDeviceState         = GetDeviceState ();
+  DEVICE_STATE  InsecureDeviceStateSetting = GetInsecureDeviceStateSetting ();
+
+  UINT32  NumberOfDeviceStates = sizeof (DEVICE_STATE_STRINGS) / sizeof (DEVICE_STATE_STRINGS[0]);
```
makubacki: Please use `ARRAY_SIZE`.

```suggestion
  UINT32  NumberOfDeviceStates = ARRAY_SIZE (DEVICE_STATE_STRINGS);
```

makubacki: If there is not a compelling reason for `NumberOfDeviceStates` to be a `UINT32`, it should be a `UINTN`.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+      String = DEVICE_STATE_STRINGS[i];
+      AsciiStrnCatS (Buffer, MaxSize, String, AsciiStrLen (String));
+    }
+  }
 
-  return DevState;
+  //
+  // Check OEM-defined device states
+  //
+  for (UINT32 i = DEVICE_STATE_PLATFORM_MODE_INDEX_START; i <= DEVICE_STATE_PLATFORM_MODE_INDEX_END; i++) {
```
makubacki: We typically don't declare variables later in functions. For consistency, I suggest placing the variable declaration at the beginning of the function.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
makubacki: The loop index should also be of type `UINTN` unless there is a strong need otherwise since it sized to the pointer size of the platform and the index is used as an offset from the base address of the array.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+  DEVICE_STATE  CurrentDeviceState         = GetDeviceState ();
+  DEVICE_STATE  InsecureDeviceStateSetting = GetInsecureDeviceStateSetting ();
+
+  UINT32  NumberOfDeviceStates = sizeof (DEVICE_STATE_STRINGS) / sizeof (DEVICE_STATE_STRINGS[0]);
+
+  //
+  // Check standard device states
+  //
+  for (UINT32 i = 0; i < NumberOfDeviceStates; i++) {
+    BitmaskFromIndex = 1 << i;
```
makubacki: A check to make sure the shift operation of `1 << i` is within range of `BitmaskFromIndex` should occur before this (i.e. `DEVICE_STATE_MAX` specifies the maximum as `31`).

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
makubacki: If I understand correctly, I suggest calling this variable `NumberOfDeviceStateStrings`.

```suggestion
  UINT32  NumberOfDeviceStateStrings = sizeof (DEVICE_STATE_STRINGS) / sizeof (DEVICE_STATE_STRINGS[0]);
```

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+      //
+      // An OEM insecure device state is detected. Append to the insecure device state string.
+      //
+      String = DEVICE_STATE_PLATFORM_MODE_STRING;
+      AsciiStrnCatS (Buffer, MaxSize, String, AsciiStrLen (String));
+      break;
+    }
+  }
+
+  return AsciiStrnLenS (Buffer, MaxSize) + 1;
```
makubacki: Was there a reason not to call `AsciiStrnSizeS()`? It includes the null character taking into account platform character width.

https://github.com/microsoft/mu_basecore/blob/a45519aa3d4e033caed186eff927557381282bf4/MdePkg/Library/BaseLib/SafeString.c#L1762-L1770

```suggestion
  return AsciiStrnSizeS (Buffer, MaxSize);
```

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -13,30 +13,119 @@ MU_CHANGE: new file
...
+EFIAPI
+GetInsecureDeviceStateSetting (
+  )
+{
+  return FixedPcdGet32 (PcdInsecureDeviceState);
 }
 
 /**
-Function to Add additional bits to the device state
+ * Check if a specific bitmask is set in the device state
```
makubacki: Is a function definition missing between these comments?

```c
/**
 * Check if a specific bitmask is set in the device state
 *
 * @param[in] CurrentDeviceState - the current device state
 * @param[in] DeviceStateBitmask - the bitmask to check
 * @retval TRUE if all the bits in the bitmask are set, FALSE otherwise
 */

/**
 * Function to Get current device state
 *
 * @retval the current DEVICE_STATE
 */
```

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
```diff
@@ -1,4 +1,4 @@
-/** @file
+/** @file DeviceStateLib.c
 Functions used to support Getting and Setting Device States.
```
makubacki, line 2: Since this talks about `PcdDeviceStateBitmask`, you might also want to describe how this library instance uses `PcdInsecureDeviceState` (i.e. fixed PCD).

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
 
 #define DEVICE_STATE_MAX  (1 << (31))
 
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  20
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    27
```
makubacki: Not a big deal, but personal preference is to directly associate the values with the reason they were chosen.

```suggestion
#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  DEVICE_STATE_PLATFORM_MODE_0
#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    DEVICE_STATE_PLATFORM_MODE_7
```

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -32,27 +32,71 @@ MU_CHANGE: new file
 
 #define DEVICE_STATE_MAX  (1 << (31))
 
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_START  20
+#define DEVICE_STATE_PLATFORM_MODE_INDEX_END    27
+
+#define MAX_INSECURE_DEVICE_STATE_STRING_SIZE  512
+
+//
+// DEFINE the device state strings here. These are used for logging and measuring purposes.
```
makubacki:
```suggestion
// These strings are used to represent device state for logging and measurements.
```

MdeModulePkg/Include/Library/DeviceStateLib.h
```diff
@@ -1,4 +1,4 @@
-/** @file
+/** @file DeviceStateLib.h
```
makubacki, line 1: The integration instructions for the PR should have details for platforms to understand how to integrate and use this change (it is currently `N/A`).

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
os-d: Ha, for this and the loop index below Vivian had them as UINTN and I said if there isn’t a compelling reason to have them UINTN, they should be a fixed type. I think we see a lot of weirdness around UINTN, generally not for a good reason. But, I defer to your judgement

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
os-d: This I agree though, that since Index (or i) is being used in pointer arithmetic, it makes sense to have as UINTN and then since it is being compared to NumberOfDeviceStates that might as well be UINTN, too. Sorry for the noise, Vivian.

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
makubacki: Yeah, `UINTN` is generally best for counts and indices because those tend to get used in pointer arithmetic and represent values that scale with pointer width. For example, `UINTN` on a 64-bit machine can accurately represent the number of bytes for a valid offset where that offset can be larger than `MAX_UINT32`.

There are of course cases where smaller integers make sense. Especially in firmware, we have fixed size fields in data structures irrespective of the current architecture. In that case, the maximum allowable value is constrained by the space allocated for the data rather than the architecture.

In both cases, it can simplify casting and result in overall better integer handling to apply the appropriate type for the use case. In this case, I see count and indexing into an array being used which is classically better handled with `UINTN`. And you usually need the count and index type to match for their comparison in the loop to be safe.

MdeModulePkg/Include/Library/DeviceStateLib.h
VivianNK: Changed the function prototype to return EFI_STATUS. AddDeviceState uses RETURN_STATUS but is an EFIAPI - should that be changed as well?

Also, what do you think of the IN/OUT UINTN* for StringSize? Presumably the caller can just call AsciiStrnSizeS themselves?

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
VivianNK: Replaced with macro

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
VivianNK: Added a check but shouldn't this be guaranteed from the NumberOfDeviceStates?

MdeModulePkg/Library/DeviceStateLib/DeviceStateLib.c
VivianNK: Removed the function earlier and this was missed. Cleaned up now.

MdeModulePkg/Include/Library/DeviceStateLib.h
VivianNK: This would change the values 
1 << 20 instead of 20. 
I did make a change to consume the macros for the start and end mode macros... not sure if that helps. Open to suggestions.

spbrogan (changes requested):
I have some questions I would like to discuss before merging.

VivianNK:
Offline discussion determined it wasn't useful to measure other device states through this library.

PR #1187, opened 2024-10-17
[Rebase & FF] Fix PI status code

kuqin12:
## Description

We upstreamed the correct PI status code to PI spec and should get rid of MU_CHANGE to avoid conflict.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change only involves definition migration.

## Integration Instructions

For applicable components, platform needs to update these definitions in their code.

`EFI_SW_EC_MEMORY_TYPE_INFORMATION_CHANGE` -> `EFI_SW_EC_ILLEGAL_SOFTWARE_STATE`
`EFI_SW_EC_RELEASE_ASSERT` -> `EFI_SW_EC_ILLEGAL_SOFTWARE_STATE`

### DIFF
--- a/MdePkg/Include/Pi/PiStatusCode.h
+++ b/MdePkg/Include/Pi/PiStatusCode.h
@@ -967,29 +967,27 @@ typedef struct {
 /// These are shared by all subclasses.
 ///
 ///@{
-#define EFI_SW_EC_NON_SPECIFIC                    0x00000000
-#define EFI_SW_EC_LOAD_ERROR                      0x00000001
-#define EFI_SW_EC_INVALID_PARAMETER               0x00000002
-#define EFI_SW_EC_UNSUPPORTED                     0x00000003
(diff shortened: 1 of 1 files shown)


apop5:
This is back porting these changes into the 202311 branch.

Is there a benefit to bringing this back when 202405 has changed this?

Also, if this is going into 202311, do the (two?) places where these MU defines were used need to be changed in their respective 202311 branches as well?

kuqin12:
> This is back porting these changes into the 202311 branch.
> 
> Is there a benefit to bringing this back when 202405 has changed this?

yeah, the addition of these definitions after updating the spec is a mistake in the first place. The consumers are still using these proprietary definitions, which is making the pipeline receiving mixed value for the same events.

apop5 (approved):

VivianNK (approved):

PR #1199, opened 2024-10-26
[CHERRY-PICK] .pytool/UncrustifyCheck: Show errors in output

makubacki:
## Description

Shows code formatting errors directly in build output. Previously only the filenames were in build output and the user had to look at the test result file either locally or in CI to find details.

It is still recommended that users configure their local environment to run Uncrustify so it can automatically fix problems as opposed to manually correcting code based on the output shown in the terminal. In any case, it is easier to see what is expected now.

**Uncrustify reference material:**

https://github.com/tianocore/tianocore.github.io/wiki/EDK-II-Code-Formatting https://github.com/tianocore/edk2/tree/master/.pytool/Plugin/UncrustifyCheck#readme

Some logging levels are also updated to refocus log output by current message importance and relevance.

(cherry picked from commit 10783187ddb71808a4ea361e887800d3b899b85f)

---

**Before**

image

**After**

image

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Verified log output with uncrustify errors 

## Integration Instructions

- N/A - Developers may find this more convenient

### DIFF
--- a/.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
+++ b/.pytool/Plugin/UncrustifyCheck/UncrustifyCheck.py
@@ -384,9 +384,9 @@ class UncrustifyCheck(ICiBuildPlugin):
                 file_template_path = pathlib.Path(os.path.join(self._plugin_path, file_template_name))
                 self._file_template_contents = file_template_path.read_text()
         except KeyError:
-            logging.warning("A file header template is not specified in the config file.")
+            logging.info("A file header template is not specified in the config file.")
         except FileNotFoundError:
-            logging.warning("The specified file header template file was not found.")
+            logging.info("The specified file header template file was not found.")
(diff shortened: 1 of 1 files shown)


os-d (approved):
Do you want to revert the old commit and then cherry-pick this one?

Unless you think cleansing fire will show the original commit as 0 lines without reverting.

makubacki:
> Do you want to revert the old commit and then cherry-pick this one?
> 
> Unless you think cleansing fire will show the original commit as 0 lines without reverting.

I'll revert and add the new one. I think it is better to resolve that upfront.

apop5 (approved):

Javagedes (approved):

PR #1246, opened 2024-12-18
[CHERRY-PICK] UefiCpuPkg: x86 CpuDxe: Allocate AP Exception Stack Below 4GB

os-d:
## Description

When setting up the APs' exception stacks, the x86 CpuDxe allocates any range and then copies over the existing GDT and IDT and adds the appropriate new entries for this AP, then installs them.

This can cause an issue if the allocated buffer is over 4GB because the next time the AP is started, it goes through an INIT-SIPI-SIPI, stepping through real mode -> protected mode -> long mode and when it is in protected mode it needs a 32 code segment descriptor or else it will fault when trying to execute. If the GDT lives above 4GB, it cannot be accessed by the protected mode code and the triple fault is seen.

This patch updates CpuDxe's MP management code to allocate the exception stacks for all APs below 4GB explicitly to avoid this problem, such as it does with the BSP's GDT that first gets populated to the APs.

Cherry-picked from: https://github.com/tianocore/edk2/commit/924780f2ccb5d4bed09d7b79a85817f15906a608

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a platform that had allocated the AP exception stacks above 4GB and so triple faulted. It booted after this patch.

## Integration Instructions

N/A.

### DIFF
--- a/UefiCpuPkg/CpuDxe/CpuMp.c
+++ b/UefiCpuPkg/CpuDxe/CpuMp.c
@@ -663,7 +663,12 @@ InitializeMpExceptionStackSwitchHandlers (
   UINT8                           *Buffer;
 
   SwitchStackData = AllocateZeroPool (mNumberOfProcessors * sizeof (EXCEPTION_STACK_SWITCH_CONTEXT));
-  ASSERT (SwitchStackData != NULL);
+  if (SwitchStackData == NULL) {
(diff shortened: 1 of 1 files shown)


makubacki (approved):

apop5 (approved):

vincent-j-zimmer (approved):
This change looks reasonable.

PR #1258, opened 2025-01-22
Add resource descriptor hob v2 for memory attributes usage

kuqin12:
## Description

This change adds the implementation of supporting "resource descriptor hob type 2", which will extends the existing definition of resource descriptor hob with an attribute field.

This field can be used to provide the actual attributes needs to be set for the memory region, instead of relying on the state of the system.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU Q35 and SBSA systems and booted to Windows.

## Integration Instructions

For platforms needing to propagate specific memory attributes (especially the cache attributes) to the DXE phase, `BuildResourceDescriptorWithCacheHob` should be used to produce resource descriptors (v2).

### DIFF
--- a/MdePkg/Include/Library/HobLib.h
+++ b/MdePkg/Include/Library/HobLib.h
@@ -205,6 +205,7 @@ BuildResourceDescriptorWithOwnerHob (
   );
 
 // MU_CHANGE Start: Add BuildResourceDescriptorV2 function
+
 /**
   Builds a HOB that describes a chunk of system memory with memory attributes.
 
@@ -232,6 +233,7 @@ BuildResourceDescriptorV2 (
   IN UINT64                       EfiMemoryAttributes,
   IN EFI_GUID                     *OwnerGUID OPTIONAL
   );
+
 // MU_CHANGE End
 
 /**
--- a/MdePkg/Include/Library/HobLib.h
+++ b/MdePkg/Include/Library/HobLib.h
@@ -205,6 +205,32 @@ BuildResourceDescriptorWithOwnerHob (
   );
 
 /**
+  Builds a HOB that describes a chunk of system memory with memory attributes.
+
+  This function builds a HOB that describes a chunk of system memory.
+  It can only be invoked during PEI phase;
+  for DXE phase, it will ASSERT() since PEI HOB is read-only for DXE phase.
+
+  If there is no additional space for HOB creation, then ASSERT().
+
+  @param  ResourceType        The type of resource described by this HOB.
+  @param  ResourceAttribute   The resource attributes of the memory described by this HOB.
+  @param  PhysicalStart       The 64 bit physical address of memory described by this HOB.
+  @param  NumberOfBytes       The length of the memory described by this HOB in bytes.
+  @param  Attributes          The memory attribute for the memory described by this HOB.
+
+**/
+VOID
+EFIAPI
+BuildResourceDescriptorWithCacheHob (
+  IN EFI_RESOURCE_TYPE            ResourceType,
+  IN EFI_RESOURCE_ATTRIBUTE_TYPE  ResourceAttribute,
+  IN EFI_PHYSICAL_ADDRESS         PhysicalStart,
+  IN UINT64                       NumberOfBytes,
+  IN UINT64                       Attributes
+  );
+
+/**
   Builds a HOB that describes a chunk of system memory.
 
   This function builds a HOB that describes a chunk of system memory.
--- a/MdePkg/Include/Library/HobLib.h
+++ b/MdePkg/Include/Library/HobLib.h
@@ -219,6 +219,7 @@ BuildResourceDescriptorWithOwnerHob (
   @param  PhysicalStart         The 64 bit physical address of memory described by this HOB.
   @param  NumberOfBytes         The length of the memory described by this HOB in bytes.
   @param  EfiMemoryAttributes   The memory attribute for the memory described by this HOB.
+  @param  OwnerGUID             GUID for the owner of this resource.
 
 **/
 VOID
--- a/MdePkg/Include/Library/HobLib.h
+++ b/MdePkg/Include/Library/HobLib.h
@@ -204,6 +204,7 @@ BuildResourceDescriptorWithOwnerHob (
   IN EFI_GUID                     *OwnerGUID
   );
 
+// MU_CHANGE Start: Add BuildResourceDescriptorV2 function
 /**
   Builds a HOB that describes a chunk of system memory with memory attributes.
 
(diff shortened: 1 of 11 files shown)


MdePkg/Include/Library/HobLib.h
```diff
@@ -204,6 +204,32 @@ BuildResourceDescriptorWithOwnerHob (
...
+  It can only be invoked during PEI phase;
+  for DXE phase, it will ASSERT() since PEI HOB is read-only for DXE phase.
+
+  If there is no additional space for HOB creation, then ASSERT().
+
+  @param  ResourceType        The type of resource described by this HOB.
+  @param  ResourceAttribute   The resource attributes of the memory described by this HOB.
+  @param  PhysicalStart       The 64 bit physical address of memory described by this HOB.
+  @param  NumberOfBytes       The length of the memory described by this HOB in bytes.
+  @param  Attributes          The memory attribute for the memory described by this HOB.
```
os-d:
```suggestion
  @param  EfiMemoryAttributes          The memory attribute for the memory described by this HOB.

```

Just to help distinguish what these attributes are vs the ResourceAttributes.

MdePkg/Include/Library/HobLib.h
```diff
@@ -204,6 +204,32 @@ BuildResourceDescriptorWithOwnerHob (
...
+
+**/
+VOID
+EFIAPI
+BuildResourceDescriptorWithCacheHob (
```
os-d:
```suggestion
BuildResourceDescriptorWithEfiMemoryAttributesHob (

```

Maybe something besides this, but trying to be a bit clearer on what this is creating, WithCache both is confusing (are you caching the hob, etc.) and maybe too restrictive, it is cache attributes for now, could be extended in the future.

MdePkg/Include/Pi/PiHob.h
```diff
@@ -332,6 +333,25 @@ typedef struct {
...
+///
+typedef struct {
+  ///
+  /// The HOB generic header. Header.HobType = EFI_HOB_TYPE_RESOURCE_DESCRIPTOR.
+  ///
+  EFI_HOB_RESOURCE_DESCRIPTOR    V1;
+  ///
+  /// The memory attributes (paging and caching) of the resource region.
+  ///
+  UINT64                         Attributes;
```
os-d, line 352:
```suggestion
  UINT64                         EfiMemoryAttributes;

```

Again, doesn't have to be this, but something to clarify what this is vs ResourceAttributes.

MdePkg/Include/Pi/PiHob.h
```diff
@@ -332,6 +333,25 @@ typedef struct {
   UINT64                         ResourceLength;
 } EFI_HOB_RESOURCE_DESCRIPTOR;
 
+/// PI Spec Status: Pending.
+/// This change is checked in as a code first approach. The PI spec will be updated
+/// to reflect this change in the future.
+///
+/// Describes the resource properties of all fixed,
+/// nonrelocatable resource ranges found on the processor
+/// host bus during the HOB producer phase.
```
os-d: This is the same comment as v1, probably should add a sentence on what distinguishes v2.

MdePkg/Library/PeiHobLib/HobLib.c
```diff
@@ -342,6 +343,48 @@ BuildResourceDescriptorWithOwnerHob (
...
+  @param  ResourceType        The type of resource described by this HOB.
+  @param  ResourceAttribute   The resource attributes of the memory described by this HOB.
+  @param  PhysicalStart       The 64 bit physical address of memory described by this HOB.
+  @param  NumberOfBytes       The length of the memory described by this HOB in bytes.
+  @param  Attributes          The memory attribute for the memory described by this HOB.
+
+**/
+VOID
+EFIAPI
+BuildResourceDescriptorWithCacheHob (
```
os-d: need WithOwner variant?

os-d (approved):

MdePkg/Include/Library/HobLib.h
apop5: Is the WIthCacheHob attempting to say that only coachability attributes are being reported, or that the hob is in cache?

Does this also encompass the bits covered by EFI_MEMORY_ACCESS_MASK and EFI_MEMORY_ATTRIBUTE_MASK?

apop5 (approved):

MdePkg/Include/Library/HobLib.h
kuqin12: Updated. Thanks.

MdePkg/Include/Pi/PiHob.h
kuqin12: Updated.

MdePkg/Library/PeiHobLib/HobLib.c
kuqin12: I renamed the interface to be build rsc desc v2, which will take both the attribute and the guid. But the guid is optional.

MdePkg/Include/Library/HobLib.h
kuqin12: Updated the name to be v2. For the current implementation, C core will not act on these hobs different than the v1 hobs.

MdePkg/Library/PeiHobLib/HobLib.c
```diff
@@ -342,6 +343,55 @@ BuildResourceDescriptorWithOwnerHob (
...
+  for DXE phase, it will ASSERT() since PEI HOB is read-only for DXE phase.
+
+  If there is no additional space for HOB creation, then ASSERT().
+
+  @param  ResourceType          The type of resource described by this HOB.
+  @param  ResourceAttribute     The resource attributes of the memory described by this HOB.
+  @param  PhysicalStart         The 64 bit physical address of memory described by this HOB.
+  @param  NumberOfBytes         The length of the memory described by this HOB in bytes.
+  @param  EfiMemoryAttributes   The memory attribute for the memory described by this HOB.
+
```
makubacki, line 363: `OwnerGUID` param description is missing.

makubacki (approved):

MdePkg/Library/PeiHobLib/HobLib.c
kuqin12: Added. Nice call.

MdePkg/Include/Pi/PiHob.h
kuqin12: Updated.

PR #1293, opened 2025-03-06
[SQUASH_ON_REBASE] Remove unused variable in DEBUG_CLEAR_MEMORY call

kenlautner:
## Description

In project MU we changed the DEBUG_CLEAR_MEMORY define to do nothing.  This causes an issue in the new file MemoryAllocationLibPosix.c which calls it with a newly defined variable.  To get around this variable is removed and instead we use what it was deriving its value from instead.

This should be merged with [230503 ](https://github.com/microsoft/mu_basecore/commit/234050304aa77dd1e7e7164b1e7c5074a063c566) on a rebase

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on the pipelines.  No issues

## Integration Instructions

N/A

### DIFF
--- a/UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.c
+++ b/UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.c
@@ -274,7 +274,8 @@ FreeAlignedPages (
 {
   PAGE_HEAD  *PageHeadPtr;
   VOID       *AllocatedBuffer;
-  //UINTN      Length;  MU_CHANGE
+
+  // UINTN      Length;  MU_CHANGE
 
   ASSERT (Buffer != NULL);
 
@@ -288,6 +289,7 @@ FreeAlignedPages (
   AllocatedBuffer = PageHeadPtr->AllocatedBuffer;
 
   // MU_CHANGE [START]: Remove variable that isn't used in project MU
+
   /* Length          = EFI_PAGES_TO_SIZE (PageHeadPtr->TotalPages);
 
   DEBUG_CLEAR_MEMORY (AllocatedBuffer, Length); */
--- a/UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.c
+++ b/UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.c
@@ -274,7 +274,7 @@ FreeAlignedPages (
 {
   PAGE_HEAD  *PageHeadPtr;
   VOID       *AllocatedBuffer;
-  UINTN      Length;
+  //UINTN      Length;  MU_CHANGE
 
   ASSERT (Buffer != NULL);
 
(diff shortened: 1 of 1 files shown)


kenlautner:
@os-d are you okay with this type of change surrounding DEBUG_CLEAR_MEMORY or would you rather the macro have dummy functionality so we don't run into this problem in the future?

zurcher (approved):

os-d:
> @os-d are you okay with this type of change surrounding DEBUG_CLEAR_MEMORY or would you rather the macro have dummy functionality so we don't run into this problem in the future?

I think just doing this here for now is okay, if it crops up again, we can look at it further. I would mark this [SQUASH_ON_REBASE] and in the commit message point to https://github.com/microsoft/mu_basecore/commit/234050304aa77dd1e7e7164b1e7c5074a063c566 to be squashed with.

I am starting to get some of the memory protections work upstreamed, so that should hopefully resolve this issue from spreading.

However, we probably have broken the ASAN on tests, or at least part of it, as it seems to be relying on this functionality.

os-d (approved):

UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.c
```diff
@@ -286,9 +287,14 @@ FreeAlignedPages (
...
 
   AllocatedBuffer = PageHeadPtr->AllocatedBuffer;
-  Length          = EFI_PAGES_TO_SIZE (PageHeadPtr->TotalPages);
 
-  DEBUG_CLEAR_MEMORY (AllocatedBuffer, Length);
+  // MU_CHANGE [START]: Remove variable that isn't used in project MU
+
+  /* Length          = EFI_PAGES_TO_SIZE (PageHeadPtr->TotalPages);
+
+  DEBUG_CLEAR_MEMORY (AllocatedBuffer, Length); */
```
makubacki, line 295: I might be missing some context but I assume there was a reason `DEBUG_PROPERTY_CLEAR_MEMORY_ENABLED` in `PcdDebugProperyMask` wasn't used to disable the functionality in the original macro implementation?

zurcher: I don't see any comments around that change in #768 but I also can't find any other code actually checking that property

os-d: This was dropped for two reasons, one it broke with RP on free memory because it would try to clear free memory which was now unmapped. But it had also been on our radar to drop because you see bugs that only exist in release and not debug because of this macro, due to memory being cleared in one scenario and not the other.

So the desire was to have the core completely ignore the PCD and not let a platform set it because it is deceiving and will break the platform.

The UnitTestFrameworkPkg one, for supporting ASAN, should probably just be updated to clear memory through a different mechanism.

makubacki: On a side note, I just realized it is spelled incorrectly (`PcdDebugProperyMask` instead of `PcdDebugPropertyMask`) in a lot of files that led to the mispelling in my initial message.

---

The PCD is the public configuration interface and so now it feels slightly deceptive that it's not hooked up to what you'd expect it to be. 

I don't think that would be such an issue if the [PCD definition in MdePkg.dec](https://github.com/microsoft/mu_basecore/blob/f82d14c76e8417abd010e3983346fb3ac4f82fe5/MdePkg/MdePkg.dec#L2354) were updated to reflect how the value is used in Project Mu and maybe if there was something like a build assert (or runtime) assert that warned the developer is the bit is set.

os-d: Yeah, I agree. I think this was something that was done in the final days before Taylor left and was a, well, let's get everything in before he goes :)

makubacki: Doesn't impact this change. I'll file a tracking issue over in mu_basecore for that.

makubacki (approved):

UnitTestFrameworkPkg/Library/Posix/MemoryAllocationLibPosix/MemoryAllocationLibPosix.c
makubacki: https://github.com/microsoft/mu_basecore/issues/1294

PR #1297, opened 2025-03-12
Add BuildReportEvaluation post build plugin

apop5:
## Description

Add a post build plugin that will parse the build report and run evaluation queries over it.

Infrastructure allows for platforms to register additional queries, either under the existing categories or creating their own.

See BuildReportEvaluationReadMe.md for more details.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Ran on virtual platforms builds with scope enabled. 
Ran on physical platform build with scope enabled. 

## Integration Instructions
Add platform-checks scope.
Ensure that BUILDREPORT_FILE is enabled and configured.


.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
```diff
@@ -0,0 +1,265 @@
...
+#  Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+from edk2toollib.database.edk2_db import Edk2DB
+from buildreport_table import BrComponent, BrPcd, BrLibrary
+import logging
+
+
+def CompiledButButNotIncludedInAnyFV(edk2db: Edk2DB, reportlevel: str) -> int:
```
makubacki: It is common in a platform with various build "flavors" to build drivers that do not end up in the FV to ensure the overall build is not broken due to modifying a common library even though the driver does not get placed in the flash image for the build.

It seems like this is a natural set of "false positives" that would be nice to eliminate given the motivation for this check.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
```diff
@@ -0,0 +1,265 @@
...
+        reportlevel (str): The level of reporting.
+    Returns:
+        int: The number of occurrences of 'SetupDxe' in the FV.
+    """
+    error_count = 0
+    with edk2db.session() as session:
+        results = (
+            session.query(BrComponent)
+            .filter(BrComponent.fv_name != None)
+            .filter(BrComponent.path.contains("SetupDxe"))
```
makubacki: I think referencing specific drivers outside its package is leaking code outside the scope it should be contained in (the package). Based on this code alone, it is not clear where `SetupDxe` was originally flagged from and why this error would occur if someone added a new `SetupDxe` driver to their FV.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
makubacki, line 1: It seems these queries should be defined in a more declarative (e.g. JSON) than procedural (Python) format. That would more succinctly describe what needs to be done, abstract the actions from code implementation details allowing the data declaration to stay constant over time with underlying changes in `Edk2DB` and reduce likelihood of error in data input.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
```diff
@@ -0,0 +1,265 @@
...
+    Returns:
+        int: The count of occurrences where PcdConOutUgaSupport is not zero.
+    """
+    error_count = 0
+
+    with edk2db.session() as session:
+        results = (
+            session.query(BrComponent)
+            .join(BrPcd)
+            .filter(BrPcd.token == "PcdConOutUgaSupport" and BrPcd.value == "1")
```
makubacki: Similar to driver names leaking globally, I think something needs to be done to constrain the PCD. At least including the PCD token space GUID in addition to the name (which is here).

I think this rolls out into two generic requests: (1) be able to map this PCD to the exact instance it is checking with the full two-part definition (token space GUID and name) (2) do not check for this PCD in builds that are not using the instance the rule was written for.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
```diff
@@ -0,0 +1,265 @@
...
+            .filter(BrComponent.fv_name is not None)
+            .filter(BrComponent.type != "APPLICATION")
+            .filter(BrComponent.type != "USER_DEFINED")
+            .all()
+        )
+
+        for result in results:
+            stack_found = False
+            for lib in result.libraries:
+                if "StackCheckLib" in lib.path:
```
makubacki: Similar to previous comments, this code is in `.pytool/Plugin` and `StackCheckLib` is an implementation detail outside its scope in the overall workspace hierarchy. If the library class were renamed, removed, etc. the same code owner should be responsible for updating the rule in their ownership path in the workspace hierarchy as opposed to the library class name being disconnected in this global Python code and falling onto the responsibility of the `.pytool` owner.

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationPlugin.py
makubacki, line 1: I suggest considering a design that allows an independent CLI to be used to run queries against a build report and simply drive that from the plugin. Not necessarily through the CLI but the CLI or the Python module interface behind the CLI.

apop5: I agree with this, I'll look into making changes.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
apop5: There are cases where a platform compiles something, but does not use it intentionally.

This is intended to alert the platform owner that they are doing this, and they should have the ability to ignore the warning message that is generated. Highlighting that this is occurring is the goal, and the platform owner has the ability to ignore the warning message.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
apop5: Agreed. Removed this from the public implementation and add this to the appropriate place for platform owners who need it.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
apop5: The resulting json files has improved the check to make sure its the gEfiMdeModulePkgTokenSpaceGuid instance that this maps to.

Since Uga has essentially been deprecated since uefi 2.0 when GOP was introduced, this should be acceptable for platforms being compiled today.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
apop5: New version has switched to using a json files for the queries.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
apop5: With StackCheck being integrated into Edk2, I think this is less of an implementation detail, and more of a security check.

Being alerted to any modules compiled that do not contain a StackCheckLib instance is useful information for a platform owner. Once they are alerted, they can investigate and see if there is corrective action which needs to take place.

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationPlugin.py
apop5: CLI has been implemented in latest version.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
makubacki: If it's clear to them what they need to do, that sounds reasonable.

makubacki (approved):

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
makubacki: Can you add some detail around this in the warning message?

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
makubacki: I'm not arguing against it being useful for a platform owner or that they should not know about it automatically but that it would be more extensible to allow package-specific checks like this to be picked from up from a set of rules defined in the corresponding package (`MdePkg`) instead of directly coded in the .pytool Python file. If you'd like to go down this path, that's okay too for what's in here right now but it's not as extensible over time.

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationPlugin.py
```diff
@@ -0,0 +1,202 @@
...
+        ],
+        help="""Specify the queries json file to be used to evaluate the build report. ignored otherwise.""",
+    )
+    arguments = parser.parse_args()
+    arguments.WorkSpace = os.path.abspath(arguments.WorkSpace)
+
+    return arguments
+
+
+def generate_db(env, pathtool):
```
Javagedes: nit: needs typing specified

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationPlugin.py
```diff
@@ -0,0 +1,202 @@
...
+            db = generate_db(env, builder.edk2path)
+
+            evaluate_build_report_json(db, queries_json)
+            return 0
+
+except ImportError:
+    pass
+
+
+def arguments():
```
Javagedes: nit: needs typing specified

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationReadMe.md
```diff
@@ -0,0 +1,49 @@
...
+## Expanding
+
+Platforms can generate additonal queries that should be run by creating their own json file in the format of
+    {
+        "query": "<query to run against the database>",
+        "warning_message": "<Warning message to be displayed when modules fail query>",
+        "helpful_description": "<Message that describes why the query is important>"
+    },
+
+Queries are expected to be SQL, and buildreport_table can be examined to see the tables generated.
```
Javagedes, line 44: You could specify here that if they want to keep the existing reports and just add to it, then they would do something (untested) like

``` python
br = "BUILD_REPORT_EVALUATION_JSONS"
extra = []
self.env.SetValue(br, self.env.GetValue(br, []) + extra, "Override Default jsons")
```

.pytool/Plugin/BuildReportEvaluation/queries.json
```diff
@@ -0,0 +1,27 @@
...
+        "query": "SELECT br_component.path FROM br_component WHERE fv_name IS NOT NULL AND path LIKE '%EbcDxe/EbcDxe.inf';",
+        "warning_message": "EBC driver is contained in the project.",
+        "helpful_description": "EBC (Efi Byte Code) is an emulator for EBC option roms. EBC oproms are not secure boot compatiable."
+    },
+    {
+        "query": "SELECT br_component.path FROM br_component WHERE fv_name IS NOT NULL AND type != 'FREE_FORM' AND br_component.guid NOT IN (SELECT br_library.guid FROM br_library WHERE br_library.cls = 'StackCheckLib');",
+        "warning_message": "Module included in Fv that does not use a StackCheck Library Instance",
+        "helpful_description": "StackCheckLib is used to protect execution against stack corruption during runtime. If a module does not contain a StackCheckLib instance, it may be vulnerable to stack corruption attacks."
+    }
+]
```
Javagedes: nit: needs newline

Javagedes (approved):

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationPlugin.py
apop5: Fixed in the latest.

.pytool/Plugin/BuildReportEvaluation/BuildReportEvaluationPlugin.py
apop5: Fixed in the latest.

.pytool/Plugin/BuildReportEvaluation/queries.json
apop5: Fixed in the latest.

.pytool/Plugin/BuildReportEvaluation/BuildReportQueries.py
apop5: I agree with the sentiment, kind of how the codeql filters are implemented in the per repo method.

In practical terms, I attempting to split the queries up comes with its own set of problems. So many of them are on the platform decision level that it is hard to create a cohesive set of rules that doesn't bleed into the implementation of other packages/instances of libraries.

PR #1164, opened 2024-09-26
[Rebase & FF] Bring in changes from release/202405

VivianNK:
## Description
MU repos recently transitioned to using a dev branch for new changes after a release. In the process of transition, a few commits got checked in to release/202405. 
 
This PR brings in the following commits from release/202405:

91a25384a89fa68a29193eeeecb9aa4db62ac6f3
88b6c1e50013f8270ec5f0435e191f0d83e4c203
95a58f8ff1baa410405b6b0c56cb4666b7c23364
ba672c3aca5c96e493fbf3628ee779388d82634b

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested in release/202405

## Integration Instructions

N/A

PR #1313, opened 2025-03-25
[REBASE & FF] BaseTools: Update tools_def for AARCH64

VivianNK:
## Description

Update tools def for aarch64 for new 202502 build tools changes.
/WX was added in 202502 as a build flag and now Aarch64 builds
fail. The first failure is because of an undocumented linker warning: 4226 Alignment specified exceeds target machine page size. The second is because the minimum alignment for MSVC ARM64 is 64.

Changes made:
- Ignore Linker Warning LNK4226. This linker warning has always occurred and can be ignored. 
- Change Aarch64 /ALIGN value from 32 to 64 to fix Linker error `#1164 from BaseLib.lib(CpuBreakpoint.obj) : fatal section 0x1 alignment (64) greater than /ALIGN value`. 64 is the minimum alignment for MSVC ARM64, which differs from MSVC x64 and was missed when checking into edk2 because ARM64 MSVC CI isn't ran.
- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI and platform build with aarch64

## Integration Instructions

N/A
To be upstreamed to edk2


os-d (approved):
This is good to get in basecore once you have decided this is the set of changes needed for booting with this toolchain, but then we should upstream to edk2.

apop5 (approved):

BaseTools/Conf/tools_def.template
makubacki, line 1: The commit messages need additional context about each change being made. Root cause for the change and upstream plan.

makubacki: Also, the PR template should be filled out to roll into the release notes.

os-d: I would update the first commit message to indicate that this linker warning has always occurred and can be ignored.

os-d: And the second one I would update to indicate that `/ALIGN:64` is the minimum alignment for MSVC ARM64, which differs from MSVC x64 and that MSVC ARM64 CI is not run in edk2, so this was missed when checking in to edk2.

os-d (approved):

makubacki (approved):

PR #1317, opened 2025-03-29
Add CPU_INFORMATION_HEADER struct to STM_HEADER

kenlautner:
## Description

Created a new struct called CPU_INFORMATION_HEADER that is now being included in the STM_HEADER.  It siphons space from the MSEG_HEADER which has a significant amount of padding.  Currently CPU_INFORMATION_HEADER only contains a UINT32 representing the number of CPUs on the system but extra space is reserved for future fields.

Note that this reduces the MSEG_HEADER size to 1KB and the CPU_INFORMATION_HEADER to 1KB.  This is to maintain the 4KB size requirement for the STM_HEADER.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Intel physical system that uses an STM.  The system is able to operate with no issues with the change.

## Integration Instructions

N/A


MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -52,10 +52,19 @@ typedef struct {
...
-  MSEG_HEADER            HwStmHdr;
-  SOFTWARE_STM_HEADER    SwStmHdr;
+  UINT32    NumberOfCpus;
+  /// Pad to take up 1KB of space
+  UINT8     Reserved[SIZE_1KB - sizeof (UINT32)];
+} CPU_INFORMATION_HEADER;
+
+typedef struct {
+  MSEG_HEADER               HwStmHdr;
+  CPU_INFORMATION_HEADER    CpuInfoHdr;
```
kuqin12: I thought we are going to add this after the software header?

makubacki: `MSEG_HEADER` ends with `UINT8     Reserved[SIZE_2KB - 8 * sizeof (UINT32)];`, `SOFTWARE_STM_HEADER` is expected to immediately follow at that offset. This would have to come after it to make the change `STM_HEADER` backward compatible.

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -52,10 +52,19 @@ typedef struct {
...
-  MSEG_HEADER            HwStmHdr;
-  SOFTWARE_STM_HEADER    SwStmHdr;
+  UINT32    NumberOfCpus;
+  /// Pad to take up 1KB of space
+  UINT8     Reserved[SIZE_1KB - sizeof (UINT32)];
+} CPU_INFORMATION_HEADER;
+
+typedef struct {
+  MSEG_HEADER               HwStmHdr;
+  SOFTWARE_STM_HEADER       SwStmHdr;
```
makubacki, line 74: The order looks good now, but you're going to have to add reserved padding to the end of `SOFTWARE_STM_HEADER` so it is exactly 1KB in size and `CPU_INFORMATION_HEADER` begins at a 1KB boundary.

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -52,10 +52,19 @@ typedef struct {
   ///
 } SOFTWARE_STM_HEADER;
 
+// MU_CHANGE - [START]: Add new CPU information header for STM
 typedef struct {
-  MSEG_HEADER            HwStmHdr;
-  SOFTWARE_STM_HEADER    SwStmHdr;
+  UINT32    NumberOfCpus;
```
makubacki, line 67: I think it'd be best to add a unique signature to the beginning of this structure. Then, a consumer could generically test for it at MSEG header + 2k to see if this structure is present.

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -15,11 +15,14 @@
...
 #include <Register/Intel/StmResourceDescriptor.h>
 #include <Register/Intel/ArchitecturalMsr.h>
+#include <Base.h>
 
 #pragma pack (1)
 
 // MU_CHANGE - Added SMM_REV_ID definition, according to STM spec
 #define STM_SMM_REV_ID  0x80010100
+// MU_CHANGE - Added signature for CPU_INFORMATION_HEADER
+#define CPU_INFORMATION_HEADER_SIGNATURE  SIGNATURE_32('C', 'p', 'u', 'I')
```
makubacki: Because this is C, it'd be better to prefix the macro to scope it.

```suggestion
#define STM_CPU_INFORMATION_HEADER_SIGNATURE  SIGNATURE_32('C', 'p', 'u', 'I')
```

MdePkg/Include/Register/Intel/ArchitecturalMsr.h
```diff
@@ -517,7 +517,7 @@ typedef struct {
   ///
   /// Pad header so total size is 2KB
   ///
-  UINT8     Reserved[SIZE_2KB - 8 * sizeof (UINT32)];
+  UINT8     Reserved[SIZE_1KB - 8 * sizeof (UINT32)];  // MU_CHANGE - Reduce padding
```
kuqin12: Correct me if I am missing something, but I think we should leave this intact?

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -50,12 +53,25 @@ typedef struct {
...
 } SOFTWARE_STM_HEADER;
 
+// MU_CHANGE - [START]: Add new CPU information header for STM
 typedef struct {
-  MSEG_HEADER            HwStmHdr;
-  SOFTWARE_STM_HEADER    SwStmHdr;
+  UINT32    Signature;
+  UINT32    NumberOfCpus;
+  /// Pad to take up 1KB of space
+  UINT8     Reserved[SIZE_2KB - 2 * sizeof (UINT32)];
```
kuqin12: Shouldn't this be 1KB - 2 UINT32?

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -50,12 +53,25 @@ typedef struct {
   ///
   /// The total STM_HEADER should be 4K.
   ///
+  
+  // Pad to take up 1KB of space
+  UINT8     Reserved2[SIZE_1KB - 5 * sizeof (UINT32) - 3 * sizeof (UINT8) - sizeof (UINT16) - sizeof (STM_FEAT)];
```
makubacki: I think the best way to deal with these is to statically verify them against the compiler generated value.

For example:

```c
#define SOFTWARE_STM_HEADER_PADDING_SIZE (SIZE_1KB - 5 * sizeof (UINT32) - 3 * sizeof (UINT8) - sizeof (UINT16) - sizeof (STM_FEAT))

// ... Later in the struct definition...

  ///
  /// The total STM_HEADER should be 4K.
  ///

  // Pad to take up 1KB of space
  UINT8     Reserved2[SOFTWARE_STM_HEADER_PADDING_SIZE];
} SOFTWARE_STM_HEADER;

// Afterward, confirm...

STATIC_ASSEERT ((SIZE_1KB - OFFSET_OF (SOFTWARE_STM_HEADER, Reserved2) == SOFTWARE_STM_HEADER_PADDING_SIZE, "SOFTWARE_STM_HEADER_PADDING_SIZE mismatch!");

// Continue on...
```

makubacki: Should this only be `... 2 * sizeof (UINT8) ...`?

MdePkg/Include/Register/Intel/ArchitecturalMsr.h
makubacki: Yeah for binary compatibility the Software STM header needs to start at the same offset (2k).

MdePkg/Include/Register/Intel/StmApi.h
kuqin12: Would you want to make sure the 3 fields are 2KB, 1KB, and 1KB? Just to avoid touching too many details of the members inside those sub-structures?

makubacki: I think that makes sense to leave the original type definitions matching the upstream.

kenlautner: You're correct.

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -55,9 +56,25 @@ typedef struct {
...
-  UINT8     Reserved2[SIZE_1KB - 5 * sizeof (UINT32) - 3 * sizeof (UINT8) - sizeof (UINT16) - sizeof (STM_FEAT)];
+  UINT8 Reserved2[SOFTWARE_STM_HEADER_PADDING_SIZE];
 } SOFTWARE_STM_HEADER;
 
+// MU_CHANGE - [START]: Add new CPU information header for STM
+typedef struct {
+  UINT32    Signature;
+  UINT32    NumberOfCpus;
+  /// Pad to take up 1KB of space
+  UINT8     Reserved[SIZE_1KB - 2 * sizeof (UINT32)];
```
makubacki: It would be cleaner to also place this calculation into a macro and `STATIC_ASSERT()` it, but not as important since it is simpler.

makubacki (dismissed):

MdePkg/Include/Register/Intel/StmApi.h
kenlautner: That makes sense.  Updated.

MdePkg/Include/Register/Intel/ArchitecturalMsr.h
kenlautner: I reverted the change based on discussions.

MdePkg/Include/Register/Intel/StmApi.h
kenlautner: That's what it currently is.  Also it moved to a define above.

MdePkg/Include/Register/Intel/StmApi.h
kenlautner: Good idea.  I moved the padding calculations to defines as suggested and now they're being validated with STATIC_ASSERTs.

MdePkg/Include/Register/Intel/StmApi.h
kenlautner: Updated.

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -50,13 +58,29 @@ typedef struct {
...
+  UINT8     Reserved[CPU_INFORMATION_HEADER_PADDING_SIZE];
+} CPU_INFORMATION_HEADER;
+
+typedef struct {
+  MSEG_HEADER               HwStmHdr;
+  SOFTWARE_STM_HEADER       SwStmHdr;
+  CPU_INFORMATION_HEADER    CpuInfoHdr;
 } STM_HEADER;
 
+STATIC_ASSERT ((SIZE_1KB - OFFSET_OF (CPU_INFORMATION_HEADER, Reserved)) == CPU_INFORMATION_HEADER_PADDING_SIZE, "CPU_INFORMATION_HEADER_PADDING_SIZE mismatch!");
```
kuqin12, line 80: nit: would you want to do `OFFSET_OF (STM_HEADER, SwStmHdr) = 1KB` instead? or just check the struct size of SOFTWARE_STM_HEADER to be 1KB?

kenlautner: There are several ways to verify that we keep the headers at 1KB of size.  I'm fine with whatever we think is best.

makubacki: As long as the macro stays that holds the padding size, I'm also fine either way. Technically, we'd want to do both - `sizeof()` to check the total expected size including padding for each structure and `OFFSET_OF()` to check that the order of structs is maintained in `STM_HEADER`.

MdePkg/Include/Register/Intel/StmApi.h
```diff
@@ -34,6 +36,16 @@ typedef struct {
 #define STM_SPEC_VERSION_MAJOR  1
 #define STM_SPEC_VERSION_MINOR  0
 
+// MU_CHANGE - [START]: Add new CPU information header for STM
+#if defined (_MSC_EXTENSIONS) || defined (__cplusplus)
+#define STATIC_ASSERT  static_assert
+#else
+#define STATIC_ASSERT  _Static_assert
```
makubacki: This is already defined in [Base.h](https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Base.h#L795-L809).

kenlautner: That's true.  The issue that arises is that when building the genstm tools in BaseTools it references this file but it never looks at Base.h.  This causes it to fail with STATIC_ASSERT.

kenlautner: It's the same reason we need to include redefinitions of SIZE_1KB and SIZE_2KB in GenStm.h.  I can move this redefinition to there if you prefer

kenlautner: I moved it because it just makes sense to have it in GenStm.h.

makubacki: I forgot this was being referenced for building the app in BaseTools. I think it'd make sense to move all of it to https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/Include/Common/BaseTypes.h and then update then remove the other one-off definition (e.g. https://github.com/tianocore/edk2/blob/master/BaseTools/Source/C/Common/PcdValueCommon.h#L21) so everything uses the single common copy.

makubacki: In a separate commit so this will become rebase and fast-forward with the original change to STM_HEADER.

kenlautner: The commits have been consolidated so a rebase and fast-forward will be clean.

MdePkg/Include/Register/Intel/StmApi.h
kenlautner: I added the additional checks because I thought that's a better solution.

BaseTools/Source/C/Common/PcdValueCommon.h
```diff
@@ -18,12 +18,6 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define __FLEXIBLE_SIZE(Size, TYPE, Field, MaxIndex)   if (__FIELD_SIZE(TYPE, Field) == 0) Size = MAX((__OFFSET_OF(TYPE, Field) + __ARRAY_ELEMENT_SIZE(TYPE, Field) * (MaxIndex)), Size)
 #define __ARRAY_SIZE(Array) (sizeof(Array)/sizeof(Array[0]))
 
-#if defined(_MSC_EXTENSIONS)
```
makubacki, line 21: The changes in this commit should be "MU_CHANGES" for now.

makubacki (dismissed):

BaseTools/Source/C/Common/PcdValueCommon.h
makubacki: To be clear here, comment out and tag as a MU_CHANGE.

kenlautner: Done.

makubacki (approved):

kuqin12 (approved):

PR #1337, opened 2025-04-21
[Cherry-pick] Fix a bug exposed in MM communicate v3

kuqin12:
## Description

With current implemenation, all 3 SmmCommunication* functions go through the same routine, which will dereference the incoming pointer to inspect whether this is a V3 buffer or not.

However, the caller always pass in the physical addresses, which could cause the system to page fault after OS take over the runtime control.

This change reverted the original common routine and added a specific v3 routine to support MM communicate v3.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested with OVMF package and booted to Windows.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
+++ b/MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
@@ -554,13 +554,11 @@ SmmCommunicationCommunicate (
   IN OUT UINTN                             *CommSize OPTIONAL
   )
 {
-  EFI_STATUS                    Status;
-  EFI_SMM_COMMUNICATE_HEADER    *CommunicateHeader;
-  EFI_MM_COMMUNICATE_HEADER_V3  *CommunicateHeaderV3 = NULL;
-  BOOLEAN                       OldInSmm;
-  UINTN                         TempCommSize;
-  UINTN                         CommHeaderSize;
-  UINT64                        LongCommSize; // MU_CHANGE: BZ3398
+  EFI_STATUS                  Status;
+  EFI_SMM_COMMUNICATE_HEADER  *CommunicateHeader;
+  BOOLEAN                     OldInSmm;
+  UINT64                      LongCommSize; // MU_CHANGE: BZ3398
+  UINTN                       TempCommSize;
 
   //
   // Check parameters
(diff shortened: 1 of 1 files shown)


apop5 (approved):

os-d (approved):

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -554,13 +554,11 @@ SmmCommunicationCommunicate (
   IN OUT UINTN                             *CommSize OPTIONAL
   )
 {
```
makubacki, line 556: I know we've discussed this offline, but the commit message and/or PR description should link to more details about the issue.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
```diff
@@ -752,11 +739,98 @@ MmCommunicationMmCommunicate3 (
...
+  UINT64                        TempCommSize;
+
+  //
+  // Check parameters
+  //
+  if ((CommBufferPhysical == NULL) || (CommBufferVirtual == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  CommunicateHeader = (EFI_MM_COMMUNICATE_HEADER_V3 *)CommBufferVirtual;
```
makubacki, line 755: is `CommunicateHeader->HeaderGuid` supposed to be `gEfiMmCommunicateHeaderV3Guid` in the input buffer? Should that be validated here before consuming data from the structure?

makubacki (approved):

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
kuqin12: Since the core component will evaluate this after getting into the MM anyway, checking it here is not security critical. But indeed it will be good to have.

I plan to add the v3 support for ARM communicate driver and switch the variable driver to use v3. Do you want to change it then? Otherwise I can add a check here under MU_CHANGE tag.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
kuqin12: PR description updated.

MdeModulePkg/Core/PiSmmCore/PiSmmIpl.c
makubacki: I agree it's not critical either so please feel free to wait. It's one of those inputs that just seem simple to validate before kicking off the full MM invocation process.

kuqin12: Okay, will add it in the next wave of the edk2 PR. Thanks for the suggestion.

PR #1342, opened 2025-04-24
back port additional changes to release/202502 [REBASE&FF]

apop5:
## Description

Backporting the changes for 
b7a26a3198ea5998bde35cf565ac3472d0153081
ed338da623774d34f10b48a47e8305f127afa4c9
e675273ce6a20e412b622036513ec7db66965192
ca48fd205ac7c3fb9f6fe033174d8b4a565c1467
e4325d5bcf8253a8a44fcea18330065e9cd542af
b7712d33c5d55e0f53fce58372c7567fbff08b6a
694ba3b79cee4c36e80fa3e1d27169c673a043b4

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?
- [ ] Backport to release branch?

## How This Was Tested
Local CI

## Integration Instructions
No integration necessary

PR #1343, opened 2025-04-24
[Squash on Rebase] Update release-basetools to use newer ubuntu.

apop5:
## Description

Ubuntu 20.04 on github have been deprecated and
have stopped working as of 2025.04.15.

Update to use ubuntu 22.04.

https://github.com/actions/runner-images/issues/11101

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
CI running of pipeline

## Integration Instructions
no integration necessary.

### DIFF
--- a/.github/workflows/release-basetools.yml
+++ b/.github/workflows/release-basetools.yml
@@ -34,12 +34,12 @@ jobs:
             output_dir: "BaseTools\\Bin\\Win64"
             target: AARCH64
 
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             tool_chain: GCC5
             output_dir: "BaseTools/Source/C/bin"
             target: X64
 
-          - os: ubuntu-20.04
+          - os: ubuntu-22.04
             tool_chain: GCC5
             output_dir: "BaseTools/Source/C/bin"
             target: AARCH64


os-d (approved):

makubacki (approved):

Javagedes (approved):

.github/workflows/release-basetools.yml
```diff
@@ -34,12 +34,12 @@ jobs:
             output_dir: "BaseTools\\Bin\\Win64"
             target: AARCH64
 
-          - os: ubuntu-20.04
+          - os: ubuntu-24.04
```
Javagedes: @apop5 I actually am concerned, now that I look at it more, that this is going to cause runtime issues (as in running these tools) for people running on any ubuntu version lower than 24.04. Can you test that by using a 22.04 WSL image with these basetools?

Javagedes: One unfortunate thing about us squashing commits and cleaning up history when rebasing off of edk2, is we lose the git history - But see this PR for us going down to 20.04 for the reason I described above. https://github.com/microsoft/mu_basecore/pull/467. It is unfortunately not very descriptive, but @kuqin12 am I correct in remembering that we had to do this because using an older version of ubuntu then what it is compiled with results in dll loading errors or something like that?

I think if we have to bump it, it should only be to 22.04.

kuqin12: Yes, updating to 24.04 will pull in a later version of libc, which will not work for folks that are using older version of ubuntu.

apop5: @kuqin12 @Javagedes 

the ubuntu 20.04 runner has been depreciated on github

https://github.com/actions/runner-images/issues/11101


Attempting to use the runner will result in a failed/canceled job, as experienced on: https://github.com/microsoft/mu_basecore/pull/1342

If libc is a problem with 24.04, would limiting the update to Ubuntu 22.04 help this at all? It will give us a reprieve until the next major release of ubuntu occurs and we run into the same depreciation there.

makubacki: Can't we just build and publish both? They're both valid and we're the builder and publisher. I didn't really care too much in my review because we control most of the environments that would use this and since 24.04 is available and LTS, I chose to go with not dealing with this again as soon in the future if we choose one. But we don't have to.

Javagedes: There is no need to compile both because lower versions work with greater versions. But because GitHub is dropping support for 20.04 we are basically being forced to also drop support for 20.04 pre-compiled base tools. Unless we manually compile and publish or have a separate self-maintained 20.04 container.

I think incrementing to 22.04 is the right thing to do, with some kind of acknowledgment that if someone wants to compile on 20.04, they must compile their own base tools locally instead of using the extdep. This is already what happens if someone wants to compile with 18.04 for some reason

apop5: I changed it to 22.04.

makubacki: My statement about compiling for both is not about technical compatibility. It is about making it simple to track and deprecate individual builds in the future. If there was a 20.04 build, it would simply be removed from the workflow with the VM image change and anyone using the latest 20.04 binaries would keep working because their "latest" stays at that 20.04 based release. Anyone that wanted to move to a newer OS image would have an easy and obvious choice as to which to use because they would simply match their Ubuntu image 1:1 to the version in published feed.

makubacki: I'm obviously fine with 22.04 too to solve the immediate problem. But I don't think that's as sustainable as building and tracking releases per Ubuntu version which also allows future hot patches to easily be picked up for the given Ubuntu version, etc.

apop5: I'll open a GitHub issue to track  having specific versions based on Ubuntu releases, and switching to this for immediate resolution.

apop5: https://github.com/microsoft/mu_basecore/issues/1346

kuqin12: > I'm obviously fine with 22.04 too to solve the immediate problem. But I don't think that's as sustainable as building and tracking releases per Ubuntu version which also allows future hot patches to easily be picked up for the given Ubuntu version, etc.

Changing this to 22.04 is fine. But I wonder if we can get rid of the glibc reliance provided by the system. Otherwise we are bound to _some_ version regardless. I used to think this will get resolved by move base tools to python. But no one seems to like it. Should we resurrect the effort..?

makubacki: The proposal to associate BaseTools releases with Ubuntu versions is meant to be a middle ground for the way we actually consume the tools.

- Current: Keep running into the same VM image is deprecated can't build any more problem. Future versions published are no longer compatible with previous consumers.
- Publish Images Per Ubuntu Version: Still have to update the VM image for build. But consumption becomes cleaner by making compatibility obvious and consumers having stable packages to depend on. Hot fixes and future updates are possible to older packages for that release.
- Python: Don't worry about things like glibc and improve overall portability.

My understanding is that we concluded the Python path wasn't worth the effort recently. If someone is sufficiently motivated to do that, it would overall be better. In the bigger picture, I don't know if I'd prioritize it at the top of outstanding problems.

PR #1355, opened 2025-05-08
Add X509CompareDateTime Function

liqiqiii:
## Description

Add X509CompareDateTime function.

This function was missed before because it has a return type of INT32 which is not included in current generate_cryptodriver.py.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with X64 STANDARD/TINY_SHA, build successfully.

## Integration Instructions

When integrating this, consume the latest MU_CRYPTO_RELEASE [PR](https://github.com/microsoft/mu_crypto_release/pull/132) too.


CryptoPkg/Include/Protocol/Crypto.h
```diff
@@ -4649,6 +4670,7 @@ struct _EDKII_CRYPTO_PROTOCOL {
   EDKII_CRYPTO_X509_GET_EXTENSION_DATA                X509GetExtensionData;
   EDKII_CRYPTO_X509_GET_VALIDITY                      X509GetValidity;
   EDKII_CRYPTO_X509_FORMAT_DATE_TIME                  X509FormatDateTime;
+  EDKII_CRYPTO_X509_COMPARE_DATE_TIME                 X509CompareDateTime;
```
makubacki: I know this is roughly where the function sits in the protocol interface in [edk2](https://github.com/tianocore/edk2/blob/master/CryptoPkg/Private/Protocol/Crypto.h), but placing it here makes a breaking change for the crypto protocol as it is defined in Project Mu (which is already a different binary interface that that in edk2).

So, I'd propose this be placed at the end of the structure to at least maintain backward compatibility within the Project Mu protocol interface definition.

The protocol version should also be updated to reflect that this new function is available.

liqiqiii: I have made changes based on your comments and updated the protocol version. Thanks Michael.

makubacki (approved):

apop5 (approved):

liqiqiii:
Wait for some final validation of the new function.

liqiqiii:
> Wait for some final validation of the new function.

all set, test passed. Ready for checkin

kenlautner (approved):

PR #1357, opened 2025-05-09
Fix type annotations

gameroman:


apop5:
@GameRoMan 
Thank you for this submission.

The code that you are modifying comes from the upstream EDK2 Project

https://github.com/tianocore/edk2

These modifications should be submitted into that repo, and they will flow downstream to us here.

apop5:
Closing in basecore and will use upstream edk2 changes after it merges

https://github.com/tianocore/edk2/pull/11061

PR #1360, opened 2025-05-13
Add PiSmmCoreMmServicesTableLib

kenlautner:
## Description

Add PiSmmCoreMmServicesTableLib as a new library that the SMM_CORE can use.  With the rework of the MM_CORE performanceLib  a MmServiceTableLib is necessary which led to the creation of this library.  In the future if the SMM_CORE needs to link a MmServicesTableLib this library can be used.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on performance builds for the PiSmmCore.  The core is now able to be built and there are no issues during boot.

## Integration Instructions

N/A


makubacki:
@kenlautner, has this change been boot tested as well (PR description only mentions build)? I'm not sure without looking in detail but I thought there might be a race condition against `InSmm` with this being linked as a constructor against the MM core. As `InSmm` is sometimes checked in MM Base protocol implementations when getting the SMST/MMST table location.

kenlautner:
> @kenlautner, has this change been boot tested as well (PR description only mentions build)? I'm not sure without looking in detail but I thought there might be a race condition against `InSmm` with this being linked as a constructor against the MM core. As `InSmm` is sometimes checked in MM Base protocol implementations when getting the SMST/MMST table location.

I tested it and it is an issue.  I changed the PR to add a NULL MmServicesTableLib but if you think reworking the perf lib makes more sense I could do that.

makubacki:
> > @kenlautner, has this change been boot tested as well (PR description only mentions build)? I'm not sure without looking in detail but I thought there might be a race condition against `InSmm` with this being linked as a constructor against the MM core. As `InSmm` is sometimes checked in MM Base protocol implementations when getting the SMST/MMST table location.
> 
> I tested it and it is an issue. I changed the PR to add a NULL MmServicesTableLib but if you think reworking the perf lib makes more sense I could do that.

I think having that dependency in the perf lib is fine in itself. Have you looked at making a `MmServicesTableLib` instance of [PiSmmCoreSmmServicesTableLib](https://github.com/microsoft/mu_basecore/blob/dev/202502/MdeModulePkg/Library/PiSmmCoreSmmServicesTableLib/PiSmmCoreSmmServicesTableLib.c)? Something like `PiSmmCoreMmServicesTableLib` (`MmServicesTableLib` library class) that supports `SMM_CORE`.

As long as you're not reaching out of the core into the protocol I think that should be a reasonable `MmServicesTableLib` equivalent to `PiSmmCoreSmmServicesTableLib` for use with `PiSmmCore`.

kenlautner:
> > > @kenlautner, has this change been boot tested as well (PR description only mentions build)? I'm not sure without looking in detail but I thought there might be a race condition against `InSmm` with this being linked as a constructor against the MM core. As `InSmm` is sometimes checked in MM Base protocol implementations when getting the SMST/MMST table location.
> > 
> > 
> > I tested it and it is an issue. I changed the PR to add a NULL MmServicesTableLib but if you think reworking the perf lib makes more sense I could do that.
> 
> I think having that dependency in the perf lib is fine in itself. Have you looked at making a `MmServicesTableLib` instance of [PiSmmCoreSmmServicesTableLib](https://github.com/microsoft/mu_basecore/blob/dev/202502/MdeModulePkg/Library/PiSmmCoreSmmServicesTableLib/PiSmmCoreSmmServicesTableLib.c)? Something like `PiSmmCoreMmServicesTableLib` (`MmServicesTableLib` library class) that supports `SMM_CORE`.
> 
> As long as you're not reaching out of the core into the protocol I think that should be a reasonable `MmServicesTableLib` equivalent to `PiSmmCoreSmmServicesTableLib` for use with `PiSmmCore`.
I haven't considered that but it makes sense to have a library for that.  I'll make the change for it.

kuqin12:
This change looks odd. Why would there be a MM library linked to SMM_CORE? Do we use MM table directly from the SMM core?  Would it make better sense to have the SMM performance lib be a real SMM instance?

makubacki:
I'm going to close this PR for now while I work on an alternative approach.

PR #1362, opened 2025-05-15
[REBASE & FF] Integrate edk2 Memory Protection Fixes and Fix the Ensuing Chaos

os-d:
## Description

This PR reverts a Mu commit that was upstreamed to edk2 as well as cherry-picking some new edk2 functionality. Changing the behavior in edk2 exposed a number of Mu bugs in memory protections that are now fixed.

Revert "MdeModulePkg: Update Memory Logging"
--
This reverts commit 7006e33619f04ba7d1b8da97ccbb933f47c03504 as it has been upstreamed to edk2 in commits d6101acb083eb0b79e3ca35c18ea87003850e3c4, 5ccb5fff02a66b21898bd57f48bbd7c3cd6f4e8d, and 6c6d6f42db3f5c3dc2b8d86ce32e1ce2c7fd85d1.

Four edk2 Cherry-Picks
--
These cherry-picks cover the commit that was reverted and better align image memory protections with the GCD, to keep it in sync, enabling a use case to apply CPU_CRYPTO to all memory near the end of boot.

[SQUASH ON REBASE]MdeModulePkg: Don't Set EFI_MEMORY_RP in the GCD
--
EFI_MEMORY_RP is set on free memory when the corresponding bit in the memory protection HOB is enabled. It is directly set in the page table using the CPU arch protocol because it cannot go through the GCD; it is not feasible with the current locking structure for the paging code to unset it in the GCD, as that may require new pages to be allocated, etc.

This commit fixes the memory protection code to not apply RP in the GCD on accident when trying to apply NX protections. This
causes issues as RP is never cleared and if a driver tries to get attributes from the GCD and update a caching attribute, say, it can end up marking the page RP and then fault trying to acces it.

This also prevents UefiCpuPkg's CpuDxe from applying RP to the GCD when refreshing the GCD from paging, for the same reason.

This should be squashed with ea5dd587ba7713eb267e4b26eb304bd9bf81f19f on rebase.

[SQUASH ON REBASE] MdeModulePkg: Page Align Access Attr Removal
--
Currently, when the image loading code decides not to protect an image, it will attempt to remove any protection attributes from the image region to remove NX. However, it does not take into account that the primary reason we do not protect images is if they are not page aligned. This causes asserts when trying to unset these attributes through the GCD.

This patch aligns the access attribute removal to page boundaries.

It should be squashed on rebase with b95a42dd1ac0cdce3a94db96a4a37856b84c5e16.

[SQUASH ON REBASE] MdeModulePkg: Remove Image Access Attributes Correctly
--
We support loading non-page aligned images, however, in order to do so, we must remove any protection attributes set on them, for example NX protections applied early in DXE Core.

However, when we encounter non-page aligned images, we attempt to take their image base (which is not page aligned) and set attributes on it, which causes an assert to be hit that page attributes are only set on page aligned memory.

This commit removes the protection attributes for the entire page(s) containing the non-aligned image.

It should be squashed with e7e74b9cfc61a126bfd3519b6cb3673492147e74 on rebase.

[SQUASH ON REBASE] MdeModulePkg: Don't Unset Memory Attrs to Unprotected Images
--
Currently, UnprotectUefiImage will attempt to unset memory attributes on any image record it finds for the given image, protected or unprotected. However, this is both a waste of time for unprotected images and can cause asserts when the image is unprotected because it does not have page aligned sections.

This patch fixes that behavior and does not attempt to unset the memory protections that were never applied.

It should be squashed with b95a42dd1ac0cdce3a94db96a4a37856b84c5e16 on rebase.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an Intel server with aligned and unaligned images booting to shell.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -265,7 +265,7 @@ SetUefiImageMemoryAttributes (
 
   FinalAttributes = (Descriptor.Attributes & EFI_CACHE_ATTRIBUTE_MASK) | (Attributes & EFI_MEMORY_ATTRIBUTE_MASK);
 
-  DEBUG ((DEBUG_INFO, "SetUefiImageMemoryAttributes - 0x%016lx - 0x%016lx (0x%016lx)\n", BaseAddress, Length, FinalAttributes));
+  DEBUG ((DEBUG_VERBOSE, "SetUefiImageMemoryAttributes - 0x%016lx - 0x%016lx (0x%016lx)\n", BaseAddress, Length, FinalAttributes));
 
   ASSERT (gCpu != NULL);
   gCpu->SetMemoryAttributes (gCpu, BaseAddress, Length, FinalAttributes);
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -696,6 +696,15 @@ UnprotectUefiImage (
 
     if (ImageRecord->ImageBase == (EFI_PHYSICAL_ADDRESS)(UINTN)LoadedImage->ImageBase) {
       mImagePropertiesPrivate.ImageRecordCount--;
+      // this image was protected, so we need to unprotect it
+      if (gCpu != NULL) {
+        SetUefiImageMemoryAttributes (
+          ImageRecord->ImageBase,
+          ImageRecord->ImageSize,
+          0
+          );
+      }
+
       goto Free;
     }
   }
@@ -713,6 +722,8 @@ UnprotectUefiImage (
 
     if (ImageRecord->ImageBase == (EFI_PHYSICAL_ADDRESS)(UINTN)LoadedImage->ImageBase) {
       mNonProtectedImageRangesPrivate.NonProtectedImageCount--;
+      // this image was unprotected, so we don't unprotect it. Trying to do so can cause an assert because the
+      // image may not be page aligned
       goto Free;
     }
   }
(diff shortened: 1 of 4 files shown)


kuqin12 (approved):

apop5 (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -529,7 +632,7 @@ ProtectUefiImage (
 
   if ((ProtectionPolicy == PROTECT_IF_ALIGNED_ELSE_ALLOW)) {
     ClearAccessAttributesFromMemoryRange (
-      (EFI_PHYSICAL_ADDRESS)(UINTN)LoadedImage->ImageBase,
+      (EFI_PHYSICAL_ADDRESS)(UINTN)LoadedImage->ImageBase & ~EFI_PAGE_MASK,
```
apop5, line 635: If the ImageBase is unaligned, this might end up clearing something else that was also included in the same page.

i..e ImageBase = 0xBBB953C0 it would clear the attributes starting from BB95000 - BB95FFFF. But If something was in the memory before that, chances are it was unprotected as well.

os-d: Correct, this is intentional. If unaligned images are used, we have to clear XP from the entire page(s). This will affect whatever else is there, but only in a freeing sense, not restrictive. So either another unaligned image, in which case it also needs it or random memory, in which case marking it RWX won’t cause it to crash, just degrade the security state. But, the system already has a degraded security state because of this unaligned image and it will crash if NX is not removed.

The general advice would be to align all images to either default or runtime page granularity, depending on driver type.

PR #1364, opened 2025-05-19
[REBASE & FF] MdeModulePkg: Create IoMmuLib and add support to - PciBusDxe, PciHostBridgeDxe, NonDiscoverablePciDeviceDxe

eeshanl:
## Description

Reverts PcdRequireIommu change, removes PcdRequireIommu.
Create IoMmuLib
Create IoMmuLibNull
Wrapper library for IoMmu Protocol functions.

Add IoMmuLib support to NonDiscoverablePciDeviceDxe
Modify IoMmu protocol usage and add IoMmuLib support to PciBusDxe and PciHostBridgeDxe

- [x] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU SBSA, physical arm device, physical x64 device.

### Platform Integration

The library automatically adapts to platform capabilities. On platforms without IOMMU support, use IoMmuLibNull.
On platforms with IOMMU support, use IoMmuLib.

For integrating with a platform, in the top level DSC, you can do the following for example:

```text
  # Enable IoMmu/Smmu support
  DEFINE REQUIRE_IOMMU             = TRUE

!if $(REQUIRE_IOMMU) == TRUE
  IoMmuLib|MdeModulePkg/Library/IoMmuLib/IoMmuLib.inf
!else
  IoMmuLib|MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.inf
!endif
```

### DIFF
--- a/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
+++ b/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
@@ -15,11 +15,14 @@
 
 #include <Protocol/PciRootBridgeIo.h>
 
+#include <Library/IoMmuLib.h> // MU_CHANGE
+
 typedef struct {
   EFI_PHYSICAL_ADDRESS             AllocAddress;
   VOID                             *HostAddress;
   EFI_PCI_IO_PROTOCOL_OPERATION    Operation;
   UINTN                            NumberOfBytes;
+  VOID                             *IoMmuContext; // MU_CHANGE
 } NON_DISCOVERABLE_PCI_DEVICE_MAP_INFO;
 
 /**
@@ -731,6 +734,13 @@ CoherentPciIoMap (
   EFI_STATUS                            Status;
   NON_DISCOVERABLE_PCI_DEVICE_MAP_INFO  *MapInfo;
 
+  // MU_CHANGE [BEGIN]
+  VOID                   *IoMmuHostAddress;
+  EDKII_IOMMU_OPERATION  IoMmuOperation;
+  UINT64                 IoMmuAttribute;
+
+  // MU_CHANGE [END]
+
   if ((Operation != EfiPciIoOperationBusMasterRead) &&
       (Operation != EfiPciIoOperationBusMasterWrite) &&
       (Operation != EfiPciIoOperationBusMasterCommonBuffer))
@@ -745,6 +755,19 @@ CoherentPciIoMap (
   {
     return EFI_INVALID_PARAMETER;
   }
+
+  // MU_CHANGE [BEGIN]
+  MapInfo = AllocatePool (sizeof *MapInfo);
+  if (MapInfo == NULL) {
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  MapInfo->AllocAddress  = MAX_UINT32;
+  MapInfo->HostAddress   = HostAddress;
+  MapInfo->Operation     = Operation;
+  MapInfo->NumberOfBytes = *NumberOfBytes;
+  MapInfo->IoMmuContext  = NULL;
+  // MU_CHANGE [END]
 
   //
   // If HostAddress exceeds 4 GB, and this device does not support 64-bit DMA
@@ -761,16 +784,6 @@ CoherentPciIoMap (
       return EFI_UNSUPPORTED;
     }
 
-    MapInfo = AllocatePool (sizeof *MapInfo);
-    if (MapInfo == NULL) {
-      return EFI_OUT_OF_RESOURCES;
-    }
-
-    MapInfo->AllocAddress  = MAX_UINT32;
-    MapInfo->HostAddress   = HostAddress;
-    MapInfo->Operation     = Operation;
-    MapInfo->NumberOfBytes = *NumberOfBytes;
-
     Status = gBS->AllocatePages (
                     AllocateMaxAddress,
                     EfiBootServicesData,
@@ -783,8 +796,8 @@ CoherentPciIoMap (
       // 4 GB to begin with. There is not much we can do about that other than
       // fail the map request.
       //
-      FreePool (MapInfo);
-      return EFI_DEVICE_ERROR;
+      Status = EFI_DEVICE_ERROR; // MU_CHANGE
+      goto FreeMapInfo;          // MU_CHANGE
     }
 
     if (Operation == EfiPciIoOperationBusMasterRead) {
@@ -795,14 +808,76 @@ CoherentPciIoMap (
              );
     }
 
-    *DeviceAddress = MapInfo->AllocAddress;
-    *Mapping       = MapInfo;
+    *DeviceAddress   = MapInfo->AllocAddress;
+    IoMmuHostAddress = (VOID *)(UINTN)MapInfo->AllocAddress; // MU_CHANGE
   } else {
-    *DeviceAddress = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress;
-    *Mapping       = NULL;
+    *DeviceAddress   = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress;
+    IoMmuHostAddress = HostAddress;      // MU_CHANGE
+  }
+
+  // MU_CHANGE [BEGIN]
+  *Mapping = MapInfo;
+
+  switch (Operation) {
+    case EfiPciIoOperationBusMasterRead:
+      IoMmuOperation = EdkiiIoMmuOperationBusMasterRead;
+      IoMmuAttribute = EDKII_IOMMU_ACCESS_READ;
+      break;
+
+    case EfiPciIoOperationBusMasterWrite:
+      IoMmuOperation = EdkiiIoMmuOperationBusMasterWrite;
+      IoMmuAttribute = EDKII_IOMMU_ACCESS_WRITE;
+      break;
+
+    case EfiPciIoOperationBusMasterCommonBuffer:
+      IoMmuOperation = EdkiiIoMmuOperationBusMasterCommonBuffer;
+      IoMmuAttribute = EDKII_IOMMU_ACCESS_READ | EDKII_IOMMU_ACCESS_WRITE;
+      break;
+
+    default:
+      DEBUG ((DEBUG_ERROR, "%a - Invalid operation %d\n", __func__, Operation));
+      ASSERT (FALSE);
+      return EFI_INVALID_PARAMETER;
+  }
+
+  Status = IoMmuMap (
+             IoMmuOperation,
+             IoMmuHostAddress,
+             NumberOfBytes,
+             DeviceAddress,
+             &MapInfo->IoMmuContext
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a - IoMmuMap failed.\n", __func__));
+    ASSERT (FALSE);
+    goto FreeMapInfo;
+  }
+
+  Status = IoMmuSetAttribute (
+             NULL,
+             MapInfo->IoMmuContext,
+             IoMmuAttribute
+             );
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a - IoMmuSetAttribute failed.\n", __func__));
+    ASSERT (FALSE);
+    goto Unmap;
   }
 
   return EFI_SUCCESS;
+
+Unmap:
+  Status = IoMmuUnmap (MapInfo->IoMmuContext);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a - IoMmuUnmap failed.\n", __func__));
+    ASSERT (FALSE);
+  }
+
+FreeMapInfo:
+  FreePool (MapInfo);
+  return Status;
+
+  // MU_CHANGE [END]
 }
 
 /**
@@ -823,23 +898,43 @@ CoherentPciIoUnmap (
   )
 {
   NON_DISCOVERABLE_PCI_DEVICE_MAP_INFO  *MapInfo;
+  // MU_CHANGE [BEGIN]
+  EFI_STATUS  Status;
+
+  if (Mapping == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
 
   MapInfo = Mapping;
-  if (MapInfo != NULL) {
-    if (MapInfo->Operation == EfiPciIoOperationBusMasterWrite) {
-      gBS->CopyMem (
-             MapInfo->HostAddress,
-             (VOID *)(UINTN)MapInfo->AllocAddress,
-             MapInfo->NumberOfBytes
-             );
-    }
 
-    gBS->FreePages (
-           MapInfo->AllocAddress,
-           EFI_SIZE_TO_PAGES (MapInfo->NumberOfBytes)
-           );
-    FreePool (MapInfo);
+  Status = IoMmuSetAttribute (NULL, MapInfo->IoMmuContext, 0);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a - IoMmuSetAttribute failed.\n", __func__));
+    ASSERT (FALSE);
+    return Status;
   }
+
+  Status = IoMmuUnmap (MapInfo->IoMmuContext);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "%a - IoMmuUnmap failed.\n", __func__));
+    ASSERT (FALSE);
+    return Status;
+  }
+
+  if (MapInfo->Operation == EfiPciIoOperationBusMasterWrite) {
+    gBS->CopyMem (
+           MapInfo->HostAddress,
+           (VOID *)(UINTN)MapInfo->AllocAddress,
+           MapInfo->NumberOfBytes
+           );
+  }
+
+  gBS->FreePages (
+         MapInfo->AllocAddress,
+         EFI_SIZE_TO_PAGES (MapInfo->NumberOfBytes)
+         );
+  FreePool (MapInfo);
+  // MU_CHANGE [END]
 
   return EFI_SUCCESS;
 }
@@ -1278,6 +1373,13 @@ NonCoherentPciIoMap (
   EFI_GCD_MEMORY_SPACE_DESCRIPTOR       GcdDescriptor;
   BOOLEAN                               Bounce;
 
+  // MU_CHANGE [BEGIN]
+  VOID                   *IoMmuHostAddress;
+  EDKII_IOMMU_OPERATION  IoMmuOperation;
+  UINT64                 IoMmuAttribute;
+
+  // MU_CHANGE [END]
+
   AllocAddress = NULL; // MS_CHANGE for vs2017
 
   if ((HostAddress   == NULL) ||
@@ -1303,6 +1405,7 @@ NonCoherentPciIoMap (
   MapInfo->HostAddress   = HostAddress;
   MapInfo->Operation     = Operation;
   MapInfo->NumberOfBytes = *NumberOfBytes;
+  MapInfo->IoMmuContext  = NULL; // MU_CHANGE
 
   Dev = NON_DISCOVERABLE_PCI_DEVICE_FROM_PCI_IO (This);
 
@@ -1372,10 +1475,12 @@ NonCoherentPciIoMap (
       gBS->CopyMem (AllocAddress, HostAddress, *NumberOfBytes);
     }
 
-    *DeviceAddress = MapInfo->AllocAddress;
+    *DeviceAddress   = MapInfo->AllocAddress;
+    IoMmuHostAddress = (VOID *)(UINTN)MapInfo->AllocAddress; // MU_CHANGE
   } else {
     MapInfo->AllocAddress = 0;
     *DeviceAddress        = (EFI_PHYSICAL_ADDRESS)(UINTN)HostAddress;
+    IoMmuHostAddress      = HostAddress; // MU_CHANGE
 
     //
     // We are not using a bounce buffer: the mapping is sufficiently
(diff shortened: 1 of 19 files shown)


eeshanl:
Will add IoMmuLib support for PciIo and PciRootBridgeIo in a separate PR.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+IoMmuMap (
+  IN     EDKII_IOMMU_OPERATION  Operation,
+  IN     VOID                   *HostAddress,
+  IN OUT UINTN                  *NumberOfBytes,
+  OUT    EFI_PHYSICAL_ADDRESS   *DeviceAddress,
+  OUT    VOID                   **Mapping
+  );
+
+/**
+  Unmap a device address in the Page Table, invalidate the TLB with TLBI operation.
```
os-d: TLBI is AARCH64 specific, we should just say invalidate TLB by VA (as I'm assuming it is doing) or don't mention the TLB flush as that is library implementation detail, not relevant to the consumer.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+  @param [out]     DeviceAddress   The resulting device address.
+  @param [out]     Mapping         A handle to the mapping.
+
+  @retval EFI_SUCCESS              Success.
+  @retval EFI_INVALID_PARAMETER    Invalid parameter.
+  @retval EFI_OUT_OF_RESOURCES     Out of resources.
+**/
+EFI_STATUS
+EFIAPI
+IoMmuMap (
```
os-d, line 45: Do these APIs contain any changes you want to make to IOMMU APIs?

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+  Free a buffer allocated by IoMmuAllocateBuffer.
+
+  @param [in]  Pages         The number of pages to free.
+  @param [in]  HostAddress   The host address to free.
+
+  @retval EFI_SUCCESS            The requested pages were freed.
+  @retval EFI_INVALID_PARAMETER  Memory is not a page-aligned address or Pages is invalid.
+  @retval EFI_NOT_FOUND          The requested memory pages were not allocated with AllocatePages().
```
os-d: return statuses need updating for this function (looks like copy/paste)

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+  @param [in]      Pages         The number of pages to allocate.
+  @param [in, out] HostAddress   On input, the desired host address. On output, the allocated host address.
+  @param [in]      Attributes    The memory attributes to use for the allocation.
+
+  @retval EFI_SUCCESS           The requested pages were allocated.
+  @retval EFI_INVALID_PARAMETER 1) Type is not AllocateAnyPages or
+                                AllocateMaxAddress or AllocateAddress.
+                                2) MemoryType is in the range
+                                EfiMaxMemoryType..0x6FFFFFFF.
+                                3) Memory is NULL.
```
os-d: HostAddress

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+  IN OUT VOID               **HostAddress,
+  IN     UINT64             Attributes
+  );
+
+/**
+  Set the R/W access attributes for Mapping in the Page Table.
+
+  @param [in]  DeviceHandle  The device handle to set attributes for.
+  @param [in]  Mapping       The mapping to set attributes for.
+  @param [in]  IoMmuAccess   The IOMMU access attributes for R/W.
```
os-d: Are the only possible attributes (now and forever) RW?

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+  IN     UINTN              Pages,
+  IN OUT VOID               **HostAddress,
+  IN     UINT64             Attributes
+  );
+
+/**
+  Set the R/W access attributes for Mapping in the Page Table.
+
+  @param [in]  DeviceHandle  The device handle to set attributes for.
+  @param [in]  Mapping       The mapping to set attributes for.
```
os-d: As this is a VOID *, worth mentioning this is the mapping returned by IoMmuMap

MdeModulePkg/Library/IoMmuLib/IoMmuLib.inf
```diff
@@ -0,0 +1,21 @@
+[Defines]
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = IoMmuLib
+  FILE_GUID                      = 55af180d-a4ee-47fa-84d5-01071643ad35
+  MODULE_TYPE                    = BASE
```
os-d: Can't be BASE, uses boot services, which it should also have a LibraryClass dependency on UefiBootServicesTableLib, BaseLib, and DebugLib

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
```diff
@@ -0,0 +1,115 @@
...
+EFIAPI
+IoMmuMap (
+  IN     EDKII_IOMMU_OPERATION  Operation,
+  IN     VOID                   *HostAddress,
+  IN OUT UINTN                  *NumberOfBytes,
+  OUT    EFI_PHYSICAL_ADDRESS   *DeviceAddress,
+  OUT    VOID                   **Mapping
+  )
+{
+  return EFI_UNSUPPORTED;
```
os-d: Should these return EFI_SUCCESS so that a consumer doesn't have to know whether the null instance is being used or not?

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
```diff
@@ -0,0 +1,115 @@
...
+    This file contains all the NULL implementation of the IoMmu protocol library functions.
+
+    Copyright (c) Microsoft Corporation.
+    SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+#include <Library/BaseLib.h>
+#include <Library/DebugLib.h>
+#include <Library/IoMmuLib.h>
+#include <Library/UefiBootServicesTableLib.h>
```
os-d: Don't need DebugLib.h or UefiBootServicessTableLib.h

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.inf
```diff
@@ -0,0 +1,14 @@
...
+  INF_VERSION                    = 0x00010005
+  BASE_NAME                      = IoMmuLibNull
+  FILE_GUID                      = 743616a1-4fe5-41c2-9534-e222f6f61081
+  MODULE_TYPE                    = BASE
+  VERSION_STRING                 = 1.0
+  LIBRARY_CLASS                  = IoMmuLib
+
+[Sources]
+  IoMmuLibNull.c
+
```
os-d, line 11: LibraryClasses dependency on BaseLib

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -35,6 +35,9 @@
   #                   to allow the platform to hook pre- and post- VariablePolicy locking.
   #                   NOTE: Must be implemented as a DXE_DRIVER lib.
   NULL|Include/Library/VariablePolicyLockingExLib.h
+
+  ## @libraryclass IoMmu library that can be used for IoMmu Protocol functions.
+  IoMmuLib|Include/Library/IoMmuLib.h
```
os-d, line 42: Maybe put in own MU_CHANGE section?

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
+/** @file IoMmuLib.h
```
os-d, line 1: Maybe a small README with the library to indicate why one is chosen over the other, relationship to PCD (or if we are getting rid of PCD), why this is used over protocol directly, etc.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
```diff
@@ -1396,6 +1409,60 @@ NonCoherentPciIoMap (
...
+  }
+
+  Status = IoMmuMap (
+             IoMmuOperation,
+             IoMmuHostAddress,
+             NumberOfBytes,
+             DeviceAddress,
+             &MapInfo->IoMmuContext
+             );
+  if (Status != EFI_UNSUPPORTED) {
```
os-d: If null lib is changed, then changing this.

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: @os-d updated pr: https://github.com/microsoft/mu_silicon_arm_tiano/pull/374 to switch from TLBI ALL to TLBI_S2_IPA InputAddress of Stage 2 SMMU.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
```diff
@@ -1396,6 +1409,60 @@ NonCoherentPciIoMap (
   }
 
   *Mapping = MapInfo;
```
cfernald, line 1503: Will we leak this in even of a lower failure? Because we will set the mapping and not do the free below this point, however we are returning an error so the caller would shouldn't assume its valid.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+  IN     VOID                   *HostAddress,
+  IN OUT UINTN                  *NumberOfBytes,
+  OUT    EFI_PHYSICAL_ADDRESS   *DeviceAddress,
+  OUT    VOID                   **Mapping
+  );
+
+/**
+  Unmap a device address in the Page Table, invalidate the TLB with TLBI operation.
+
+  @param [in]  This      Pointer to the IOMMU protocol instance.
```
makubacki: Can drop this parameter.

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: done, added.

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: yes, added StreamId and SmmuBase

MdeModulePkg/Library/IoMmuLib/IoMmuLib.inf
eeshanl: what should the type be instead?

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
eeshanl: In PciIo, and PciRootBridgeIo, the IoMmu map, and unmap call's statuses are imediatly returned, whereas in NonDiscoverable, we continue on with the rest of the function after the IoMmu map or unmap. Due to this difference, we need to be able to distinguish between a NULL implementation to skip the logic essentially, or a real one.

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.inf
eeshanl: done

MdeModulePkg/MdeModulePkg.dec
eeshanl: done

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
eeshanl: Fixed to free/unmap on error.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
eeshanl: per the other comment on unsupported vs success, we would need to keep it like this because of how the iommu functions are called in PciRootBridgeIo vs NonDiscoverable.

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: fixed comments throughout

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: fixed comments throughout

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: dropped R/W from comment.

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: yes, done

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
eeshanl: done

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: done

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
os-d: Are you saying that in the nondiscoverable case we have some logic that we only do if an IoMmu is attached? Or just that there is additional logic that occurs in the driver after?

If it is the latter, it doesn't need to distinguish, right, it can just check if EFI_ERROR, return (or handle appropriately), otherwise continue.

MdeModulePkg/Library/IoMmuLib/IoMmuLib.inf
os-d: It should be `DXE_DRIVER` because it uses services only available in DXE

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
os-d: Do we though? Why can't we have the null return Success and just remove the if Status != EFI_UNSUPPORTED? We can just check for an error and fail out, right?

MdeModulePkg/Library/IoMmuLib/IoMmuLib.inf
eeshanl: done, fixed. thank you.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+/**
+  Map a host address to a device address using the Page Table.
+  Currently, this function only supports identity mapping.
+
+  @param [in]      Operation       The type of IOMMU operation.
+  @param [in]      HostAddress     The host address to map.
+  @param [in, out] NumberOfBytes   On input, the number of bytes to map. On output, the number of bytes mapped.
+  @param [out]     DeviceAddress   The resulting device address.
+  @param [out]     Mapping         A handle to the mapping. Used by IoMmuUnmap to unmap the address and IoMmuSetAttribute to set attributes.
+  @param [in]      StreamId        Optional SMMU stream ID for the mapping.
```
cfernald: Change SMMU to IOMMU here.

cfernald: Should detail the expected behavior if the StreamId and/or the SmmuBase is NULL.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+**/
+EFI_STATUS
+EFIAPI
+IoMmuFreeBuffer (
+  IN  UINTN  Pages,
+  IN  VOID   *HostAddress
+  );
+
+/**
+  Allocate a buffer for use with the IOMMU.
```
cfernald: Could you clarify what "use with the IOMMU" means? i.e. is DMA-able? is <4g? etc..

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
cfernald: I tend to agree with @os-d, If the platform is deciding to use the NULL instance, then it is effectively stubbing out the functionality under the assumption it is not needed and so returning EFI_UNSUPPORTED doesn't really add anything. Especially since this might be a real returned error from other instances.

cfernald (approved):

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
eeshanl: If you look at the IoMmuMap calls inside RootBridgeIo, it is wrapped around a if (mIommu != NULL) . As a result, when no iommu is present that whole block of logic is skipped. With the NULL IoMmuLibNull, that outter if statement is gone and the function will execute no matter what. We want it to skip the logic in this case and not return the status. So if i have both Libs return SUCCESS, i won't be able to tell when to skip or return the Status.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -0,0 +1,122 @@
...
+**/
+EFI_STATUS
+EFIAPI
+IoMmuMap (
+  IN     EDKII_IOMMU_OPERATION  Operation,
+  IN     VOID                   *HostAddress,
+  IN OUT UINTN                  *NumberOfBytes,
+  OUT    EFI_PHYSICAL_ADDRESS   *DeviceAddress,
+  OUT    VOID                   **Mapping,
+  IN     UINTN                  StreamId         OPTIONAL,
```
cfernald: make pointers for OPTIONAL

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: updated comments and renamed to RequesterId and IoMmuBase

eeshanl: done , added additional detail

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: done

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: done

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: Per our conversation in person, we are choosing to drop these extra optional params and will wait to modify this once we make the changes to the IoMmu protocol.

eeshanl: Per our conversation in person, we are choosing to drop these extra optional params and will wait to modify this once we make the changes to the IoMmu protocol.

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: Per our conversation in person, we are choosing to drop these extra optional params and will wait to modify this once we make the changes to the IoMmu protocol.

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
eeshanl: @os-d @cfernald 

https://github.com/microsoft/mu_silicon_intel_tiano/blob/dev/202502/IntelSiliconPkg/Feature/VTd/IntelVTdCoreDxe/IntelVTdCoreDxe.c#L326 turns out a real implementation can return EFI_UNSUPPORTED, so this would no longer work.

I updated the IoMmuLib to have a new function IoMmuIsPresent() . This will return true/false depending on the lib used. And allow the caller to properly use the IoMmu functions within PciRootBridgeIo.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceIo.c
eeshanl: https://github.com/microsoft/mu_silicon_intel_tiano/blob/dev/202502/IntelSiliconPkg/Feature/VTd/IntelVTdCoreDxe/IntelVTdCoreDxe.c#L326 turns out a real implementation can return EFI_UNSUPPORTED, so this would no longer work.

I updated the IoMmuLib to have a new function IoMmuIsPresent() . This will return true/false depending on the lib used. And allow the caller to properly use the IoMmu functions within PciRootBridgeIo.

cfernald (approved):

MdeModulePkg/Bus/Pci/PciBusDxe/PciBus.c
```diff
@@ -286,14 +285,6 @@ PciBusDriverBindingStart (
            );
   }
 
-  if (mIoMmuProtocol == NULL) {
-    gBS->LocateProtocol (
-           &gEdkiiIoMmuProtocolGuid,
-           NULL,
-           (VOID **)&mIoMmuProtocol
-           );
```
os-d, line 294: mu_change tags in this file (missing in a few other places, too)

MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
```diff
@@ -1666,35 +1657,35 @@ RootBridgeIoAllocateBuffer (
 
   RootBridge = ROOT_BRIDGE_FROM_THIS (This);
 
-  // MU_CHANGE [BEGIN] - Allow platform to disallow unprotected DMA access when done before IOMMU install.
-  if (FeaturePcdGet (PcdRequireIommu) && (mIoMmu == NULL)) {
```
os-d, line 1670: Are we entirely reverting the PcdRequireIommu change or is it still being used anywhere? If reverting, can we do that as a separate commit (i.e. do the git revert and check this in as rebase & FF) so that we can track that for the next integration?

os-d (approved):

eeshanl:
Will merge after @kenlautner gives green light on x64 platform test.

MdeModulePkg/Bus/Pci/PciBusDxe/PciBus.c
eeshanl: Fixed MU_CHANGE tags here and throughout.
Commented out the deleted lines and added MU_CHANGE  tags for those as well.

MdeModulePkg/Bus/Pci/PciHostBridgeDxe/PciRootBridgeIo.c
eeshanl: created an individual revert commit for the PcdRequireIommu change and marked PR for rebase & FF.

os-d (approved):

MdeModulePkg/Bus/Pci/PciBusDxe/PciIo.c
```diff
@@ -1007,30 +1007,62 @@ PciIoMap (
...
+    //                            mIoMmuProtocol,
+    //                            PciIoDevice->Handle,
+    //                            *Mapping,
+    //                            IoMmuAttribute
+    //                            );
+    // MU_CHANGE [END]
+
+    // MU_CHANGE [BEGIN] - Use IoMmuLib
+    Status = IoMmuSetAttribute (
+               PciIoDevice->Handle,
```
eeshanl, line 1056: change to PciIoDevice

kuqin12 (approved):

PR #1376, opened 2025-05-25
NVME: Fix an issue with zero-sized queue allocation

joschock:
## Description

The calculation to determine `AdminQueuePairPageCount` results in a zero page count due Submission Queue and Completion Queue NumberOfEntries fields not being initialized. On some systems, this request for a zero page allocation causes the `PciIo->AllocateBuffer()` call to fail with `EFI_INVALID_PARAMETER` which results in failure to properly initialize the NVME driver and prevents NVME functioning.

This PR adds initialization of the relevant fields from the NVME Admin Queue Attributes to ensure that `AdminQueuePairPageCount` is properly initialized to a non-zero value.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Prior to the change, can reproduce `EFI_INVALID_PARAMETER` return code from NvmeControllerInit. After the change, NVME initializes as expected and storage can be used to boot the OS.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressHci.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressHci.c
@@ -1173,8 +1173,10 @@ NvmeControllerInit (
   // Set admin queue entry size to default
   // Note we are using the spec-defined minimum SQES and CQES here.
   //
-  Private->SqData[0].EntrySize = NVME_IOSQES_MIN;
-  Private->CqData[0].EntrySize = NVME_IOCQES_MIN;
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

apop5 (approved):

VivianNK:
This bug was caused from moving WriteNvmeAdminQueueAttributes() to later in the function when using a mergetool to make the changes cleaner. An alternative fix would be to move this function back higher up as it initializes the `NumberOfEntries` from the `NVME_AQA`.

PR #1379, opened 2025-05-28
[SQUASH ON REBASE] Revert basetool MU changes

kuqin12:
## Description

This change reverts 3 commits from last Project MU release branch because they are already upstreamed to EDK2.

The double change is causing the CC and other tools not picking up the intended compiler when doing cross compilation, producing binaries that is the same as the host system instead of target system.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on Linux ARM system and verified functional.

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/C/Makefiles/header.makefile
+++ b/BaseTools/Source/C/Makefiles/header.makefile
@@ -54,13 +54,11 @@ AS ?= $(CLANG_BIN)clang
 AR ?= $(CLANG_BIN)llvm-ar
 LD ?= $(CLANG_BIN)llvm-ld
 else ifeq ($(origin CC),default)
-# MU_CHANGE STARTs: Support GCC prefix
-CC ?= $(GCC_PREFIX)gcc
-CXX ?= $(GCC_PREFIX)g++
-AS ?= $(GCC_PREFIX)gcc
-AR ?= $(GCC_PREFIX)ar
-LD ?= $(GCC_PREFIX)ld
-# MU_CHANGE ENDs
+CC = $(GCC_PREFIX)gcc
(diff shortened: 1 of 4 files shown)


BaseTools/Source/C/Makefiles/header.makefile
```diff
@@ -54,13 +54,11 @@ AS ?= $(CLANG_BIN)clang
...
 LD ?= $(CLANG_BIN)llvm-ld
 else ifeq ($(origin CC),default)
-# MU_CHANGE STARTs: Support GCC prefix
-CC ?= $(GCC_PREFIX)gcc
-CXX ?= $(GCC_PREFIX)g++
-AS ?= $(GCC_PREFIX)gcc
-AR ?= $(GCC_PREFIX)ar
-LD ?= $(GCC_PREFIX)ld
-# MU_CHANGE ENDs
+CC = $(GCC_PREFIX)gcc
```
makubacki, line 57: I assume switching these to unconditional assignments was intentional?

makubacki (approved):

apop5 (approved):

BaseTools/Source/C/Makefiles/header.makefile
kuqin12: Yeah. This is intentionally added because CC was set to the host based gcc during the process and we need it as a cross compiler for this step.

kuqin12: This is actually _the_ change needed to fix the build for ARM based systems.

PR #1383, opened 2025-05-28
Update BaseTool binaries weekly

### LINKED ISSUE #1372: [Feature]: Automate Updating the BaseTools ExtDep
### Feature Overview

Today the BaseTools extdep must be manually updated. This has led to it being 2 years out of date.

### Solution Overview

Like with our other dependencies, we should automatically update the BaseTools ext dep, ideally througha PR opened by a bot, but even if it was just an automatic notification that a new version is available would be better than today.

### Alternatives Considered

_No response_

### Urgency

Medium

### Are you going to implement the feature request?

Someone else needs to implement the feature

No maintainer feedback needed

### Anything else?

_No response_

Javagedes:
## Description

Creates a new recurring workflow that runs weekly. This workflow finds all releases that occurred in the last 7 days, and finds the most recent release for each associated branch (e.g. if three releases happened [`v202502000.0.1`, `dev-v202502000.0.1`, `v202502000.0.2`], we reduce that to only [`v202502000.0.2', 'dev-v202502000.0.1']` because two of them target the same release/202502 branch).

If we do not find any releases, we abort early. Otherwise for each release, we will create a PR for the specific branch, updating the base tools extdep to the most recent release. If a PR already exists for an older version, we will close it.

**Note: In the first execution of this workflow, The PR will change the format of the external dependency to be proper yaml format   rather than json format in a yaml file. You can see the formatted change below. You can also see proper pipeline execution below:

https://github.com/Javagedes/mu_basecore/pull/13
https://github.com/Javagedes/mu_basecore/pull/15

https://github.com/Javagedes/mu_basecore/actions/runs/15329238100
https://github.com/Javagedes/mu_basecore/actions/runs/15328647647

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. Verified that we reduce all releases down to the newest for each branch
2. Verified that we close any old PRs and replace with the new PR
3. Verified that we handle dev vs non-dev releases
4. Verified that the extdep in the PR is valid and stuart_update can download the extdep by copying the updated external dependency (and changing `microsoft` -> `javagedes` and updating to the correct sha).

## Integration Instructions

N/A

### DIFF
--- a/.github/workflows/update-basetools.yml
+++ b/.github/workflows/update-basetools.yml
@@ -85,7 +85,6 @@ jobs:
 
     permissions:
       contents: write
-      pull-requests: write
     
     env:
       ARTIFACT_BASE_URL: https://github.com/${{ github.repository }}/releases/download
@@ -103,19 +102,19 @@ jobs:
       # If testing on a local fork with less strict permissions, you can remove this step, and update token usage to
       # use ${{ secrets.GITHUB_TOKEN }} instead of the app token. You must also add the permission `pull-requests: write`
       # to this job.
-      # - name: Generate Token
-      #   id: app-token
-      #   uses: actions/create-github-app-token@v2
-      #   with:
-      #     app-id: ${{ vars.MU_ACCESS_APP_ID }}
-      #     private-key: ${{ secrets.MU_ACCESS_APP_PRIVATE_KEY }}
-      #     owner: ${{ github.repository_owner }}
+      - name: Generate Token
+        id: app-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ vars.MU_ACCESS_APP_ID }}
+          private-key: ${{ secrets.MU_ACCESS_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
 
       - name: Check if PR already exists
         id: pr-check
         shell: bash
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token }}
           PR_BRANCH_NAME: ${{ env.PR_BRANCH_NAME }}
           REPO: ${{ env.REPO }}
         run: |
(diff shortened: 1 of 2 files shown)


os-d (approved):

Javagedes:
Moving to draft as feedback is to make this a recurring workflow instead of one that fires for each release we do.

.github/workflows/update-basetools.yml
```diff
@@ -0,0 +1,214 @@
...
+      - name: Commit and push changes
+        if: steps.pr-check.outputs.exists != 'true'
+        env:
+          BASETOOLS_PATH: ${{ env.BASETOOLS_PATH }}
+          TAG_NAME: ${{ env.TAG_NAME }}
+          PR_BRANCH_NAME: ${{ env.PR_BRANCH_NAME }}
+        shell: bash
+        run: |
+          git config user.name = "uefibot"
+          git config user.email = "uefibot@microsoft.com"
```
makubacki: Shouldn't this drop the `=`?

```suggestion
          git config user.name "uefibot"
          git config user.email "uefibot@microsoft.com"
```

.github/workflows/update-basetools.yml
```diff
@@ -0,0 +1,214 @@
...
+                  end
+                )
+              })
+            ')
+
+            echo "Mapped releases: $MAPPED_RELEASES"
+            echo "releases=$MAPPED_RELEASES" >> $GITHUB_OUTPUT
+  
+  update:
+    name: "[${{ matrix.release.branchName }}] Update BaseTools Binary to ${{ matrix.release.tagName }}"
```
makubacki, line 78: Was `name` required to be outside the `if` block? If this job is skipped, does the name resolve?

.github/workflows/update-basetools.yml
Javagedes: Yes, good catch. I missed what the commit commiter looked like in github:

image

Will fix.

.github/workflows/update-basetools.yml
Javagedes: The `if` is an actual entry in the yaml, not a macro expansion that completely hides the job, so the location of `name` vs `if` does not matter.

You are correct though, if the matrix is empty, then the name does look weird - but I'm not sure if I can get around that:

image

apop5 (approved):

.github/workflows/update-basetools.yml
makubacki: Okay, I don't think it's a big deal was mostly curious how it would resolve.

makubacki (approved):

PR #1397, opened 2025-06-03
UefiCpuPkg: Remove Nx From EfiBootServicesCode allocation.

apop5:
## Description

MpLib allocates a buffer for Ap Loop code. It manually removes the Nx protection but never applies any other protections to the buffer, leaving it RWX. After the code is copied into the allocated buffer, apply Read Only permissions so that nothing can overwrite it while the APs are executing from it.

In the update to 202502, Edk2 modified their code layout, resulting in the AP buffer being allocated with Nx, having Nx removed, and then never applying any other memory protections.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Executing DxePagingAudit test application on Q35 found one region of memory marked RWX.
Applying these modifications, DxePagingAudit test application no longer reports a RWX buffer.

## Integration Instructions
No integration necessary.

### DIFF
--- a/UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/DxeMpLib.c
@@ -39,14 +39,12 @@ CPU_MP_DEBUG_PROTOCOL  mCpuMpDebugProtocol = {
 
 #define  AP_SAFE_STACK_SIZE  128
 
-CPU_MP_DATA       *mCpuMpData                  = NULL;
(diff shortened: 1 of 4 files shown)


os-d (approved):
edk2 upstream also?

kuqin12 (approved):

apop5:
> edk2 upstream also?

I'll update the PR that is waiting for review there.

PR #1411, opened 2025-06-25
ImageValidation: Update how profiles are determined.

Javagedes:
## Description

For image validation to work, each binary must be associated with a specific profile. This profile sets the "approved" values for each type of validation that is performed. During validation, there are two ways in which we determine the profile for the binary.

1. When the binary is associated with an INF, whether it be a pre-compiled binary in a [Binaries] section of the INF, or a typical INF with source files that the build tools compile, the profile is determined by parsing the INF for the `MODULE_TYPE`
2. When the binary is pre-compiled and placed manually in the FDF using a `FILE <type> = <guid> { ... }` definition, we use the `<type>` as the profile.

This PR specifically fixes an issue in the second scenario. In this scenario, one of the `<type>`'s is `DRIVER`. The issue is that `DRIVER` is actually fairly vague, as it could theoretically be a `DXE_DRIVER` or a `DXE_RUNTIME_DRIVER`. Due to the fact that we should be as specific as possible, in this scenario, we now attempt to parse the subsystem out of the PE/COFF header and use that to determine if it should be a `DXE_DRIVER` or a `DXE_RUNTIME_DRIVER`.

We also check if it is a `UEFI_APPLICATION`, however this is as a precaution. Theoretically, if someone was doing this, they should be using the `FILE UEFI_APPLICATION` in the FDF instead.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Locally verified that a `FILE DRIVER` entry successfully determines it's profile by parsing the subsystem

## Integration Instructions

N/A

### DIFF
--- a/.pytool/Plugin/ImageValidation/ImageValidation.py
+++ b/.pytool/Plugin/ImageValidation/ImageValidation.py
@@ -8,7 +8,7 @@
 import os
 import re
 from pathlib import Path
-from pefile import PE
+from pefile import PE, SUBSYSTEM_TYPE
 from edk2toolext.environment.plugintypes.uefi_build_plugin import IUefiBuildPlugin
 from edk2toolext.image_validation import (
     Result, TestManager, TestInterface, TestWriteExecuteFlags,
@@ -19,7 +19,7 @@ from edk2toollib.uefi.edk2.parsers.dsc_parser import DscParser
 from edk2toollib.uefi.edk2.parsers.inf_parser import InfParser
 from edk2toollib.gitignore_parser import parse_gitignore_lines
 import yaml
-from typing import List
+from typing import List, Optional
 import logging
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


apop5 (approved):

.pytool/Plugin/ImageValidation/ImageValidation.py
makubacki, line 1: Was this just broken before?

For example, the types associated with subsystems listed in the config example - https://github.com/microsoft/mu_basecore/blob/dev/202502/.pytool/Plugin/ImageValidation/image_validation.cfg.

makubacki: Please explain the scenario this update addresses - binary integration of these modules types so the impact to consumers is clear.

makubacki (approved):

.pytool/Plugin/ImageValidation/ImageValidation.py
Javagedes: I've better described the scenario and what this PR is doing in the PR description :)

PR #1424, opened 2025-07-02
Fixing a case where the incoming is not a file path

kuqin12:
## Description

When there are file systems that do not belong to `FILEPATH_DEVICE_PATH`, the casting will be incorrect AND could potentially cause page fault. i.e. if the path belongs to a FV enumerated by FvFileSystemDxe driver, the `PathName` is actually a GUID and does not end with 0.

This change will parse out the file path using DevicePathLib and then perform the string concatenation.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This fixed a hang in the QEMU SBSA UEFI shell when searching for the startup.nsh file.

## Integration Instructions

N/A

### DIFF
--- a/ShellPkg/Application/Shell/Shell.c
+++ b/ShellPkg/Application/Shell/Shell.c
@@ -1242,8 +1242,10 @@ LocateStartupScript (
   CHAR16        *TempSpot;
   CONST CHAR16  *MapName;
   UINTN         Size;
+  CHAR16        *ConvertedPath;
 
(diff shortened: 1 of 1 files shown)


kuqin12:
Does this change make sense? More accurately, is the DevicePathLib interface the correct to be invoked here?

apop5:
superseded by #1447

PR #1429, opened 2025-07-08
Update memory map descriptor merging logic to honor the bucket type information

kuqin12:
## Description

The current memory map merging logic in `CoreAddRange` and `MergeMemoryMapByAttribute` only checks the memory type without considering the bucket type. However, the `CoreGetMemoryMap` routine will only report the entries that are entirely inside a specific bucket type. This will confuse the consumer and report false free memory that belongs to special memory buckets.

This change fixes the issue by gating the merging logic by checking the bucket type first. The merging is only allowed either if the bucket type is not one of the special types, or if the bucket type is the same for both entries.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on proprietary hardware.

## Integration Instructions

N/A


MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +151,40 @@ RemoveMemoryMapEntry (
...
+  // Find the bucket type for the incoming memory region.
+  for (BucketType = 0; BucketType < EfiMaxMemoryType; BucketType++) {
+    //
+    // If the number of pages for this memory type is not zero, the input region
+    // better be within the same bucket.
+    //
+    if (mMemoryTypeStatistics[BucketType].Special && (mMemoryTypeStatistics[BucketType].NumberOfPages != 0)) {
+      if ((PhysicalStart >= mMemoryTypeStatistics[BucketType].BaseAddress) &&
+          (PhysicalEnd <= mMemoryTypeStatistics[BucketType].MaximumAddress)) {
+        break;
```
os-d, line 212: Should we add an assert here if the memory region partially intersects a bucket but isn't fully contained?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +151,40 @@ RemoveMemoryMapEntry (
...
+  IN EFI_PHYSICAL_ADDRESS  PhysicalEnd
+  )
+{
+  EFI_MEMORY_TYPE  BucketType;
+
+  // Find the bucket type for the incoming memory region.
+  for (BucketType = 0; BucketType < EfiMaxMemoryType; BucketType++) {
+    //
+    // If the number of pages for this memory type is not zero, the input region
+    // better be within the same bucket.
```
os-d: Update the comment for why the `Special` field is used?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +151,40 @@ RemoveMemoryMapEntry (
   }
 }
 
+EFI_MEMORY_TYPE
```
os-d, line 192: STATIC and function comments?

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -161,14 +205,54 @@ CoreAddRange (
...
+  MEMORY_MAP       *Entry;
+  EFI_MEMORY_TYPE  BucketType;
+  EFI_MEMORY_TYPE  MergeType;
 
   ASSERT ((Start & EFI_PAGE_MASK) == 0);
   ASSERT (End > Start);
 
   ASSERT_LOCKED (&gMemoryLock);
 
+  // Find the bucket type for the incoming memory region.
```
os-d, line 290: Probably need some more comments for this section. Looks like the goal is to split up this range into up to three descriptors: before the bucket, in the bucket, and after the bucket, is that right?

kuqin12: Yeah, also just added a case that an incoming might have all 3 :)

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: I added the assert in the get memory map call instead, which I think is where it really matters

os-d: It would be good to catch here too, though, right? Because this is where things could go wrong, whereas the get memory map call is when things are already messed up.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Not a static function as it is being used in memory protection code as well. Added function declaration in header file.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +151,71 @@ RemoveMemoryMapEntry (
...
+EFI_MEMORY_TYPE
+GetBucketMemoryType (
+  IN EFI_PHYSICAL_ADDRESS  PhysicalStart,
+  IN EFI_PHYSICAL_ADDRESS  PhysicalEnd
+  )
+{
+  EFI_MEMORY_TYPE  BucketType;
+
+  // Find the bucket type for the incoming memory region.
+  for (BucketType = 0; BucketType < EfiMaxMemoryType; BucketType++) {
```
makubacki: An explicit cast might be needed to avoid a warning for inconsistent type assignment in some compilers.

```suggestion
  for (BucketType = (EFI_MEMORY_TYPE)0; BucketType < EfiMaxMemoryType; BucketType++) {
```

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -141,6 +151,71 @@ RemoveMemoryMapEntry (
...
+        break;
+      } else if (((PhysicalStart >= mMemoryTypeStatistics[BucketType].BaseAddress) &&
+                 (PhysicalStart <= mMemoryTypeStatistics[BucketType].MaximumAddress)) ||
+                 ((PhysicalEnd >= mMemoryTypeStatistics[BucketType].BaseAddress) &&
+                 (PhysicalEnd <= mMemoryTypeStatistics[BucketType].MaximumAddress)))
+      {
+        // The start and end overlap the bucket, but not fully inclusive. We should not allow this.
+        DEBUG ((
+          DEBUG_ERROR,
+          "%a: %lx-%lx intersects bucket %d (%lx-%lx)\n",
```
makubacki: To clarify the number maps to a bucket type.

```suggestion
          "%a: %lx-%lx intersects bucket type %d (%lx-%lx)\n",
```

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -161,14 +236,78 @@ CoreAddRange (
...
+  EFI_MEMORY_TYPE  BucketType;
+  EFI_MEMORY_TYPE  MergeType;
 
   ASSERT ((Start & EFI_PAGE_MASK) == 0);
   ASSERT (End > Start);
 
   ASSERT_LOCKED (&gMemoryLock);
 
+  // Find the bucket type for the incoming memory region.
+  for (BucketType = 0; BucketType < EfiMaxMemoryType; BucketType++) {
```
makubacki:
```suggestion
  for (BucketType = (EFI_MEMORY_TYPE)0; BucketType < EfiMaxMemoryType; BucketType++) {
```

MdeModulePkg/Core/Dxe/Mem/Page.c
makubacki: I think you should just put the condition in a "memory bucket validation" function and call it from both places.

makubacki (approved):

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -161,14 +238,98 @@ CoreAddRange (
...
+    //
+    // If the number of pages for this memory type is not zero, the input region
+    // better be within the same bucket. Otherwise, we will handle the ones we care,
```
cfernald: nit: care about

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Also added some comments about the 3 cases we are handling here.

MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -161,14 +238,98 @@ CoreAddRange (
...
   ASSERT_LOCKED (&gMemoryLock);
 
+  // Find the bucket type for the incoming memory region.
+  for (BucketType = (EFI_MEMORY_TYPE)0; BucketType < EfiMaxMemoryType; BucketType++) {
+    //
+    // If the number of pages for this memory type is not zero, the input region
+    // better be within the same bucket. Otherwise, we will handle the ones we care,
+    // the special memory types, in chunks.
+    //
+    if (mMemoryTypeStatistics[BucketType].Special && (mMemoryTypeStatistics[BucketType].NumberOfPages != 0)) {
```
cfernald, line 298: You could simplify this a bit by instead of checking
1. start in, end over
2. else, start before, end in
3. else, start before, end over

You could check

1. Start before, end after start
2. end after, start before end

and recursive the prefix & post fix individually without having to explicitly check for the scenario where both true.

cfernald (approved):

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated. Thanks.

MdeModulePkg/Core/Dxe/Mem/Page.c
kuqin12: Updated as suggested.

apop5 (approved):

os-d (approved):

PR #1430, opened 2025-07-08
[CHERRY-PICK] [REBASE & FF] Don't Allocate Page 0

os-d:
## Description

This PR reverts a Mu memory protections [commit](https://github.com/microsoft/mu_basecore/commit/7d5a58b7168a7ef5b99e41874010913409538a41) that was partially upstreamed to [edk2](https://github.com/tianocore/edk2/commit/91bb5cee36fe52ef012ddbead4f5932b3f9fd862) and then reapplies the Mu commit on top. This is done to avoid changing the edk2 cherry-pick and instead amend the Mu commit to accommodate the new edk2 behavior.

The rest of the Mu commit will be upstreamed to edk2 soon, but this change is taken to further Mu's goal of not allowing page 0 to be allocatable.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested by booting virtual platforms to an OS.

## Integration Instructions

This is not marked as a breaking change because Mu based platforms were always expected to not allocate page 0 and have that be the null detection page. If any platform had disabled this behavior, they would need to stop allocating page 0. This does not affect compatibility mode, which can still map page 0 as required.


MdeModulePkg/Core/Dxe/Mem/Page.c
```diff
@@ -1504,6 +1504,11 @@ CoreInternalAllocatePages (
   // return EFI_NOT_FOUND.
   //
   if (Type == AllocateAddress) {
+    // Page 0 is not allowed to be allocated as it is reserved for null pointer detection
+    if (Start == 0) {
+      return EFI_NOT_FOUND;
```
makubacki, line 1509: Can you update the function description to describe this case?

os-d: Unfortunately, this is already upstreamed to edk2 here: https://github.com/tianocore/edk2/commit/91bb5cee36fe52ef012ddbead4f5932b3f9fd862 and I don't want to add more delta since this is part of my upstreaming work :). I can put up a separate edk2 PR to update the function description if you want.

MdeModulePkg/Core/DxeIplPeim/X64/VirtualMemory.c
makubacki, line 1: Is this logic still needed in `ToSplitPageTable()`?

```c
  // MU_CHANGE START Remove checks to memory protection settings
  if (Address == 0) {
    // IsNullDetectionEnabled () && Address == 0) {
    return TRUE;
  }
```

makubacki, line 1: Should `IsNullDetectionEnabled()` just assert if `BIT0` is set in `PcdNullPointerDetectionPropertyMask`? It seems like that, as a whole, is a deprecated interface in Mu now.

os-d: Yes and no. To the extent that edk2 has that check, the check is still legitimate: we want to split the page table such that page 0 is mapped at 4KB granularity so that we can unmap it to enable null pointer detection.

However, this doesn't really need to be done in PEI, DXE can do this as well. There is maybe some small performance benefit to doing it when building the table instead of later, but it would be miniscule at best and one time.

So, valid yes, but doesn't need to happen here. I will clean this up in the next upstream to edk2 (the rest of this commit).

makubacki (approved):

MdeModulePkg/Core/DxeIplPeim/X64/VirtualMemory.c
os-d: The PCD is removed entirely, so there should be no references to it. Did I accidentally bring one back in?

kuqin12 (approved):

apop5 (approved):

MdeModulePkg/Core/DxeIplPeim/X64/VirtualMemory.c
makubacki: It is commented out in MdeModulePkg.dec. That's sufficient.

MdeModulePkg/Core/Dxe/Mem/Page.c
makubacki: That is fine. Not a big deal.

PR #1431, opened 2025-07-08
[CHERRY-PICK] Add resource descriptor hob v2 for memory attributes usage

Javagedes:
## Description

This change adds the implementation of supporting "resource descriptor hob type 2", which will extends the existing definition of resource descriptor hob with an attribute field.

This field can be used to provide the actual attributes needs to be set for the memory region, instead of relying on the state of the system.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU Q35 and SBSA systems and booted to Windows.

## Integration Instructions

For platforms needing to propagate specific memory attributes (especially the cache attributes) to the DXE phase, `BuildResourceDescriptorWithCacheHob` should be used to produce resource descriptors (v2).

### DIFF
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -2350,7 +2350,7 @@ CoreInitializeMemoryServices (
       //
       Count = 0;
       for (Hob.Raw = *HobStart; !END_OF_HOB_LIST (Hob); Hob.Raw = GET_NEXT_HOB (Hob)) {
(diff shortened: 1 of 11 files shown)


makubacki (approved):

makubacki:
This is needed in the 2405 release branch for platforms that rely on the "release" branch (all except Mu Tiano Platforms) to start making updates for V2 resource descriptor HOBs.

os-d (approved):

kuqin12 (approved):

PR #1436, opened 2025-07-14
[Rebase & FF] Fix USB reset issue

berlin-with0ut-return:
## Description
There has been a page fault observed when unplugging a USB in frontpage in some Intel environments. This PR fixes both bugs associated with the fault.

1. **`Urb->DataMap` use after free**: the `DataMap` and `Data` fields of the URB share pointers to the same memory. Previously, we would free `Data` first, then attempt to unmap `DataMap`, which would lead to a use after free. This PR modifies `XhcFreeUrb` to handle freeing `Data` as well, and to do it after unmapping `DataMap` when necessary.
2. **`XhcUpdateAsyncRequest` and callback**: `XhcMonitorAsyncRequests` contained a subtle bug where sometimes, the URB's `Callback` would be a function that removed or freed the URB (this occurs in error scenarios when something like `StopEndpoint` fails; for example when the USB is unplugged and the endpoint is no longer valid). Since the callback sometimes removed and freed the URB, the subsequent call to `XhcUpdateAsyncRequest` led to a use after free. This PR moves the call to `XhcUpdateAsyncRequest` before the callback, which is safe since `XhcUpdateAsyncRequest` does not depend on the URB state after the callback. 

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Booted to frontpage on Intel device and observed the page fault no longer occurs when unplugging USB externals.

## Integration Instructions
N/A


MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.h
```diff
@@ -1462,7 +1462,8 @@ XhcCreateUrb (
 VOID
 XhcFreeUrb (
   IN USB_XHCI_INSTANCE  *Xhc,
-  IN URB                *Urb
+  IN URB                *Urb,
```
Javagedes: The documentation directly above this function needs to be updated to reflect the new parameter

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -191,7 +191,8 @@ XhcCreateUrb (
 VOID
 XhcFreeUrb (
   IN USB_XHCI_INSTANCE  *Xhc,
-  IN URB                *Urb
+  IN URB                *Urb,
```
Javagedes: The documentation directly above this function needs to be updated to reflect the new parameter

Javagedes (commented):
As we talked about offline, you need MU_CHANGE tags for this. I also think you need to add documentation for each XhcFreeUrb call explaining why each particular call is set to TRUE or FALSE

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -202,6 +203,13 @@ XhcFreeUrb (
     Xhc->PciIo->Unmap (Xhc->PciIo, Urb->DataMap);
   }
 
+  // If freeing URB data, the free must come after `DataMap` is unmapped.
+  // This is because `DataMap` depends on internal state built around `Data`.
+  // If `Data` is freed before `DataMap`, it may lead to a fault when unmapping.
+  if (FreeUrbData && (Urb->Data != NULL)) {
```
os-d: I'm a bit confused here. Why do we only conditionally free `Urb->Data`? We always free `Urb` at the end, won't this leak the `Urb->Data` memory? I see that the existing code didn't do this, but seems like that is an additional bug fix here? Or in the non-freeing cases does `Urb->Data` have a second pointer that lives on?

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -1713,6 +1719,10 @@ XhcMonitorAsyncRequests (
       CopyMem (ProcBuf, Urb->Data, Urb->Completed);
     }
 
+    // The update call must occur before the callback since the callback
+    // may remove and free the URB, leading to a fault.
+    XhcUpdateAsyncRequest (Xhc, Urb);
```
os-d, line 1733: `XhcUpdateAsyncRequest` updates `Urb` including some of the fields passed to the callback: https://github.com/tianocore/edk2/blob/bdca3681e90b924bfa89b378c2687eef4e22db91/MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c#L247. Is that okay? Do we need to save off these values and pass them to the callback?

Looking at an example callback, it looks like we would need to: https://github.com/tianocore/edk2/blob/bdca3681e90b924bfa89b378c2687eef4e22db91/MdeModulePkg/Bus/Usb/UsbMouseDxe/UsbMouse.c#L744.

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -1736,8 +1746,6 @@ XhcMonitorAsyncRequests (
     if (ProcBuf != NULL) {
       gBS->FreePool (ProcBuf);
     }
-
```
os-d, line 1707: While changing here, can you fix the grammar in the comment above RestoreTpl in line 1738?

os-d (commented):
I think this should be a [REBASE & FF] PR (i.e. structure this into two commits for each separate bug and then choose the rebase option when completing the PR). This helps separate out the different bugs, but more importantly, we should take this to edk2 after checking in here and it will need to be in that form there.

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
berlin-with0ut-return: so i initially took out the unconditional free because it would cause a crash.

however, based on what i see in the code, Urb and Urb->data should always have a 1:1 mapping, and Urb->data should be freed when URB is. it warrants further investigation at least to see what's accessing Urb->data after URB is freed, if any.

berlin-with0ut-return: i've confirmed there is a crash if we unconditionally free `Urb->data`. i believe this is because the sometimes the payload isn't managed or allocated at the time of URB creation; for example if you look at `XhcTransfer`, the data is passed in as an argument and the `Urb->data` field gets assigned, but the data is allocated as part of calling `XhcSyncInterruptTransfer` in the USB protocol.

as such i think keeping the existing behavior, where `FreeUrb` only frees data when it directly manages it, is correct

os-d: I think in this case we shouldn't mingle freeing the Urb with freeing the data buffer, then, if they have two separate lifetimes. I think that the callers that do want to free the data buffer after freeing the Urb should manage that themselves. If the Urb code does not manage data buffers, we shouldn't have it partially manage them.

Now, we of course need to maintain the ordering, so we could create a new wrapper function to `XhcFreeUrb` that calls `XhcFreeUrb` and then frees data (storing off the ptr ahead of time) or just in the existing places just store the ptr, then call free after, with comments on both.

I would probably lean towards the second option, where the creator does the freeing. I would add comments to `XhcFreeUrb` indicating that the caller is responsible for freeing the data buffer (as today) and that it must be done after calling `XhcFreeUrb`. Then I would add comments in the owner of the data buffer that it must free it after calling `XhcUrbFree`.

MdeModulePkg/Bus/Pci/XhciDxe/Xhci.c
```diff
@@ -837,7 +837,13 @@ XhcTransfer (
   }
 
   Xhc->PciIo->Flush (Xhc->PciIo);
+  // MU_CHANGE [BEGIN] - Fix USB reset issue
+  //
+  // Do not free URB data, since it is passed in as an external argument
+  // and allocated and managed by the caller.
+  //
   XhcFreeUrb (Xhc, Urb);
+  // MU_CHANGE [END] - Fix USB reset issue
```
os-d: nit: Could move this end to the end of the comment, so just future changes don't make this cherry-pick fail

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -106,7 +106,13 @@ XhcCmdTransfer (
     Status = EFI_SUCCESS;
   }
 
+  // MU_CHANGE [BEGIN] - Fix USB reset issue
+  //
+  // Do not free URB data, since `XhcCreateCmdTrb` does not allocate any data
+  // and the `Data` field is not used in command transfers.
+  //
   XhcFreeUrb (Xhc, Urb);
+  // MU_CHANGE [END] - Fix USB reset issue
```
os-d: same nit.

os-d (approved):
As noted before, I think this should be made into a REBASE & FF PR with two commits: the change to free Urb->Data and the change to move the update before the callback, as these are two distinct bug fixes (and will need to be taken to edk2 as such).

Javagedes (approved):

makubacki:
@berlin-with0ut-return, if this is "Rebase & FF", please rebase locally to drop the merge commit.

image

MdeModulePkg/Bus/Pci/XhciDxe/XhciSched.c
```diff
@@ -1750,6 +1764,21 @@ XhcMonitorAsyncRequests (
...
+    //
+    cbCompleted = Urb->Completed;
+    cbResult    = Urb->Result;
+    cbContext   = Urb->Context;
+
+    //
+    // The update call must occur before the callback since the callback
+    // may remove and free the URB, leading to a fault.
+    //
+    XhcUpdateAsyncRequest (Xhc, Urb);
```
os-d, line 1733: One thing that gives me pause is that your change is undoing a change from 13 years ago that had the update call in exactly the same spot :)

https://github.com/tianocore/edk2/commit/a50f7c4c0985f1659d4b24e94f6d3aee9f6922dd#diff-09ccf56838e59a81ec03a723d04b42a97abe9695f4e64ab2b907157570875a1cL1222-R1311

I think this is okay, because the data is copied. Completed and Result are uints, so that's okay, but context is a void *. If Urb is deleted or changed, the callback could then refer to this bad context. Maybe this is okay, since the context is caller managed, but also, if it is keeping some state between transfers that may get corrupted.

I feel like the pattern in the XHCI driver is bad overall. The callback shouldn't be allowed to delete the Urb, imo. Or we shouldn't use the same Urb for the next transfer. But, that would be a large lift to fix.

One other concern, though this is more edge case, is now that the Urb is prepped for the next transfer before the callback, is it possible we get out of order transactions? The TPL should protect against that, although then we call the callback at a lower TPL, but that is what has always been done, so I guess we just say oh well, hope it works.

TL;DR: Without a lot of extra plumbing, I don't think there is a better solution unless we create a new URB that is a copy of the old one and keep the function call order the same. Even that I'm not confident on, because the fields may change. So I think this is okay, but if we had more confidence on the context not changing, I would be more comfortable.

os-d (approved):

berlin-with0ut-return:
Closing since this was merged into edk2 and needs to be cherry-picked

PR #1437, opened 2025-07-15
[SQUASH ON REBASE] Fix bucket type matching as soon as the bucket is ready

kuqin12:
## Description

The merging logic check could be voided if the incoming memory is non-special between 2 special buckets.

This change removes the bypass when the incoming memory is non-special. This change also moves the bucket size to be set before the bucket memory is freed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on proprietary hardware platforms.

## Integration Instructions

N/A


apop5 (approved):

os-d (commented):
nit: This should be `SQUASH ON REBASE` right? Doesn't matter too much since you have the edk2 PR up.

Do we not need the other change added to the edk2 PR?

os-d (approved):

PR #1487, opened 2025-08-27
UefiCpuPkg: X64 MpLib: Spinlock RestoreVolatileRegisters

os-d:
## Description

All APs share the same GDT and segment descriptors. Each time an AP is started up, the wake up function restores the GDT and other system registers with the RestoreVolatileRegisters function.

A General Protection fault has recently been observed on multiple generations of Intel silicon when the APs have been left in an mwait loop by CpuMpPei and the initial wake up state in DXE is not init-sipi-sipi. The cores all come up at the same time, where they attempt to restore the GDT from the same address. However, the TSS is particular cannot be loaded simultaneously: when the busy bit is set in the TSS (which occurs when loaded by the ltr instruction in RestoreVolatileRegisters) the ltr instruction will fail with a GP fault with an exception context of 0x40, the entry in the GDT corresponding to the TSS.

In other scenarios, the cores will not come up at exactly the same time, either because they've been serialized through an init-sipi-sipi or the timing is right enough that they load the TSS separately. RestoreVolatileRegisters does attempt to clear the busy bit of the TSS before setting it via ltr, but this is not protected by a lock and so the other APs can execute ltr inbetween an AP setting the busy bit to 0 and then doing ltr.

This commit fixes this behavior by using the MpLock, shared among all APs (and the BSP) to serialize execution of RestoreVolatileRegisters. This resolves the bug seen. On systems where the timing was not hitting this, it is essentially a no-op.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on multiple generations of Intel physical platforms that were experiencing the mentioned bug and saw it resolve them.

## Integration Instructions

N/A.

### DIFF
--- a/UefiCpuPkg/Library/MpInitLib/MpLib.c
+++ b/UefiCpuPkg/Library/MpInitLib/MpLib.c
@@ -32,17 +32,21 @@ SaveVolatileRegisters (
   OUT CPU_VOLATILE_REGISTERS  *VolatileRegisters
   );
 
+// MU_CHANGE START: Spinlock RestoreVolatileRegisters
+
 /**
   Restore the volatile registers following INIT IPI.
 
   @param[in]  VolatileRegisters   Pointer to volatile resisters
-  @param[in]  IsRestoreDr         TRUE:  Restore DRx if supported
-                                  FALSE: Do not restore DRx
+  @param[in]  MpLock              Pointer to spin lock to synchronize access to GDT
 **/
 VOID
 RestoreVolatileRegisters (
-  IN CPU_VOLATILE_REGISTERS  *VolatileRegisters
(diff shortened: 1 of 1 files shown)


makubacki (approved):

apop5 (approved):

kuqin12 (approved):

makubacki:
@os-d, you need an Uncrustify push before this can complete.

makubacki:
> @os-d, you need an Uncrustify push before this can complete.

I went ahead and rebased and pushed the Uncrustify update.

makubacki (approved):

apop5 (approved):

PR #1497, opened 2025-09-08
tools_def.template: Remove /FILEALIGN from X64 and AARCH64 CLANGPDB DLINK Flags

makubacki:
## Description

`FileAlignment` in the PE/COFF Specification states:

```
The alignment factor (in bytes) that is used to align the raw data of
sections in the image file. The value should be a power of 2 between
512 and 64 K, inclusive. The default is 512. If the SectionAlignment
is less than the architecture's page size, then FileAlignment must
match SectionAlignment.
```

Although in UEFI we often use sizes less than 512, that is outside the allowed spec-defined range. Today, /FILEALIGN is not set in X64 VS DLINK flags and is set to a smaller value (often `32`) for IA32 VS DLINK flags.

This change modifies the tools_def.template value to:

1. Be PE/COFF spec-compliant by not explicitly specifying a size less than the spec-supported minimum.
2. Make the `/FILELALIGN` value consistent for X64 modules between the VS and CLANGPDB flags.

Note: A smaller `/FILEALIGN` value is used to save flash space, particularly for uncompressed images that reside directly in flash.

Platforms with 64-bit PEIMs, can explicitly set `/FILEALIGN` for their active toolchain as needed. That is a platform decision to opt out of PE/COFF Specification compatibility to achieve platform size requirements.

This is particularly important because some tools that operate on PE/COFF images, strictly follow the specification behavior. With this change, by default CLANGPDB images will be compliant with those expectations.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Build CLANGPDB X64 images and check the file alignment in the .efi binary

## Integration Instructions

- Review the information in this PR description. If a smaller file alignment value is needed, set that in a platform location such as `[BuildOptions]` in a DSC file.
- This is marked as a breaking change because a larger file alignment value for X64 modules may cause more space flash to be consumed if using CLANGPDB and platform build flags might need to be adjusted for section and file alignment to be compatible with linker requirements.

### DIFF
--- a/BaseTools/Conf/tools_def.template
+++ b/BaseTools/Conf/tools_def.template
@@ -26,8 +26,9 @@
 # 3.20 - Move GccLto files to a tools path to be more repository layout agnostic
 # 3.21 - Can't have comment inline as it breaks concatenation see - # MU_CHANGE - Move GccLto to tools dir for better alignment
 # 3.22 - VS2022 toolchain incorrectly defined for ASL
+# 3.23 - MU_CHANGE - Remove /FILEALIGN from *_CLANGPDB_*_DLINK_FLAGS
 #
-#!VERSION=3.02
+#!VERSION=3.23
 
 IDENTIFIER = Default TOOL_CHAIN_CONF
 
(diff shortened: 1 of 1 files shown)


apop5 (approved):

makubacki:
Note: I am taking this to edk2. Posting here in parallel since we have some use cases that could benefit from using spec-compliant file alignment.

os-d (approved):
I'm okay with this change, however, it does seem like the PE/COFF spec (such as it is) should be updated to match what MSVC allows, which is any power of 2 https://learn.microsoft.com/en-us/cpp/build/reference/filealign?view=msvc-170. Since the PE/COFF spec is more or less a description of what MSVC does. That being said, it can be a challenge to update the PE/COFF spec.

In general, I do think it is valuable to have the smallest possible file alignment to not waste flash space, even for DXE/MM. I'm not aware of any benefit of greater file alignment, except for in XIP scenarios.

I think that the updates to external modules is a good solution and this is a reasonable approach to also make things work in the meantime, but if we cared to go to the effort, the PE/COFF spec update seems like a good change.

makubacki:
> I'm okay with this change, however, it does seem like the PE/COFF spec (such as it is) should be updated to match what MSVC allows, which is any power of 2 https://learn.microsoft.com/en-us/cpp/build/reference/filealign?view=msvc-170. Since the PE/COFF spec is more or less a description of what MSVC does. That being said, it can be a challenge to update the PE/COFF spec.
> 
> In general, I do think it is valuable to have the smallest possible file alignment to not waste flash space, even for DXE/MM. I'm not aware of any benefit of greater file alignment, except for in XIP scenarios.
> 
> I think that the updates to external modules is a good solution and this is a reasonable approach to also make things work in the meantime, but if we cared to go to the effort, the PE/COFF spec update seems like a good change.

I've started some conversations about the PE/COFF Spec change. It is mostly reflecting "OS expectations" which as you've noted is not necessarily what should drive the spec as producers use it for various purposes.

PR #1499, opened 2025-09-09
Change the scope of BmpCheckPlugin to global

pohancn:
## Description

Change the scope of BmpCheckPlugin to global. 
Requiring platforms to include the 'surfacefamily' scope in order to use this plug-in is unreasonable.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified that BmpCheckPlugin will be loaded by include scope "global".

## Integration Instructions
This will enable a new post build plugin, where it was previously only scoped to surfacefamily.

If this plugin breaks in a project, it means that the BMP file the project is using is either malformed or that it contains excessive data that is not part of the BMP. Close examination of the BMP is recommend, or regeneration of the BMP.

### DIFF
--- a/BaseTools/Plugin/BmpCheck/BmpCheck_plug_in.json
+++ b/BaseTools/Plugin/BmpCheck/BmpCheck_plug_in.json
@@ -1,5 +1,5 @@
 {
-  "scope": "surfacefamily",
+  "scope": "global",
   "name": "Bmp Image Encoding Check",
   "module": "BmpCheckPlugin"
 }


makubacki (approved):
@spbrogan will have more context of the prior scope restriction, but I agree with the request for it to be broadened.

apop5 (approved):

PR #1510, opened 2025-09-22
Remove MemoryAllocationLib for ArmFfaSecLib

kuqin12:
## Description

UefiPayloadEntry module provides memory allocation related APIs through its own file, instead of going through the library interface. Adding a library linkage will enforce a failure on the build if UefiPayloadEntry is used for SEC phase on ARM64 platforms.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on ARM64 platform and fixed the build break.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.inf
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaSecLib.inf
@@ -33,7 +33,6 @@
   BaseMemoryLib
   DebugLib
   HobLib
-  MemoryAllocationLib
   TimerLib  # MU_CHANGE
 
 [Pcd]


makubacki:
I consider this a bug in `UefiPayloadEntry`. Why does it hijack the `MemoryAllocationLib` interface and not link a proper library instance?

kuqin12:
> I consider this a bug in `UefiPayloadEntry`. Why does it hijack the `MemoryAllocationLib` interface and not link a proper library instance?

Yeah, I was thinking about the same earlier. Hold off on this one for now.

PR #1515, opened 2025-09-24
[REBASE & FF] Prevent AP GP Fault

os-d:
## Description

This PR reverts a MU CHANGE to prevent the AP deadlock with a spinlock. This was later determined to not be necessary and instead separate a GDT/TSS is used for each AP, in addition to forcing a graceful transition to those.

This PR also reverts and reapplies a MU_CHANGE as part of that MU_CHANGE was moved to the above new commit and will be upstreamed to edk2 separately. In addition, another piece of this memory protections MU_CHANGE was already changed in https://github.com/microsoft/mu_basecore/commit/2e259b70bc1b100fbb95ecba3651b523f489810b, so that is also dropped.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical Intel client and server platforms that were experiencing the AP deadlock.

## Integration Instructions

N/A.

### DIFF
--- a/UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuDxeSmm.inf
+++ b/UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuDxeSmm.inf
@@ -103,7 +103,6 @@
   CpuPageTableLib
   MmSaveStateLib
   SmmCpuSyncLib
-  MmMemoryProtectionHobLib                   ## MU_CHANGE
   PanicLib                                   ## MU_CHANGE
 
 [Protocols]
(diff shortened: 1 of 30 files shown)


apop5 (approved):

UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuDxeSmm.inf
```diff
@@ -103,8 +103,8 @@
   CpuPageTableLib
   MmSaveStateLib
   SmmCpuSyncLib
+  PanicLib
```
makubacki: Did `PanicLib` intentionally drop the `MU_CHANGE` tag? It is not in https://github.com/tianocore/edk2/blob/master/UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuDxeSmm.inf.

makubacki (approved):

UefiCpuPkg/PiSmmCpuDxeSmm/PiSmmCpuDxeSmm.inf
os-d: That was accidental, I have added it back, thanks.

apop5 (approved):

PR #1516, opened 2025-09-30
[CHERRY-PICK] workflow: Update BaseTool extdep weekly

Javagedes:
## Description

Creates a new recurring workflow that runs weekly. This workflow finds all releases that occurred in the last 7 days, and finds the most recent release for each associated branch (e.g. if three releases happened [v202502000.0.1, dev-v202502000.0.1, v202502000.0.2], we reduce that to only [v202502000.0.2', 'dev-v202502000.0.1'] because two of them target the same release/202502 branch).

If we do not find any releases, we abort early. Otherwise for each release, we will create a PR for the specific branch, updating the base tools extdep to the most recent release. If a PR already exists for an older version, we will close it.

Cherry-pick of #1383

An additional commit after this cherry-pick is applied to remove the dev branch logic.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A

### DIFF
--- a/.github/workflows/update-basetools.yml
+++ b/.github/workflows/update-basetools.yml
@@ -13,8 +13,8 @@ jobs:
   # This job pulls all releases made in the last 7 days (This job runs weekly), and finds the newest unique tag, which
   # is then used to generate a matrix for the next job, responsible for updating the BaseTools binaries for each branch.
   #
-  # By "unique tag", it is meant that the semver versioning scheme is ignored, and only the `dev` and date prefix is used.
-  # e.g. for `dev-v2025020000.0.2`, it is finding the newest unique `dev-v202502`.
+  # By "unique tag", it is meant that the semver versioning scheme is ignored, and only the date prefix is used.
+  # e.g. for `v2025020000.0.2`, it is finding the newest unique `202502`.
   check:
     name: Find New Releases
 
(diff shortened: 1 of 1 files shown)


apop5 (approved):

.github/workflows/update-basetools.yml
```diff
@@ -0,0 +1,214 @@
...
+  workflow_dispatch: # Manual trigger
+
+env:
+  REPO: ${{ github.repository }}
+
+jobs:
+  # This job pulls all releases made in the last 7 days (This job runs weekly), and finds the newest unique tag, which
+  # is then used to generate a matrix for the next job, responsible for updating the BaseTools binaries for each branch.
+  #
+  # By "unique tag", it is meant that the semver versioning scheme is ignored, and only the `dev` and date prefix is used.
```
makubacki: Can this be simplified by removing the `dev-` support now? It seems like this change to move it into the `release` branch is a good time to drop it.

Javagedes: It can; my plan was to do it in a follow up PR so that this can simply be a cherry-pick of an already reviewed PR. But I can do it now if you prefer.

makubacki: I think it's a little cleaner to just get it in the first time in this case.

Javagedes: I've removed the dev branch logic.

makubacki (approved):

makubacki:
Thanks for removing the "dev-" comment. Was posting that when you pushed the update.

apop5 (approved):

PR #1519, opened 2025-10-06
MdeModulePkg/Variable: Move RT cache buffer allocation to DXE

makubacki:
## Description

Commit `d8f513de3e3ef228af7e6facf0ad3e35c3224032` (an edk2 commit) added runtime memory allocation to VariablePei. That will cause runtime memory bucket fragmentation affecting hibernate stability as these runtime allocations are separate from the main runtime memory type bucket allocation made by the DXE Core.

To preserve the existing functionality but allow the runtime buffer to come from the DXE bucket, the allocations in PEI are changed to boot services data allocations and the contents are moved to the runtime memory buffers (and those pages unblocked to MM) in the DXE variable driver after the bucket has been allocated by the DXE core.

---

> Note: I will upstream this to edk2 but it is urgently needed in Project Mu.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- CI
- Boot to EFI shell on QEMU Q35 and verify allocations come from the DXE allocated bucket buffer

## Integration Instructions

- N/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
@@ -1535,6 +1535,27 @@ InitVariableStoreHeader (
   }
 }
 
+//
+// MU_CHANGE [BEGIN] - Add helper function for runtime cache cleanup
+//
+
+/**
+  Invalidates all runtime cache pointers to prevent access to boot services memory.
+
+**/
+VOID
+InvalidateAllRuntimeCaches (
+  VOID
+  )
+{
+  mVariableRtCacheInfo.CacheInfoFlagBuffer        = 0;
+  mVariableRtCacheInfo.RuntimeHobCacheBuffer      = 0;
+  mVariableRtCacheInfo.RuntimeNvCacheBuffer       = 0;
+  mVariableRtCacheInfo.RuntimeVolatileCacheBuffer = 0;
+}
+
+// MU_CHANGE [END] - Add helper function for runtime cache cleanup
+
 /**
   Initialize the runtime variable cache related content.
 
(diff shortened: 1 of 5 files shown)


makubacki:
I was originally planning to take this straight to edk2, but it's needed here now. I'll add `MU_CHANGE` tags before completing it.

makubacki:
> I was originally planning to take this straight to edk2, but it's needed here now. I'll add `MU_CHANGE` tags before completing it.

Done - `MU_CHANGE` tags added

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
```diff
@@ -1577,9 +1586,123 @@ InitVariableCache (
...
+                      &RuntimeBufferPhysicalAddress
+                      );
+      if (!EFI_ERROR (Status)) {
+        RuntimeBuffer = (VOID *)(UINTN)RuntimeBufferPhysicalAddress;
+        CopyMem (RuntimeBuffer, (VOID *)(UINTN)mVariableRtCacheInfo.CacheInfoFlagBuffer, sizeof (CACHE_INFO_FLAG));
+        mVariableRtCacheInfo.CacheInfoFlagBuffer = (UINTN)RuntimeBuffer;
+
+        Status = MmUnblockMemoryRequest (RuntimeBufferPhysicalAddress, Pages);
+        if ((Status != EFI_UNSUPPORTED) && EFI_ERROR (Status)) {
+          return Status;
```
os-d: If I'm reading this correctly, in the original implementation if we failed here, the mVariableRtCacheInfo.*Buffer pointer would not be set (either null or worse garbage). In this case, it will point back to the boot services buffer. Would it be safer to null the pointers in cases of failures (here and all the other error cases) so that runtime service code doesn't try to access boot services data? This is just an optimization, right, so we could meaningfully make progress if this was cleaned up?

If we do, then we should either continue and try to cleanup the other buffers, or null all buffers (I don't know if they have to be used as a group or not).

os-d (approved):

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
makubacki: Addressed in https://github.com/microsoft/mu_basecore/pull/1519/commits/a88d5062cab74e247ab4af95e356773a2d290b9f

apop5 (approved):

PR #1526, opened 2025-10-11
Update FFA run interface to intake context as input argument

kuqin12:
## Description

The original implementation of ArmFfaLibRun does not return register values when invoked. However, FFA_RUN is normally some operations that requires retry, i.e. after FFA_YIELD or FFA_INTERRUPTED. Thus the context returned by FFA_RUN will be eaten if this is not returned.

The change added an extra input argument to potentially return the output from monitor calls.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on QEMU SBSA and booted to UEFI shell.

## Integration Instructions

One needs to pass in a pointer to context registers if the return values are needed.


Raymond-MS (approved):
LGTM!

liqiqiii (approved):

cfernald (approved):

PR #1546, opened 2025-10-27
Add Repo Remote Check Plugin

makubacki:
## Description

Adds a plugin that is not active by default but can be opted into that validates git repositories against remote commit criteria before the build starts. It allows you to define rules that check whether the current HEAD commit in specified repositories is present or absent in specified remote URLs.

This is useful for enforcing that your local repositories are using commits from the correct upstream sources, and not from incompatible forks or remotes. For example, some workspaces might use a "public" and an "internal" upstream and need to verify that both work as expected and are set to the appropriate upstream when used for a given platform build.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

- Verify that the plugin does not run unless the scope is set
- Verify JSON and YAML files can be parsed
- Verify repo check logic returns expected results

## Integration Instructions

- See the plugin readme file

### DIFF
--- a/.pytool/Plugin/RepoRemoteCheck/RepoRemoteCheck_plug_in.yaml
+++ b/.pytool/Plugin/RepoRemoteCheck/RepoRemoteCheck_plug_in.yaml
@@ -1,0 +1,11 @@
+## @file
+# Checks that a git repository meets certain remote URL criteria.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+{
+    "scope": "remote_repo_check",
+    "name": "Repository Remote Check Plugin",
+    "module": "RepoRemoteCheck"
+}
--- a/.pytool/Plugin/RepoRemoteCheck/ReadMe.md
+++ b/.pytool/Plugin/RepoRemoteCheck/ReadMe.md
@@ -1,0 +1,185 @@
+# Repository Remote Check - Pre-Build Git Repository Validation Plugin
+
+This plugin validates git repositories against remote commit criteria before the build starts. It allows you to define
+rules that check whether the current HEAD commit in specified repositories is present or absent in specified remote
+URLs.
+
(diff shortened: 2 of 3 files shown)


.pytool/Plugin/RepoRemoteCheck/RepoRemoteCheck_plug_in.yaml
```diff
@@ -0,0 +1,11 @@
...
+# Checks that a git repository meets certain remote URL criteria.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+{
+    "scope": "remote_repo_check",
+    "name": "Repository Remote Check Plugin",
+    "module": "RepoRemoteCheck"
+}
```
os-d: nit: newline at EOF

os-d (approved):

apop5 (approved):

berlin-with0ut-return:
why do we not enable this by default? it seems that this is meant to catch mistakes, but if someone isn't fully aware of the secure/nonsecure repo rules they probably also wouldn't know to turn on this plugin

makubacki:
> why do we not enable this by default? it seems that this is meant to catch mistakes, but if someone isn't fully aware of the secure/nonsecure repo rules they probably also wouldn't know to turn on this plugin

Two main reasons - (1) not every Mu-based platform repo needs this such as mu_tiano_platforms (2) in repos that do need it, the platform owner must determine the best policy for their repo using the scope. This can range from "always run" to "run under these specific conditions such as when running in server-side CI to not hamper local development in a submodule".

The plugin only needs to be enabled once by the platform owner. After that, it will catch developer mistakes. This also follows the philosophy that it is better to enforce an explicit and effective repo check policy as opposed to an implicit and incorrect policy.

os-d (approved):

PR #1547, opened 2025-10-28
MdeModulePkg: Add gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid for deterministic UniqueId support in NonDiscoverablePciDeviceDxe

eeshanl:
## Description

MdeModulePkg: Add gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid for deterministic UniqueId support in NonDiscoverablePciDeviceDxe

With this new protocol, a platform can register a UniqueId to be used to identify the PciIo location.
For example, if a device registers with the NonDiscoverableDeviceRegistrationLib, they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice.
NonDiscoverablePciDeviceDxe will consume this UniqueId if found and the device's UniqueId to this value.
If not found, it will back up to the prior method of assigning the UniqueId via a static counter.
The platform must ensure all UniqueIds assigned are unique across the firmware.

Platforms may need to define their own UniqueIds through this method because the current UniqueId that gets used in NonDiscoverablePciDeviceDxe is a static counter, which can lead to non-deterministic value's being returned by PciIo->GetLocation().
In order to have a deterministic UniqueId returned by PciIo->GetLocation(), this protocol can be used to specify a specific UniqueId for a given Handle.
For example, if a platform needs to Locate a specific NonDiscoverable PciIo protocol, they could use the determinstic UniqueId assigned through this protocol to find the expected PciIo instance.
This protocol allows us to have a 1-1 defined mapping between a platform defined UniqueId and the NonDiscoverable PciIo protocol instance.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical device

## Integration Instructions

When assigning a unique ID via this protocol, the platform should ensure that the ID is unique across all instances of this protocol across the platform.
The UniqueId must be in the range of:
    MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID

Usage example:
```
EFI_HANDLE Handle = NULL;
UINTN      Size; // NonDiscoverableDevice Size
VOID       *Base; // NonDiscoverableDevice Base

Status = RegisterNonDiscoverableMmioDevice (
            NonDiscoverableDeviceTypeAhci,
            NonDiscoverableDeviceDmaTypeCoherent,
            NULL,
            &Handle,
            1,
            Base,
            Size
          );

NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL * UniqueIdProtocol = (NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL*) AllocateZeroPool (sizeof (NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL));
if (UniqueIdProtocol == NULL) {
  return EFI_OUT_OF_RESOURCES;
}
UniqueIdProtocol->Revision = NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION;
UniqueIdProtocol->UniqueId = 0x1000; // MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID

Status = gBS->InstallMultipleProtocolInterfaces (&Handle, &gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid, UniqueIdProtocol, NULL);
```

### DIFF
--- a/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
+++ b/MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
@@ -9,8 +9,11 @@
 #include "NonDiscoverablePciDeviceIo.h"
 
 #include <Protocol/DriverBinding.h>
+// MU_CHANGE [BEGIN]
+#include <Protocol/NonDiscoverableDeviceUniqueId.h>
 
-#define MAX_NON_DISCOVERABLE_PCI_DEVICE_ID  (32 * 256)
+// #define MAX_NON_DISCOVERABLE_PCI_DEVICE_ID  (32 * 256)
+// MU_CHANGE [END]
 
 STATIC UINTN           mUniqueIdCounter = 0;
 EFI_CPU_ARCH_PROTOCOL  *mCpu;
@@ -147,9 +150,14 @@ NonDiscoverablePciDeviceStart (
 {
   NON_DISCOVERABLE_PCI_DEVICE  *Dev;
   EFI_STATUS                   Status;
+  // MU_CHANGE [BEGIN]
+  NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL  *UniqueIdProtocol;
 
-  ASSERT (mUniqueIdCounter < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID);
-  if (mUniqueIdCounter >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID) {
+  // ASSERT (mUniqueIdCounter < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID);
+  // if (mUniqueIdCounter >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID) {
+  ASSERT (mUniqueIdCounter < (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2));
+  if (mUniqueIdCounter >= (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2)) {
+    // MU_CHANGE [END]
     return EFI_OUT_OF_RESOURCES;
   }
 
@@ -187,7 +195,44 @@ NonDiscoverablePciDeviceStart (
     goto CloseProtocol;
   }
 
-  Dev->UniqueId = mUniqueIdCounter++;
+  // MU_CHANGE [BEGIN]
+  Status = gBS->HandleProtocol (DeviceHandle, &gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid, (VOID **)&UniqueIdProtocol);
+  if (Status == EFI_SUCCESS) {
+    if (UniqueIdProtocol->Revision != NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: UniqueId protocol revision mismatch. Expected %d, got %d\n",
+        __func__,
+        NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION,
+        UniqueIdProtocol->Revision
+        ));
+      Status = EFI_INCOMPATIBLE_VERSION;
+      goto CloseProtocol;
+    }
+
+    if ((UniqueIdProtocol->UniqueId < (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2)) || (UniqueIdProtocol->UniqueId >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: UniqueId Protocol failed. UniqueId=0x%llx must be greater than or equal to 0x%llx and less than 0x%llx.\n",
+        __func__,
+        UniqueIdProtocol->UniqueId,
+        (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2),
+        MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
+        ));
+      Status = EFI_INCOMPATIBLE_VERSION;
+      goto CloseProtocol;
+    }
+
+    // Assign UniqueId from protocol if available.
+    Dev->UniqueId = UniqueIdProtocol->UniqueId;
+    DEBUG ((DEBUG_VERBOSE, "%a: UniqueId=0x%llx\n", __func__, Dev->UniqueId));
+  } else {
+    // Otherwise, backup and assign from counter.
+    Dev->UniqueId = mUniqueIdCounter++;
+    DEBUG ((DEBUG_VERBOSE, "%a: UniqueId Protocol failed. Backup to mUniqueIdCounter=0x%llx\n", __func__, Dev->UniqueId));
+  }
+
+  // MU_CHANGE [END]
 
   return EFI_SUCCESS;
 
--- a/MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
+++ b/MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
@@ -1,0 +1,74 @@
+/** @file NonDiscoverableDeviceUniqueId.h
+  With this new protocol, a platform can register a UniqueId to be used to identify the PciIo location.
+  For example, if a device registers with the NonDiscoverableDeviceRegistrationLib,
+  they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice.
+  NonDiscoverablePciDeviceDxe will consume this UniqueId if found and the device's UniqueId to this value.
+  If not found, it will back up to the prior method of assigning the UniqueId via a static counter.
+  The platform must ensure all UniqueIds assigned are unique across the firmware.
+
+  Platforms may need to define their own UniqueIds through this method because
+  the current UniqueId that gets used in NonDiscoverablePciDeviceDxe is a static counter,
+  which can lead to non-deterministic value's being returned by PciIo->GetLocation().
+  In order to have a deterministic UniqueId returned by PciIo->GetLocation(), this protocol
+  can be used to specify a specific UniqueId for a given Handle.
+  For example, if a platform needs to Locate a specific NonDiscoverable PciIo protocol,
+  they could use the deterministic UniqueId assigned through this protocol to find the expected PciIo instance
+  because a platform can match against the GetLocation value returned by using the assigned UniqueId.
+  This protocol allows us to have a 1-1 defined mapping between a platform defined UniqueId and the NonDiscoverable PciIo protocol instance.
+
+  The unique ID must be in the range of:
+    MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
+    MAX_NON_DISCOVERABLE_PCI_DEVICE_ID == 32 * 256
+
+  Usage example:
+    EFI_HANDLE Handle = NULL;
+    UINTN      Size; // NonDiscoverableDevice Size
+    VOID       *Base; // NonDiscoverableDevice Base
+
+    Status = RegisterNonDiscoverableMmioDevice (
+                NonDiscoverableDeviceTypeAhci,
+                NonDiscoverableDeviceDmaTypeCoherent,
+                NULL,
+                &Handle,
+                1,
+                Base,
+                Size
+              );
+
+    NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL * UniqueIdProtocol = (NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL*) AllocateZeroPool (sizeof (NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL));
+    if (UniqueIdProtocol == NULL) {
+      return EFI_OUT_OF_RESOURCES;
+    }
+    UniqueIdProtocol->Revision = NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION;
+    UniqueIdProtocol->UniqueId = 0x1000; // MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
+
+    Status = gBS->InstallMultipleProtocolInterfaces (&Handle, &gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid, UniqueIdProtocol, NULL);
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_H
+#define NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_H
+
+#include <Uefi/UefiBaseType.h>
+
+#define EDKII_NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_GUID \
+  { 0xd60d0e74, 0x5c4e, 0x4d45, { 0x9f, 0xe0, 0x2a, 0x3e, 0x4f, 0x9d, 0x55, 0xc1 } }
+
+#define NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION  1
+
+// We will use the upper half of the ID space for this UniqueId protocol assignment.
+// The lower half can be used as a backup by NonDiscoverablePciDeviceDxe when the protocol is not available.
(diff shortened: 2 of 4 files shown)


MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
```diff
@@ -187,7 +195,42 @@ NonDiscoverablePciDeviceStart (
...
+  if (Status == EFI_SUCCESS) {
+    if (UniqueIdProtocol->Revision != NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: UniqueId protocol revision mismatch. Expected %d, got %d\n",
+        __func__,
+        NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION,
+        UniqueIdProtocol->Revision
+        ));
+      return EFI_INCOMPATIBLE_VERSION;
```
os-d: In these two failure cases, should we assert but still fall back to the static counter? I'm not sure the security implications here.

MdeModulePkg/MdeModulePkg.dec
```diff
@@ -824,6 +824,12 @@
   ## Include/Protocol/NonDiscoverableDevice.h
   gEdkiiNonDiscoverableDeviceProtocolGuid = { 0x0d51905b, 0xb77e, 0x452a, {0xa2, 0xc0, 0xec, 0xa0, 0xcc, 0x8d, 0x51, 0x4a } }
 
+  ## MU_CHANGE [BEGIN]
+  ## Include/Protocol/NonDiscoverableDeviceUniqueId.h
+  # This protocol allows platforms to configure a UniqueId for the GetLocation implemtnation for NonDiscoverablePciDeviceDxe
```
os-d:
```suggestion
  # This protocol allows platforms to configure a UniqueId for the GetLocation implementation for NonDiscoverablePciDeviceDxe
```

os-d (commented):
I'm okay with this approach, but what if we just updated the Register API to include a unique ID? It's a breaking change, but it feels less clunky than this, we don't need the extra logic and error handling. Especially if we expect every consumer to use this. It isn't a spec defined interface, we can change it.

Regardless, I do think more guidance is needed as to what the unique ID is and why a driver would want their own versus the static one.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: My thought process was that if a platform wanted to publish this protocol, then that would mean their intent is to actually assign a deterministic UniqueId VS the original static counter. If so, it may be better to fail explicitly so that it is clear that they either got the Revision wrong or their UnqueId is not in the allowable range.

MdeModulePkg/MdeModulePkg.dec
eeshanl: Fixed

eeshanl:
> I'm okay with this approach, but what if we just updated the Register API to include a unique ID? It's a breaking change, but it feels less clunky than this, we don't need the extra logic and error handling. Especially if we expect every consumer to use this. It isn't a spec defined interface, we can change it.
> 
> Regardless, I do think more guidance is needed as to what the unique ID is and why a driver would want their own versus the static one.

Updated PR description and protocol definition .h file with why this protocol should be used vs the static counter and showed a usage example.

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
+/** @file NonDiscoverableDeviceUniqueId.h
+  With this new protocol, a platform can register a UniqueId to be used to identify the PciIo location.
+  For example, if a device registers with the NonDiscoverableDeviceRegistrationLib,
+  they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice.
+  NonDiscoverablePciDeviceDxe will consume this UniqueId if found and set it to the device's UniqueId.
```
os-d:
```suggestion
  NonDiscoverablePciDeviceDxe will consume this UniqueId if found and the device's UniqueId to this value.
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
+/** @file NonDiscoverableDeviceUniqueId.h
+  With this new protocol, a platform can register a UniqueId to be used to identify the PciIo location.
+  For example, if a device registers with the NonDiscoverableDeviceRegistrationLib,
+  they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice.
+  NonDiscoverablePciDeviceDxe will consume this UniqueId if found and set it to the device's UniqueId.
+  If not found, will back up to the prior method of assigning the UniqueId via a static counter.
```
os-d:
```suggestion
  If not found, it will back up to the prior method of assigning the UniqueId via a static counter.
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
+/** @file NonDiscoverableDeviceUniqueId.h
+  With this new protocol, a platform can register a UniqueId to be used to identify the PciIo location.
+  For example, if a device registers with the NonDiscoverableDeviceRegistrationLib,
+  they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice.
+  NonDiscoverablePciDeviceDxe will consume this UniqueId if found and set it to the device's UniqueId.
+  If not found, will back up to the prior method of assigning the UniqueId via a static counter.
+  The platform must maintain all UniqueId's assigned are unique for every instance of the protocol.
```
os-d:
```suggestion
  The platform must ensure all UniqueIds assigned are unique across the firmware.
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
...
+  they can also publish this new protocol on the same EFI_HANDLE as RegisterNonDiscoverableMmioDevice.
+  NonDiscoverablePciDeviceDxe will consume this UniqueId if found and set it to the device's UniqueId.
+  If not found, will back up to the prior method of assigning the UniqueId via a static counter.
+  The platform must maintain all UniqueId's assigned are unique for every instance of the protocol.
+
+  The platform may need to use this new protocol defined UniqueId instead because
+  the current UniqueId that gets used in NonDiscoverablePciDeviceDxe is a static counter,
+  which can lead to non-deterministic value's being returned by PciIo->GetLocation().
+  In order to have a deterministic UniqueId returned by GetLocation() this protocol
+  can be used to specify a specific UniqueId for a given Handle.
```
os-d:
```suggestion
  Platforms may need to define their own UniqueIds through this method because
  the current UniqueId that gets used in NonDiscoverablePciDeviceDxe is a static counter,
  which can lead to non-deterministic value's being returned by PciIo->GetLocation().
  In order to have a deterministic UniqueId returned by GetLocation() this protocol
  can be used to specify a specific UniqueId for a given Handle.
```

I think adding another sentence here to give an example scenario where you'd want a deterministic UniqueId would be useful.

os-d (approved):

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
...
+  Usage example:
+    EFI_HANDLE Handle = NULL;
+
+    Status = RegisterNonDiscoverableMmioDevice (
+                NonDiscoverableDeviceTypeAhci,
+                NonDiscoverableDeviceDmaTypeCoherent,
+                NULL,
+                &Handle,
+                1,
+                x, y
```
makubacki: I know you're trying to express a variadic function, but I think the `x` and `y` are confusing as-is since most other names refer to actual content. If nothing else, a comment would help.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
```diff
@@ -187,7 +195,42 @@ NonDiscoverablePciDeviceStart (
...
+    if ((UniqueIdProtocol->UniqueId < (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2)) || (UniqueIdProtocol->UniqueId >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID)) {
+      DEBUG ((
+        DEBUG_ERROR,
+        "%a: UniqueId Protocol failed. UniqueId=0x%llx must be greater than or equal to 0x%llx and less than 0x%llx.\n",
+        __func__,
+        UniqueIdProtocol->UniqueId,
+        (MAX_NON_DISCOVERABLE_PCI_DEVICE_ID / 2),
+        MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
+        ));
+      return EFI_INCOMPATIBLE_VERSION;
```
makubacki: I think you still want to close the protocol before returning.

makubacki:
```suggestion
      Status = EFI_INCOMPATIBLE_VERSION;
      goto CloseProtocol;
```

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
makubacki:
```suggestion
      Status = EFI_INCOMPATIBLE_VERSION;
      goto CloseProtocol;
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
...
+#include <Uefi/UefiBaseType.h>
+
+#define EDKII_NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_GUID \
+  { 0xd60d0e74, 0x5c4e, 0x4d45, { 0x9f, 0xe0, 0x2a, 0x3e, 0x4f, 0x9d, 0x55, 0xc1 } }
+
+#define NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION  1
+
+// We will use the upper half of the ID space for this UnqueId protocol assignment.
+// The lower half can be used as a backup by NonDiscoverablePciDeviceDxe when the protocol is not available.
+// The uniqud ID must be in the range of: MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
```
makubacki:
```suggestion
// The unique ID must be in the range of: MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
...
+#define NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_H
+
+#include <Uefi/UefiBaseType.h>
+
+#define EDKII_NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_GUID \
+  { 0xd60d0e74, 0x5c4e, 0x4d45, { 0x9f, 0xe0, 0x2a, 0x3e, 0x4f, 0x9d, 0x55, 0xc1 } }
+
+#define NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION  1
+
+// We will use the upper half of the ID space for this UnqueId protocol assignment.
```
makubacki:
```suggestion
// We will use the upper half of the ID space for this UniqueId protocol assignment.
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
...
+  If not found, will back up to the prior method of assigning the UniqueId via a static counter.
+  The platform must maintain all UniqueId's assigned are unique for every instance of the protocol.
+
+  The platform may need to use this new protocol defined UniqueId instead because
+  the current UniqueId that gets used in NonDiscoverablePciDeviceDxe is a static counter,
+  which can lead to non-deterministic value's being returned by PciIo->GetLocation().
+  In order to have a deterministic UniqueId returned by GetLocation() this protocol
+  can be used to specify a specific UniqueId for a given Handle.
+
+  The uniqud ID must be in the range of:
```
makubacki:
```suggestion
  The unique ID must be in the range of:
```

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
```diff
@@ -0,0 +1,67 @@
...
+
+#define NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL_REVISION  1
+
+// We will use the upper half of the ID space for this UnqueId protocol assignment.
+// The lower half can be used as a backup by NonDiscoverablePciDeviceDxe when the protocol is not available.
+// The uniqud ID must be in the range of: MAX_NON_DISCOVERABLE_PCI_DEVICE_ID/2 <= UniqueId < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID
+#define MAX_NON_DISCOVERABLE_PCI_DEVICE_ID  (32 * 256)
+
+typedef struct _NON_DISCOVERABLE_DEVICE_UNIQUE_ID_PROTOCOL {
+  UINTN    Revision;
```
makubacki: nit: I think this is clearer, but feel free to close.

```suggestion
  UINTN    ProtocolRevision;
```

makubacki (approved):

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: done

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: done

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: done

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: done, added:
For example, if a platform needs to Locate a specific NonDiscoverable PciIo protocol,
they could use the deterministic UniqueId assigned through this protocol to find the expected PciIo instance
because a platform can match against the GetLocation value returned by using the assigned UniqueId.
This protocol allows us to have a 1-1 defined mapping between a platform defined UniqueId and the NonDiscoverable PciIo protocol instance.

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: switched to using directly the example as in mu_tiano_platforms, which shows how to use the Device's Size and Base to be passed into those 2 parameters.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: good catch, added.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: done

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: fixed

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: fixed

MdeModulePkg/Include/Protocol/NonDiscoverableDeviceUniqueId.h
eeshanl: fixed

cfernald (approved):

PR #1549, opened 2025-10-29
Adding check for underflow/overflow when iterating through resource descriptor hobs

kuqin12:
## Description

During DXE core memory service initialization, the system would check available resource descriptor hobs against the memory top from PHIT hob.

However, it is possible that a given resource descriptor hob will not be larger than the cover the memory top, causing the Length calculation to underflow. Or the hob is not properly prepared and cause the calculation to overflow.

This change adds a check for potential underflow/overflow before performing the subtraction.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This change was tested on QEMU SBSA and booted to both UEFI shell and Windows VOS.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -2448,10 +2448,10 @@ CoreInitializeMemoryServices (
     Attributes  = PhitResourceHob->ResourceAttribute;
     BaseAddress = PageAlignAddress (PhitHob->EfiMemoryTop);
     // MU_CHANGE START - Check for potential underflow before subtraction
-    if (PhitHob->EfiMemoryTop > (ResourceHob->PhysicalStart + ResourceHob->ResourceLength)) {
+    if (BaseAddress > (ResourceHob->PhysicalStart + ResourceHob->ResourceLength)) {
       Length = 0;
     } else {
-      Length      = PageAlignLength (ResourceHob->PhysicalStart + ResourceHob->ResourceLength - BaseAddress);
+      Length = PageAlignLength (ResourceHob->PhysicalStart + ResourceHob->ResourceLength - BaseAddress);
       FindLargestFreeRegion (&BaseAddress, &Length, (EFI_HOB_MEMORY_ALLOCATION *)GetFirstHob (EFI_HOB_TYPE_MEMORY_ALLOCATION));
     }
     // MU_CHANGE END - Check for potential underflow before subtraction
--- a/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
+++ b/MdeModulePkg/Core/Dxe/Gcd/Gcd.c
@@ -2462,6 +2462,7 @@ CoreInitializeMemoryServices (
       Length = PageAlignLength (HighAddress - BaseAddress);
       FindLargestFreeRegion (&BaseAddress, &Length, (EFI_HOB_MEMORY_ALLOCATION *)GetFirstHob (EFI_HOB_TYPE_MEMORY_ALLOCATION));
     }
+
     // MU_CHANGE END - Check for potential underflow before subtraction
 
     if (Length < MinimalMemorySizeNeeded) {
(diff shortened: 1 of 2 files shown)


os-d (approved):
Approved, assuming you are also taking to edk2, of course.

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
```diff
@@ -2447,8 +2447,15 @@ CoreInitializeMemoryServices (
     //
     Attributes  = PhitResourceHob->ResourceAttribute;
     BaseAddress = PageAlignAddress (PhitHob->EfiMemoryTop);
-    Length      = PageAlignLength (ResourceHob->PhysicalStart + ResourceHob->ResourceLength - BaseAddress);
-    FindLargestFreeRegion (&BaseAddress, &Length, (EFI_HOB_MEMORY_ALLOCATION *)GetFirstHob (EFI_HOB_TYPE_MEMORY_ALLOCATION));
+    // MU_CHANGE START - Check for potential underflow before subtraction
+    if (BaseAddress > (ResourceHob->PhysicalStart + ResourceHob->ResourceLength)) {
```
apop5: would using SafeUint64Add just be easier? https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Library/SafeIntLib.h#L2155C1-L2155C15

kuqin12: Updated. It does not exactly help the original issue. But it is an extra level of check I think we should have. Thanks.

apop5 (approved):

apop5 (approved):

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
```diff
@@ -2447,8 +2448,22 @@ CoreInitializeMemoryServices (
     //
     Attributes  = PhitResourceHob->ResourceAttribute;
     BaseAddress = PageAlignAddress (PhitHob->EfiMemoryTop);
-    Length      = PageAlignLength (ResourceHob->PhysicalStart + ResourceHob->ResourceLength - BaseAddress);
-    FindLargestFreeRegion (&BaseAddress, &Length, (EFI_HOB_MEMORY_ALLOCATION *)GetFirstHob (EFI_HOB_TYPE_MEMORY_ALLOCATION));
+    // MU_CHANGE START - Check for potential underflow before subtraction
+    Status = SafeUint64Add (ResourceHob->PhysicalStart, ResourceHob->ResourceLength, &HighAddress);
```
makubacki: Line 2427 is essentially performing the same calculation (`ResourceHob->PhysicalStart + ResourceHob->ResourceLength`):

```c
    if (PhitHob->EfiFreeMemoryTop > (ResourceHob->PhysicalStart + ResourceHob->ResourceLength)) {
      continue;
    }
```

(also a similar one later on line 2542)

For consistency, it seems like calculating this once and using it later would be cleaner.

makubacki: I don't have a problem with `SafeIntLib`, but in the past edk2 feedback has encouraged use of more straightforward C code when possible. DId you consider something like this?

```c
if (ResourceHob->ResourceLength > (MAX_UINT64 - ResourceHob->PhysicalStart)) {
  ASSERT_EFI_ERROR (Status);
  HighAddress = MAX_UINT64;
} else {
  HighAddress = ResourceHob->PhysicalStart + ResourceHob->ResourceLength;
}
```

kuqin12: Updated.

apop5 (approved):

MdeModulePkg/Core/Dxe/Gcd/Gcd.c
kuqin12: Thanks for the suggestion. I switched over to this check and updated the rest of the file.

makubacki (approved):

PR #1556, opened 2025-11-04
[Rebase & FF] Standalone MM IPL fixes

kuqin12:
## Description

This change cherry-picks 2 commits from EDK2 that could cause some platforms to not work with Standalone MM as expected.

The last commit is an add-on to resolve the feedback on this PR and will be taken to edk2.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU Q35 platform and booted to Windows desktop.

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
@@ -664,6 +664,10 @@ ExecuteMmCoreFromMmram (
   PE_COFF_LOADER_IMAGE_CONTEXT          ImageContext;
   STANDALONE_MM_FOUNDATION_ENTRY_POINT  Entry;
   EFI_MMRAM_HOB_DESCRIPTOR_BLOCK        *Block;
+  EFI_PEI_MM_ACCESS_PPI                 *MmAccess;
+  UINTN                                 Size;
+  UINTN                                 Index;
+  UINTN                                 MmramRangeCount;
 
   MmFvBase = 0;
   MmFvSize = 0;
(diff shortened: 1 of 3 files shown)


apop5 (approved):

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -680,7 +721,7 @@ ExecuteMmCoreFromMmram (
   //
   Status = PeCoffLoaderGetImageInfo (&ImageContext);
   if (EFI_ERROR (Status)) {
-    return Status;
+    goto Done;
```
makubacki, line 725: The function returns `Status`. At this point, `Status` is some error. Are you expecting it to be overwritten by the status returned from `MmAccess->Close()` and `MmAccess->Lock()` in the `Done` block?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -753,6 +795,44 @@ ExecuteMmCoreFromMmram (
...
+        //
+        // Print error message that the MMRAM failed to lock...
+        //
+        DEBUG ((DEBUG_ERROR, "MM IPL could not lock MMRAM (Index %d) after executing MM Core %r\n", Index, Status));
+        ASSERT (FALSE);
+        return Status;
+      }
+
+      //
+      // Print debug message that the MMRAM window is now closed.
```
makubacki:
```suggestion
      // Print debug message that the MMRAM window is now locked.
```

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -753,6 +795,44 @@ ExecuteMmCoreFromMmram (
...
+  if (MmAccess != NULL) {
+    //
+    // Close all MMRAM ranges, if MmAccess is available.
+    //
+    for (Index = 0; Index < MmramRangeCount; Index++) {
+      Status = MmAccess->Close ((EFI_PEI_SERVICES **)GetPeiServicesTablePointer (), MmAccess, Index);
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "MM IPL failed to close MMRAM windows index %d - %r\n", Index, Status));
+        ASSERT (FALSE);
+        return Status;
```
makubacki: Should this really return instead of attempting to close and lock any potentially remaining windows?

makubacki:
I know this is a cherry-pick but I had a few questions before finishing the review.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -756,6 +795,44 @@ ExecuteMmCoreFromMmram (
...
+      }
+
+      //
+      // Print debug message that the MMRAM window is now closed.
+      //
+      DEBUG ((DEBUG_INFO, "MM IPL closed MMRAM window index %d\n", Index));
+
+      //
+      // Lock the MMRAM (Note: Locking MMRAM may not be supported on all platforms)
```
os-d, line 820: If some platforms don't support locking, will they return an error here? In which case, should we explicitly check for that error code? e.g. if it is EFI_UNSUPPORTED. If it isn't uniform, then we can't always assert and return the status, right?

kuqin12: The idea (at least from Ray) is that those platforms should not produce MM access before IPL runs. In that case, this whole block will be skipped.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: A separate status is added to host returned results.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: Updated.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: This is updated.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: But a check is added.

makubacki (approved):

apop5 (approved):

PR #1570, opened 2025-11-14
BaseCryptOnOneCrypto Support

Flickdm:
## Description

This pull request adds support for forwarding cryptographic library calls to the new OneCrypto protocol across DXE, SMM, and Standalone MM phases, enabling modular and version-aware cryptographic provider integration. It introduces new protocol GUIDs, implements library constructors for protocol lookup, and updates build configuration to support these changes. Additionally, it provides documentation and abstracts for the new libraries, and adds a disabled external binary dependency descriptor.

**OneCrypto Protocol Integration:**

* Introduced new protocol GUIDs for OneCrypto Private, MM, and DXE protocols in `CryptoPkg.dec` to support phase-specific cryptographic operations.
* Implemented DXE, SMM, and Standalone MM library modules (`DxeCryptLib.c`, `SmmCryptLib.c`, `StandaloneMmCryptLib.c`) that locate and expose the OneCrypto protocol for cryptographic services, with corresponding INF build files for each phase. [[1]](diffhunk://#diff-0cce00a8895fcfa11c3c6d94c8848f47289df9bd6274150a07f2c96e09e6b927R1-R68) [[2]](diffhunk://#diff-f8f5c27ebe0e42bfcd08c8c0560cebfe8d92d8f5c8451a706c10c0618a3007bcR1-R46) [[3]](diffhunk://#diff-e42b9ec95aa65234bba458644ff74b428eb6530b31fe38e7ae64eb87f759361dR1-R70) [[4]](diffhunk://#diff-0663d5e952a302e39ec809d00f615e405bc0e55793f2383de3738a28509337f2R1-R43) [[5]](diffhunk://#diff-a9905c4823bda1f1512d3d97a09cca3a5caf6c0ffeac07749f3b2fd1ba0d6d8aR1-R70) [[6]](diffhunk://#diff-b28eb8bf90f88fa9a7288a0165c28e23bdf17d2114ac5ee841937772438060f0R1-R45)

**Documentation and Descriptions:**

```mermaid
---
config:
  layout: elk
---
classDiagram

    namespace Protocol {
        class OneCrypto:::highlight {
            <<Protocol Interface>>
            UINT16 Major;
            UINT16 Minor;
            // v1.0 ------------------------------------------
            ONE_CRYPTO_PROVIDED_FUNC ONE_CRYPTO_PROVIDED_FUNC;
            // ...
        }
    }

    namespace Library {
        class BaseCryptLib:::highlight {
        }

        class TlsLib:::library {

        }

        class HashApiLib:::library {

        }
    }

    namespace LibraryInstance {
        class BaseCryptOnOneCrypto:::highlight  {
            <<Library Interface>>
            EFI_STATUS CryptoFunc(..)
        }
    }

    namespace Driver {
        class OneCryptoBinSupvMm:::driver {
            <<Driver>>
            Phase Agnostic Crypto Driver
            Shared across Phases

            Crypto Entry(..) // phase agnostic code
            SupvMm Entry(..) // stub entry
        }

        class OneCryptoLoaderSupvMm:::driver {
            <<Driver>>
        }

        class OneCryptoLoaderDxe:::driver {
            <<Driver>>
        }
    }

    namespace GenericModules {
        class GenericDriver:::module {
            <<Driver>>
            Any driver that needs crypto
        }

        class GenericLibrary:::module {
            <<Library Interface>>
            Any library that needs crypto
        }

    }

    OneCryptoLoaderSupvMm --> OneCryptoBinSupvMm : Loads Phase Agnostic Code
    OneCryptoLoaderDxe --> OneCryptoBinSupvMm : Loads Phase Agnostic Code

    OneCryptoBinSupvMm ..> OneCryptoLoaderSupvMm : Provides Crypto
    OneCryptoBinSupvMm ..> OneCryptoLoaderDxe : Provides Crypto

    OneCryptoLoaderSupvMm ..|> OneCrypto : Publishes
    OneCryptoLoaderDxe ..|> OneCrypto : Publishes

    BaseCryptOnOneCrypto ..|> OneCrypto : Locates

    BaseCryptOnOneCrypto <.. BaseCryptLib : Backed By
    BaseCryptOnOneCrypto <.. TlsLib : Backed By

    TlsLib ..> BaseCryptLib : Uses
    HashApiLib ..> BaseCryptLib : Uses
    
    GenericLibrary ..> BaseCryptLib : Uses
    GenericDriver ..> BaseCryptLib : Uses
    GenericLibrary ..> TlsLib : Uses
    GenericDriver ..> HashApiLib : Uses

    classDef driver fill:#187286,stroke:#333,stroke-width:2px
    classDef protocol fill:#d63230,stroke:#333,stroke-width:2px
    classDef library fill:#1c77c3,stroke:#333,stroke-width:2px
    classDef module fill:#ff8c00,stroke:#333,stroke-width:2px
    classDef highlight fill:#69995d,stroke:#333,stroke-width:2px
```

Specifically this PR is addressing the following (sage green in the chart):

**Build and Configuration Updates:**

* Removed inclusion of test-specific PCDs from the main DSC configuration to streamline production builds.
* Added a disabled external dependency descriptor for the OneCrypto binary driver in `OneCrypto_ext_dep.json.disabled`.

**BaseCryptLib API Enhancements:**

* Added a new API to retrieve cryptographic provider version information in `BaseCryptLib.h`, improving diagnostics and compatibility checks.

**Code Organization and Comments:**

* Improved comments and documentation in `CryptoPkg.dec` regarding deprecated usage and auto-generated sections for cryptographic provider configuration.

**BaseCryptOnOneCrypto** 

* Library implementation and safe calling usage of the phase agnostic code.

**OneCryptoProtocol**

* Protocol implementation and strict guidance.

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [X] Includes documentation?

## How This Was Tested

QemuQ35

## Integration Instructions

N/A


CryptoPkg/Binaries/OneCrypto_ext_dep.json.disabled
Flickdm, line 1: Placeholder ext dep that uses python -m simple.http server to supply the ext dep to emulate real setup

CryptoPkg/CryptoPkg.dec
Flickdm, line 1: Changes to support OneCrypto. Note we cannot remove the PCDs until we deprecate BaseCryptOnProtocol or make additional changes outside of the scope of this PR.

CryptoPkg/Test/Crypto.pcd.ALL.inc.dsc
Flickdm, line 1: Testing should test all crypto functionality. PCDs are being removed here and end the tests to prevent any PCD from skipping a test that should be ran. Note this will cause regression on BaseCryptOnProtocolPpi (from skip to failure) because those binaries don't carry the additional functionality

CryptoPkg/CryptoPkg.dsc
Flickdm, line 1: Removing the ALL PCDS from testing

CryptoPkg/Include/Library/BaseCryptLib.h
Flickdm, line 1: Reorganize structures that were further down to the defines section

Flickdm: additionally added a way to retrieve a "text" string of the crypto provider version, it will be up to the implementor to populate

CryptoPkg/Include/Protocol/OneCrypto.h
Flickdm, line 1: The OneCrypto Protocol

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.c
Flickdm, line 1: DXE Crypto Library (similar to Standalone) should be VERY simple and just publish

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
Flickdm, line 1: The library calls into OneCryptoLib. This is also where version control is performed and errors populate from depending on for example BaseCryptLib defining a function we will not provide (at least for now) the only example of this today is Sha256ParallelHashing

CryptoPkg/Library/BaseCryptLibOnOneCrypto/StandaloneMmCryptLib.inf
```diff
@@ -0,0 +1,45 @@
...
+
+[Sources]
+  StandaloneMmCryptLib.c
+  OneCryptoLib.c
+
+[Protocols]
+  gOneCryptoMmProtocolGuid  ## CONSUMES
+
+[Depex]
+  gOneCryptoMmProtocolGuid
```
Flickdm: add the line ending

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
```diff
@@ -119,7 +119,8 @@
   gEdkiiVariableRuntimeCacheInfoHobGuid
```
Flickdm, line 123: Bug in VariableSmmRuntimeDxe easily can be its own PR

Flickdm:
Working on fixing the failing gates - uncrustify is failing.

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5446 @@
+/** @file
+  Implements the BaseCryptLib and TlsLib using the services of the EDK II Crypto
+  Protocol/PPI.
```
os-d, line 3: Comment needs update

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5446 @@
+/** @file
+  Implements the BaseCryptLib and TlsLib using the services of the EDK II Crypto
+  Protocol/PPI.
+
+  Copyright (C) Microsoft Corporation. All rights reserved.
+  Copyright (c) 2019 - 2020, Intel Corporation. All rights reserved.<BR>
```
os-d: `Copyright (c) Microsoft Corporation.`

No Intel, this just came from you.

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5446 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+  UPDATED: 2025-Nov-11
+  VERSION: 1.0
+**/
+
+#ifndef ONE_CRYPTO_PROTOCOL_
+#define ONE_CRYPTO_PROTOCOL_
+
+#include <Uefi.h>
```
os-d: Sean didn't want Uefi.h here. I'm not convinced on that, a protocol is phase specific anyway

CryptoPkg/Include/Library/BaseCryptLib.h
```diff
@@ -4357,4 +4358,31 @@ EcDsaVerify (
...
+                               On output, the size of the data copied to the buffer (including null terminator).
+                               If Buffer is NULL, returns the required buffer size.
+
+  @retval  EFI_SUCCESS            The version string was successfully copied to the buffer.
+  @retval  EFI_BUFFER_TOO_SMALL   The buffer is too small. BufferSize contains the required size.
+  @retval  EFI_INVALID_PARAMETER  BufferSize is NULL.
+**/
+EFI_STATUS
+EFIAPI
+GetCryptoProviderVersionText (
```
os-d: You'll also need to add this to existing BaseCryptLibs

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.c
```diff
@@ -0,0 +1,68 @@
+/** @file
+  Implements the GetCryptoServices() API that retuns a pointer to the EDK II
+  Crypto Protocol.
```
os-d: OneCrypto, here and throughout all the files

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.inf
```diff
@@ -0,0 +1,46 @@
+## @file
+# Implements the BaseCryptLib and TlsLib using the services of the EDK II Crypto
+# Protocol.
+#
+# Copyright (c) Microsoft Corporation.
+# Copyright (c) 2020, Hewlett Packard Enterprise Development LP. All rights reserved.<BR>
+# Copyright (c) 2022, Loongson Technology Corporation Limited. All rights reserved.<BR>
```
os-d: Just msft, here and all new files

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.inf
```diff
@@ -0,0 +1,46 @@
...
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = DXE_DRIVER
+  LIBRARY_CLASS                  = BaseCryptLib | DXE_DRIVER UEFI_DRIVER UEFI_APPLICATION DXE_CORE
+  LIBRARY_CLASS                  = TlsLib       | DXE_DRIVER UEFI_DRIVER UEFI_APPLICATION
+  CONSTRUCTOR                    = DxeCryptLibConstructor
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+# VALID_ARCHITECTURES = IA32 X64 ARM AARCH64 RISCV64 LOONGARCH64
```
os-d: edk2 has not wanted valid archs unless there is actually arch specific code. We can drop this comment entirely from all infs

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
```diff
@@ -0,0 +1,6520 @@
...
+
+  @param  Function          Name of the EDK II Crypto Protocol service to call.
+  @param  Args              The argument list to pass to Function.
+  @param  ErrorReturnValue  The value to return if the protocol is NULL or the
+                            service in the protocol is NULL.
+  @param  MinMajor          Minimum required major version for this function.
+  @param  MinMinor          Minimum required minor version for this function.
+
+**/
+#define CALL_CRYPTO_SERVICE(Function, Args, ErrorReturnValue, MinMajor, MinMinor)  \
```
os-d: I'm not opposed to this being in a macro, but I think having it be in a validation function would be easier to read, update, and then you don't need two copies.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
```diff
@@ -0,0 +1,6520 @@
...
+//
+typedef enum {
+  CRYPTO_ERROR_SUCCESS = 0,
+  CRYPTO_ERROR_PROTOCOL_NULL,
+  CRYPTO_ERROR_VERSION_MISMATCH,
+  CRYPTO_ERROR_FUNCTION_NULL,
+  CRYPTO_ERROR_NOT_SUPPORTED
+} CRYPTO_ERROR_CODE;
+
+STATIC CRYPTO_ERROR_CODE  mLastCryptoError = CRYPTO_ERROR_SUCCESS;
```
os-d: I'm not sure how much value this adds. Because this library is built by the platform, it will have asserts, debug prints, etc., if the platform wants it to. In which case you can rely on prints/asserts in the validation function.

CryptoPkg/CryptoPkg.dec
```diff
@@ -47,6 +47,23 @@
   # F46B2EB2-E0D7-4C96-A3B1-CB7C572EB300
   gEdkiiSmmCryptoProtocolGuid =  { 0xF46B2EB2, 0xE0D7, 0x4C96, { 0xA3, 0xB1, 0xc, 0x61, 0xbb, 0x24, 0x5c, 0x42 }}
 
+  ## OneCrypto Private Protocol
+  #
+  # The "Private" protocol is meant for the driver to communicate through the Binary entry point in 
+  # the OneCryptoBin to the Driver in the same phase. It's a simple protocol that exposes the crypto entrypoint.
+  gOneCryptoPrivateProtocolGuid = { 0x854bce61, 0x8d35, 0x4ff5, { 0x9d, 0xb7, 0x30, 0x3a, 0xfb, 0x79, 0x80, 0xe2 }}
+  
+  ## OneCrypto MM Protocol
```
os-d: I'm guessing StMM and SMM share this protocol guid? In which case the commit message is wrong

CryptoPkg/CryptoPkg.dec
```diff
@@ -47,6 +47,23 @@
   # F46B2EB2-E0D7-4C96-A3B1-CB7C572EB300
   gEdkiiSmmCryptoProtocolGuid =  { 0xF46B2EB2, 0xE0D7, 0x4C96, { 0xA3, 0xB1, 0xc, 0x61, 0xbb, 0x24, 0x5c, 0x42 }}
 
+  ## OneCrypto Private Protocol
+  #
+  # The "Private" protocol is meant for the driver to communicate through the Binary entry point in 
+  # the OneCryptoBin to the Driver in the same phase. It's a simple protocol that exposes the crypto entrypoint.
+  gOneCryptoPrivateProtocolGuid = { 0x854bce61, 0x8d35, 0x4ff5, { 0x9d, 0xb7, 0x30, 0x3a, 0xfb, 0x79, 0x80, 0xe2 }}
```
os-d: I'm not clear on what this is. The driver is communicating with the Driver :).

I thought the flow was:

Consumer -> BaseCryptLibOnOneCrypto -> OneCryptoProtocol -> PhaseSpecificLoader -> Direct Call to OneCryptoBin

Where does the private protocol come in?

CryptoPkg/Binaries/OneCrypto_ext_dep.json.disabled
os-d: I would recommend we don't check this in. Leave this for your local testing, but basecore will be updated without the ext dep, then crypto_release gets the new basecore with its updated changes and then makes a release, which we pull into basecore.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
os-d: If this isn't blocking you, I would recommend splitting it out, since this is such a large PR.

CryptoPkg/CryptoPkg.dec
Flickdm: So the SupvMm version is a little bit different than the DXE version. 

The DXE Loader can locate the binary in the FV and then locate the export table.

The SupvMM loader actually uses the fact that the CryptoBinary has a SupvMM entry point.

That entry point will publish a protocol that has one function "Constructor" then the loader can call it without having to do the complex loading behavior that would have required changes to the Supv granting it more permissions than we would like.

All this being said, this Protocol is actually meant for communication between the loader and the crypto binary and no one else. Given tha - this protocol should really be in OneCryptoPkg.dec instead of CryptoPkg.dec that allows for communication of the libraries

CryptoPkg/CryptoPkg.dec
Flickdm: Yeah - Since they dont/cant overlap I don't think its an issue that they share the protocol guid. However I'm not sure what a good name is? do you have thoughts?

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
Flickdm: This sadly will cause OneCryptoPkg to fail due to the way the depex is handeling the loaders. I can split it out but it would mean I would need that to go in first.

CryptoPkg/CryptoPkg.dec
os-d: Thanks for the explanation and agreed on moving this to OneCryptoPkg

CryptoPkg/CryptoPkg.dec
os-d: I think Mm is fine, just pointing out your commit message says there will be three protocol guids, but you have 2.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
os-d: Let’s do that first, it is a small change and easy to review on its own. But I think useful to have split from this so it doesn’t get lost.

CryptoPkg/CryptoPkg.dec
```diff
@@ -47,6 +47,23 @@
...
+  
+  ## OneCrypto MM Protocol
+  #
+  # The actual OneCryptoProtocol that provides the cryptographic interfaces for MM
+  gOneCryptoMmProtocolGuid =  { 0x2C2275C9, 0x3A7B, 0x426F, { 0xBE, 0x54, 0x2D, 0x22, 0xBD, 0x9D, 0x10, 0x93 }}
+
+  ## OneCrypto DXE Protocol
+  #
+  # The actual OneCryptoProtocol that provides the cryptographic interfaces for DXE
+  gOneCryptoDxeProtocolGuid =   { 0x2C2275C9, 0x3A7B, 0x426F, { 0xBE, 0x54, 0x2D, 0x22, 0xBD, 0x9D, 0x10, 0x94 }}
```
makubacki: We have to generate entirely new GUIDs. Please update one of the GUIDs to be newly generated versus manually altering one into the other.

CryptoPkg/Include/Library/BaseCryptLib.h
```diff
@@ -4357,4 +4357,32 @@ EcDsaVerify (
...
+                               If Buffer is NULL, returns the required buffer size.
+
+  @retval  EFI_SUCCESS            The version string was successfully copied to the buffer.
+  @retval  EFI_BUFFER_TOO_SMALL   The buffer is too small. BufferSize contains the required size.
+  @retval  EFI_INVALID_PARAMETER  BufferSize is NULL.
+  @retval  EFI_UNSUPPORTED        The function is not provided by the Crypto provider.
+**/
+EFI_STATUS
+EFIAPI
+GetCryptoProviderVersionText (
```
makubacki: Is there a particular reason this is called `GetCryptoProviderVersionText()` instead of `GetCryptoProviderVersionString()`?

CryptoPkg/CryptoPkg.dec
```diff
@@ -47,6 +47,23 @@
...
 
+  ## OneCrypto Private Protocol
+  #
+  # The "Private" protocol is meant for the driver to communicate through the Binary entry point in 
+  # the OneCryptoBin to the Driver in the same phase. It's a simple protocol that exposes the crypto entrypoint.
+  gOneCryptoPrivateProtocolGuid = { 0x854bce61, 0x8d35, 0x4ff5, { 0x9d, 0xb7, 0x30, 0x3a, 0xfb, 0x79, 0x80, 0xe2 }}
+  
+  ## OneCrypto MM Protocol
+  #
+  # The actual OneCryptoProtocol that provides the cryptographic interfaces for MM
```
makubacki: Are the protocol interfaces expected to be different between `gOneCryptoMmProtocolGuid` and `gOneCryptoDxeProtocolGuid`?

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
```diff
@@ -0,0 +1,6309 @@
...
+  } while (FALSE);
+
+/**
+  Internal worker function that returns the pointer to an EDK II Crypto
+  Protocol/PPI.  The layout of the PPI, DXE Protocol, and SMM Protocol are
+  identical which allows the implementation of the BaseCryptLib functions that
+  call through a Protocol/PPI to be shared for the PEI, DXE, and SMM
+  implementations.
+**/
+VOID *
```
makubacki: Why doesn't this just return `ONE_CRYPTO_PROTOCOL *` instead of `VOID *`?

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
```diff
@@ -0,0 +1,6309 @@
...
+GetCryptoServices (
```
makubacki: Is `GetCryptoServices()` defined in a header file (I don't see it i one).

CryptoPkg/Include/Library/BaseCryptLib.h
Flickdm: Nope just preference, happy to change it.

CryptoPkg/CryptoPkg.dec
Flickdm: No, I assumed I should keep them separate due to 

```
  ## EDK II Crypto DXE protocol
  # 2C2275C9-3A7B-426F-BE54-2D22BD9D1092
  gEdkiiCryptoProtocolGuid =  { 0x2C2275C9, 0x3A7B, 0x426F, { 0xBE, 0x54, 0x2D, 0x22, 0xBD, 0x9D, 0x10, 0x92 }}

  ## EDK II Crypto SMM protocol
  # F46B2EB2-E0D7-4C96-A3B1-CB7C572EB300
  gEdkiiSmmCryptoProtocolGuid =  { 0xF46B2EB2, 0xE0D7, 0x4C96, { 0xA3, 0xB1, 0xc, 0x61, 0xbb, 0x24, 0x5c, 0x42 }}
```

But if there's no reason to I can change this

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
Flickdm: This is mostly holdover from the original implementation, I can update this

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
Flickdm: No, this is from the original implementation and I think its just linker magic. I can change this.

Flickdm: This is how BaseCryptLibOnProtocolPpi also works

Flickdm: Thinking about it more, I think the intent here was to define it in the C file to avoid putting it in a header file since no one else should need to call it. I don't think this belongs in BaseCryptLib.h but I could create a semi private header if we want that additional level of seperation. Before I make that change, what do you think?

CryptoPkg/Include/Library/BaseCryptLib.h
```diff
@@ -90,6 +90,37 @@ typedef enum {
...
+  //
+
+  The two following C-structure are used for parsing CertStack more clearly.
+**/
+#pragma pack(1)
+
+typedef struct {
+  UINT32    CertDataLength;       // The length in bytes of X.509 certificate.
+  UINT8     CertDataBuffer[0];    // The X.509 certificate content (DER).
+} EFI_CERT_DATA;
```
spbrogan: Is this structure UEFI spec defined?  Probably shouldn't be called EFI_ if not.

Flickdm: Not spec defined as far as I can tell, 

This change is just me moving the typedefs up to be with the rest of the type definitions, 

Originally this came in

https://github.com/tianocore/edk2/commit/3702637a528707704fb37defea042f7b10709ec4 8 years ago.

and the only use I can find is

https://github.com/tianocore/edk2/blob/51e2c18a53afe8f69465c331ed36a4e8c2ace5d0/SecurityPkg/Library/AuthVariableLib/AuthService.c#L2016

Flickdm: should I revert this change, and then add a TODO to correct this in EDK2?

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
makubacki: I don't have a strong preference for the scope the header file is placed at - local or package. Though, you should try to keep it as local as possible for the intended consumers. My main ask is to put it in a header file to follow conventional patterns for declaring the prototype and ensuring that shared signature is used consistently across C files.

CryptoPkg/CryptoPkg.dec
makubacki: I asked because if the protocol interfaces are the exact same and expected to remain the exact some per design in the future, I think you can just use a single GUID and that would be less confusing as it indicates that they represent the same protocol interface even though it is managed in different protocol databases depending on phase.

os-d:
A couple notes on the mermaid diagram for clarity:

image

I would label this as the library class/interface that is used. In the current diagram it could be seen that BaseCryptLib is different than BaseCryptLibOnOneCrypto, when it is not, the latter implements the former.

Just want to make sure this is correct:

image

HashApiLib doesn't interact with BaseCryptLib at all?

CryptoPkg/Include/Library/BaseCryptLib.h
```diff
@@ -90,6 +90,36 @@ typedef enum {
   RsaKeyQInv    ///< The CRT coefficient (== 1/q mod p)
 } RSA_KEY_TAG;
 
+/**
```
os-d: Can you remind me why this section moved? If just for visual aesthetics, I would probably leave alone for now.

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5471 @@
+/** @file
+  Implements the BaseCryptLib and TlsLib using the services of the OneCrypto
+  Protocol/PPI.
```
os-d, line 3:
```suggestion
  Implements the backend crypto services of BaseCryptLib and TlsLib using the OneCrypto
  Protocol.
```

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.inf
```diff
@@ -0,0 +1,38 @@
...
+##
+
+[Defines]
+  INF_VERSION                    = 0x0001001B
+  BASE_NAME                      = DxeCryptLib
+  MODULE_UNI_FILE                = CryptLib.uni
+  FILE_GUID                      = BE7B6901-E9FA-4211-A049-B9024A44FAB4
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = DXE_DRIVER
+  LIBRARY_CLASS                  = BaseCryptLib | DXE_DRIVER UEFI_DRIVER UEFI_APPLICATION DXE_CORE
```
os-d, line 17: Does DXE_CORE actually use this? I don't think it can with the depex statement. I would drop it if we don't have a current use case.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.c
```diff
@@ -0,0 +1,69 @@
...
+  Protocol/PPI.  The layout of the PPI, DXE Protocol, and SMM Protocol are
+  identical which allows the implementation of the BaseCryptLib functions that
+  call through a Protocol/PPI to be shared for the PEI, DXE, and SMM
+  implementations.
+
+  This DXE implementation returns the pointer to the EDK II Crypto Protocol
+  that was found in the library constructor DxeCryptLibConstructor().
+**/
+ONE_CRYPTO_PROTOCOL *
+GetCryptoServices (
```
os-d: Instead of this function I think you can just have mCryptoProtocol (though I would rename to be phase agnostic) be extern in the main function. Simplifies my comment in OnCryptoLib.c a little, though this is fine, too.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.c
```diff
@@ -0,0 +1,69 @@
+/** @file
```
os-d, line 1: nit on filenames, a more traditional filename structure would be:

- BaseCryptLibOnOneCrypto
  - BaseCryptLibOnOneCrypto.c
  - BaseCryptLibOnOneCrypto.uni
  - BaseCryptLibOnOneCryptoDxe.c
  - BaseCryptLibOnOneCryptoDxe.inf
  - BaseCryptLibOnOneCryptoSmm.c
  - etc.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
os-d: I meant get rid of the macros entirely and just have the functions call the validate functions. I think the macros are providing more confusion and require having two of them for the void vs non-void. I also would downgrade the DEBUG_INFO print in here to DEBUG_VERBOSE.

So, you'd have this, e.g. (with other suggestion to have extern mCryptoServices):

```c
HmacSha256New (
  VOID
  )
{
  if (mCryptoServices == NULL || mCryptoServices->HmacSha256New == NULL) {
    // print/assert
    return EFI_INVALID_PARAMETER;
  }

  if (!ValidateCryptoVersion (1, 0) {
    // print/assert
    return EFI_UNSUPPORTED;
  }

  return mCryptoServices->HmacSha256New();
}

```

I think this is easier to understand, personally. And tbh, you could get rid of ValidateCryptoVersion and just do the min/major check in the crypto function because it is the one that knows what the right version is. You also get rid of CryptoServiceNotAvailable.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
+
+# BaseCryptLibOnOneCrypto
```
os-d, line 2: nit: edk2 does either README.md or Readme.md. I tend to go with uppercase

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
+
+# BaseCryptLibOnOneCrypto
+
+## Overview
+
+BaseCryptLibOnOneCrypto provides the BaseCryptLib and TlsLib APIs by calling the OneCrypto Protocol/PPI.
```
os-d: In a lot of places you mention the PPI. There is no PPI today. I'm fine to have this be aspirational, but in some places you have the DXE version mentioning the PPI, e.g., so I would remove those cases.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+- **Stable API** (insulates platform code from protocol changes)
+- **Automatic initialization and protocol lookup**
+
+## Usage
+
+### INF Example
+
+```inf
+[LibraryClasses]
+  BaseCryptLib|CryptoPkg/Library/BaseCryptLibOnOneCrypto/BaseCryptLibOnOneCrypto.inf
```
os-d: Funnily, this is what I'm suggesting you move to, but does not currently exist :)

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+### DSC Example
+
+```dsc
+[Components]
+  $(ONE_CRYPTO_PATH)/OneCryptoLoaders/OneCryptoLoaderDxe.inf
+  $(ONE_CRYPTO_PATH)/OneCryptoLoaders/OneCryptoLoaderSupvMm.inf
+  $(ONE_CRYPTO_PATH)/OneCryptoBin/OneCryptoBinSupvMm.inf
+```
+
+### C Code Example
```
os-d: I'm fine to provide a C code example, but I would note that consumers don't change at all to use this instance of BaseCryptLib.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+            UINT16 Major;
+            UINT16 Minor;
+            // v1.0 ------------------------------------------
+            ONE_CRYPTO_PROVIDED_FUNC ONE_CRYPTO_PROVIDED_FUNC;
+            // ...
+        }
+    }
+
+    namespace Library {
+        class BaseCryptLib:::library {
```
os-d, line 54: Similarly to the overall comment, I would name this BaseCryptLib Library Class or something to indicate that it is not BaseCryptLib the instance but the class

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+    classDef protocol fill:#d63230,stroke:#333,stroke-width:2px
+    classDef library fill:#1c77c3,stroke:#333,stroke-width:2px
+    classDef module fill:#ff8c00,stroke:#333,stroke-width:2px
+
+```
+
+**Architecture Components:**
+
+- **Platform Code**: Uses standard BaseCryptLib/TlsLib/HashApiLib APIs
+- **OneCrypto Loaders**: Contain enough logic to find and load the phase agnostic code
```
os-d, line 135:
```suggestion
- **OneCrypto Loaders**: Finds and loads the phase agnostic code
```

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+  }
+
+  // The library handles all protocol interactions internally
+  Sha256Update(Sha256Context, Data, DataSize);
+  Sha256Final(Sha256Context, Hash);
+  Sha256Free(Sha256Context);
+
+  return EFI_SUCCESS;
+}
+```
```
os-d, line 228: This is all duplicate

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+
+## Adding New Cryptographic Functions (Detailed Guide)
+
+This section provides a comprehensive guide for adding new cryptographic functions to the OneCrypto platform.
```
os-d, line 232:
```suggestion
This section provides a comprehensive guide for adding new cryptographic functions to the OneCrypto protocol.
```

os-d, line 232: It seems like this would better belong in mu_crypto_release and this readme points to that (which doesn't exist yet, so this can have a todo to point to it)

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+  ASSERT (FALSE);
+  return FALSE;
+}
+```
+
+**Why This Matters:**
+
+- Allows modules to compile without linking real crypto implementations
+- Enables CI builds that don't require binary dependencies
+- The `ASSERT(FALSE)` catches accidental usage in debug builds
```
os-d, line 357: I think this section is mostly duplicative and can be dropped.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+- **Never reorder or remove existing function pointers**
+- Document which version introduced each function
+
+---
+
+### Step 4: Update BaseCryptLibOnOneCrypto to Call the Protocol
+
+Implement the BaseCryptLib function by calling the protocol using the `CALL_CRYPTO_SERVICE` macro.
+
+**Location:** `CryptoPkg/Library/BaseCryptLibOnOneCrypto/CryptoService.c` (or appropriate file)
```
os-d, line 486: This file doesn't exist. With my suggestion would be in OneCryptoLibOnOneCrypto.c

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+- **Always add to the end** of the structure (maintains binary compatibility)
+- **Increment Minor version** for new functions (Major version changes are breaking changes)
+- **Never reorder or remove existing function pointers**
+- Document which version introduced each function
+
+---
+
+### Step 4: Update BaseCryptLibOnOneCrypto to Call the Protocol
+
+Implement the BaseCryptLib function by calling the protocol using the `CALL_CRYPTO_SERVICE` macro.
```
os-d, line 484: With my suggestion this macro is gone

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+  @return  Pointer to the CRYPTO context that has been initialized.
+           If the allocations fails, GenericCryptoNew() returns NULL.
+**/
+VOID *
+EFIAPI
+CryptoServiceGenericCryptoNew (
+  VOID
+  )
+{
+  return GenericCryptoNew();  // Calls underlying OpenSSL/MbedTLS implementation
```
os-d, line 574:
```suggestion
  return GenericCryptoNew();  // Calls underlying crypto provider implementation
```

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+  )
+{
+  // Input validation
+  if (Context == NULL || OutputData == NULL || OutputSize == NULL) {
+    return FALSE;
+  }
+
+  if ((InputData == NULL) && (InputSize != 0)) {
+
+// In OneCryptoBin.c - Protocol instance initialization
```
os-d, line 606: This is confusing to have here in the middle of a function in a different file.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+  OUT  UINT8        *OutputData,
+  OUT  UINTN        *OutputSize
+  )
+{
+  // Input validation
+  if (Context == NULL || OutputData == NULL || OutputSize == NULL) {
+    return FALSE;
+  }
+
+  if ((InputData == NULL) && (InputSize != 0)) {
```
os-d, line 604: I would finish this function out.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+ CryptoProtocol->CryptoServiceGenericCryptoNew = CryptoServiceGenericCryptoNew;
+ CryptoProtocol->CryptoServiceGenericCryptoProcess = CryptoServiceGenericCryptoProcess;
+```
+
+**Implementation Notes:**
+
+- Prefix implementation functions with `CryptoService` to distinguish from library calls
+- Perform input validation before calling underlying crypto libraries
+- Handle errors gracefully and return appropriate status codes
+- Consider hardware acceleration if available
```
os-d, line 616: This isn't really up to OneCrypto, this is up to the underlying crypto provider. I think you can remove this statement.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+
+---
+
+### Step 6: Update the External Dependency (extdep)
```
os-d, line 620: Step 6 should be build and publish, then update extdep

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+2. Publish the package to your github repository
+3. Update the version number in `ext_dep.json`
+4. update the sha256 of the expected packaged crypto
+5. Test that the build system correctly downloads and extracts the new binary
+
+**Testing the Update:**
+
+```powershell
+# Clean up the old dependencies
+python PlatformBuild.py --clean # Or just delete the ext dep binary folder
```
os-d, line 655: This isn't needed, stuart will handle this.

In general here, I would use the stuart commands directly, not all PlatformBuild.py's have this wrapper built in.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+# Update the dependencies
+python PlatformBuild.py --update
+
+# Verify the correct version is present
+ls $(ONE_CRYPTO_PATH)
+```
+
+---
+
+### Step 7: Add Tests and Update Documentation
```
os-d, line 666: In order for tests to work, you also have to update BaseCryptLib itself so that you can test properly, right? Or these are runtime tests?

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+  GenericCryptoFree(CryptoContext);
+
+  return UNIT_TEST_PASSED;
+}
+
+UNIT_TEST_STATUS
+EFIAPI
+GenericCryptoTestSuiteSetup (
+  VOID
+#### Documentation Updates
```
os-d: nit: newline before new section

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,925 @@
...
+|-------------------|-----------------------|--------|
+| v1.0 functions | v1.0 binary | ✅ Works |
+| v1.0 functions | v1.1 binary | ✅ Works (backward compatible) |
+| v1.1 functions | v1.1 binary | ✅ Works |
+| v1.1 functions | v1.0 binary | ⚠️ New functions return error, old functions work |
+| v2.0 functions | v1.1 binary | ❌ Fails (major version mismatch) |
+
+This ensures smooth upgrades and prevents breaking existing platforms.
+
+## The `CALL_CRYPTO_SERVICE` Macro
```
os-d, line 810: You describe this above, only keep one section and again with my suggestion this goes away.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+- Protocol is trusted (firmware-to-firmware)
+- Input validation is the caller's responsibility
+
+## Support and Contributing
+
+See the main repository's CONTRIBUTING.md for questions or contributions.
+
+## License
+
+BSD-2-Clause-Patent. See LICENSE.txt for details.
```
os-d, line 859: You go into a lot of detail on adding a minor change, but none on how you add a major change. I would also be pedantic about that and when a major change is warranted.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/SmmCryptLib.inf
```diff
@@ -0,0 +1,43 @@
...
+  VERSION_STRING                 = 1.0
+  MODULE_TYPE                    = DXE_SMM_DRIVER
+  LIBRARY_CLASS                  = BaseCryptLib | DXE_SMM_DRIVER
+  LIBRARY_CLASS                  = TlsLib       | DXE_SMM_DRIVER
+  CONSTRUCTOR                    = SmmCryptLibConstructor
+
+#
+# The following information is for reference only and not required by the build tools.
+#
+# VALID_ARCHITECTURES = IA32 X64 ARM AARCH64
```
os-d, line 24: For every INF in this PR, you can get rid of the VALID_ARCHITECTURE comment. That's the request from edk2 currently for things that don't have anything architecture specific, because they are often wrong. For instance, this one is wrong :).

CryptoPkg/CryptoPkg.dec
os-d: You resolved this comment, but left the private protocol here. Should be removed.

CryptoPkg/CryptoPkg.dec
```diff
@@ -47,6 +47,17 @@
   # F46B2EB2-E0D7-4C96-A3B1-CB7C572EB300
   gEdkiiSmmCryptoProtocolGuid =  { 0xF46B2EB2, 0xE0D7, 0x4C96, { 0xA3, 0xB1, 0xc, 0x61, 0xbb, 0x24, 0x5c, 0x42 }}
 
+  ## OneCrypto Private Protocol
+  #
+  # The "Private" protocol is meant for the driver to communicate through the Binary entry point in
+  # the OneCryptoBin to the Driver in the same phase. It's a simple protocol that exposes the crypto entrypoint.
+  gOneCryptoPrivateProtocolGuid = { 0x854bce61, 0x8d35, 0x4ff5, { 0x9d, 0xb7, 0x30, 0x3a, 0xfb, 0x79, 0x80, 0xe2 }}
+
+  ## OneCrypto MM Protocol
```
os-d: You specify MM here and below, should be DXE/MM in both cases.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.c
Flickdm: Yeah originally (and i just tried again) to make this change and its a lot of changes. I wanted to match the original implementation as much as I could to make it easier on an integrator to know what to change

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
Flickdm: Fixing :)

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
Flickdm: I think Sean wanted this here as well, but I'm also happy to make doc changes as needed

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
Flickdm: So I can't respond directly to your comment for some reason..

Regardless - I hear your suggestions about not having the macro, my worry is that for code maintenance having that much repeated code will lead to mistakes. I could do the extern you are suggesting in the macro and I'm happy to investigate alternative approaches but I did try to expand the macro and there are 227 separate places where the code is *nearly* identical and is a rats nest. The macro keeps with the original implementation allowing me to have better confidence that I'm not introducing invalid behavior.

Flickdm:
> A couple notes on the mermaid diagram for clarity:
> 
> image
> I would label this as the library class/interface that is used. In the current diagram it could be seen that BaseCryptLib is different than BaseCryptLibOnOneCrypto, when it is not, the latter implements the former.
> 
> Just want to make sure this is correct:
> 
> image
> HashApiLib doesn't interact with BaseCryptLib at all?

HashApiLib, HmacApiLib, TlsLib, etc all do, this is where I wasn't sure if drawing out all of the place where it calls into crypto necessarily makes it clearer what is happening. I did go ahead and update the examples used though

CryptoPkg/Include/Library/BaseCryptLib.h
Flickdm: Mostly yes, It was meant to collocate all the structures together given they should have been defined at the top. However I can drop this change

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
Flickdm: The assumption is you did that in Step 1

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
os-d: Funnily, I’m more concerned about the macro usage leading to invalid behavior :). E.g. right now there are two macros. Both need to be kept updated, you need to know which to use, etc. The macro also has a fair bit of logic in it, which is almost never good :).

I think the desire to consolidate the validation logic in one place is reasonable, I just think that might be better in a single function instead of two macros. I.e. move the macro logic to the validation function and then call the crypto function directly if the validation passes. Then you don’t need two macros, a function is easier to audit, you don’t have to pass tokens like the function name, etc.

But, the macros don’t need to be a stopping point, if you still want them, you can close this. My biggest concern is that there are two of them. You could keep the macro and just move the call out to the actual function.

CryptoPkg/Include/Private/OneCryptoLib.h
Flickdm, line 1: convert this to an extern and drop the function

CryptoPkg/Library/BaseCryptLibOnOneCrypto/StandaloneMmCryptLib.inf
```diff
@@ -0,0 +1,39 @@
...
+[Defines]
+  INF_VERSION                    = 0x0001001B
+  BASE_NAME                      = StandaloneMmCryptLib
+  MODULE_UNI_FILE                = CryptLib.uni
+  FILE_GUID                      = 436D963C-7BC8-43B3-99CF-37D7A762F77B
+  VERSION_STRING                 = 1.0
+  PI_SPECIFICATION_VERSION       = 0x00010032
+  MODULE_TYPE                    = MM_STANDALONE
+  LIBRARY_CLASS                  = BaseCryptLib | MM_STANDALONE
+  LIBRARY_CLASS                  = TlsLib       | MM_STANDALONE
```
Flickdm, line 19:
```suggestion
  LIBRARY_CLASS                  = TlsLib       | MM_STANDALONE
```

CryptoPkg/CryptoPkg.ci.yaml
Flickdm, line 1: Add document to top level level of CryptoPkg describing what the options are for backing and write out that if a function is added to BaseCryptLib.h that all providers (specifically OneCrypto) needs to be updated to create the WYSIWYG contract

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,850 @@
...
+  "var_name": "BLD_*_ONE_CRYPTO_PATH"
+}
+```
+
+**Process:**
+
+1. Build the OneCrypto binary driver with your new implementation
+2. Publish the package to your github repository
+3. Update the version number in `ext_dep.json`
+4. update the sha256 of the expected packaged crypto
```
apop5:
```suggestion
4. update the sha256 of the expected packaged crypto in `ext_dep.json"

```

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
```diff
@@ -0,0 +1,850 @@
...
+5. Test that the build system correctly downloads and extracts the new binary
```
apop5: mention stuart_update for step 5?

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.inf
```diff
@@ -119,7 +119,8 @@
   gEdkiiVariableRuntimeCacheInfoHobGuid
 
 [Depex]
-  gEfiMmCommunication2ProtocolGuid
+  gEfiMmCommunication2ProtocolGuid AND
```
apop5: was this file included by accident?

CryptoPkg/Library/BaseCryptLibOnOneCrypto/readme.md
Flickdm: Edk2 doesn't use stuart right? How do ext deps work for them?

apop5 (approved):

CryptoPkg/Library/BaseCryptLibOnOneCrypto/DxeCryptLib.c
```diff
@@ -0,0 +1,49 @@
...
+#include <Library/DebugLib.h>
+#include <Library/UefiBootServicesTableLib.h>
+#include <Protocol/OneCrypto.h>
+
+extern ONE_CRYPTO_PROTOCOL  *gCryptoProtocol;
+
+/**
+  Locate the valid Crypto Protocol.
+
+  @param  ImageHandle   The firmware allocated handle for the EFI image.
```
Javagedes: nit: inconsistency in documentation. All of the one crypt header functions have `@param[out]` or `@param[in]` but this does not. I cannot remember which one is the new way and which one is the old way. But should definitely use the new way.

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5475 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+  UPDATED: 2025-Nov-11
+  VERSION: 1.0
+
+  ============================================================================
+  IMPORTANT: PLATFORM CODE SHOULD NOT USE THIS PROTOCOL DIRECTLY
+  ============================================================================
+
+  This protocol is an internal interface designed for communication between
```
Javagedes, line 15: In the readme, You mentioned `protocol structure and function signatures may change`. I would maybe put something akin to that here also

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+- **Platform Code**: Uses standard BaseCryptLib/TlsLib/HashApiLib APIs
+- **OneCrypto Loaders**: Contain enough logic to find and load the phase agnostic code
+- **OneCrypto Binary Driver**: Contains Phase agnostic cryptographic implementations
+- **BaseCryptLibOnOneCrypto Library**: Provides version checking, type safety, error handling, and API translation
+
+## How to Use BaseCryptLibOnOneCrypto
+
+### In Your Module's INF File
+
+Add BaseCryptLib to your LibraryClasses section:
```
Javagedes: This is not true. You do not specify the library implementation in the module's INF file. You specify it in your platform's DSC file. Either in the [LibraryClasses] section or in the library class override of a specific component in the [Components] section.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+
+```inf
+[LibraryClasses]
+  BaseCryptLib|CryptoPkg/Library/BaseCryptLibOnOneCrypto/BaseCryptLibOnOneCrypto.inf
+```
+
+### In Your Platform DSC File
+
+The OneCrypto binary drivers are provided as an external dependency. The build system automatically sets the
```
Javagedes, line 167: Specifying the library class implementation goes here in the DSC

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+  GenericCryptoFree(CryptoContext);
+
+  return UNIT_TEST_PASSED;
+}
+
+UNIT_TEST_STATUS
+EFIAPI
+GenericCryptoTestSuiteSetup (
+  VOID
+#### Documentation Updates
```
Javagedes: This example code is cut off by this new markdown header.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,925 @@
...
+### Breaking Changes
+None (minor version increment maintains backward compatibility)
+
+### Migration Guide
+To use GenericCrypto functions, ensure your OneCrypto binary driver is version 1.1.0 or later.
+```
+
+**3. Platform Integration Guide:**
+
+```markdown
```
Javagedes: I think this is messed up too. You have \```markdown here. Then later a \```c block, but only one ending \``` section. You might need two? It's kind of hard to tell when reviewing in github.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/SmmCryptLib.c
```diff
@@ -0,0 +1,52 @@
...
+#include <Protocol/OneCrypto.h>
+
+extern ONE_CRYPTO_PROTOCOL  *gCryptoProtocol;
+
+/**
+  Constructor looks up the One Crypto Protocol and verifies that it is
+  not NULL and has a high enough version value to support all the BaseCryptLib
+  functions.
+
+  @param  ImageHandle   The firmware allocated handle for the EFI image.
```
Javagedes: same here with inconsistent `@param` documentation.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/StandaloneMmCryptLib.c
```diff
@@ -0,0 +1,52 @@
...
+#include <Protocol/OneCrypto.h>
+
+extern ONE_CRYPTO_PROTOCOL  *gCryptoProtocol;
+
+/**
+  Constructor looks up the One Crypto Protocol and verifies that it is
+  not NULL and has a high enough version value to support all the BaseCryptLib
+  functions.
+
+  @param  ImageHandle   The firmware allocated handle for the EFI image.
```
Javagedes: and here :)

Javagedes (approved):

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5476 @@
...
+  BaseCryptLibOnOneCrypto.
+
+  ============================================================================
+**/
+
+#ifndef ONE_CRYPTO_PROTOCOL_
+#define ONE_CRYPTO_PROTOCOL_
+
+#include <Library/BaseCryptLib.h>
+
```
makubacki, line 44: The GUID should be an extern to support resolution in compilers and IDEs outside the edk2 build autogen system.

```suggestion

extern EFI_GUID gOneCryptoProtocolGuid;

```

CryptoPkg/Library/BaseCryptLibOnOneCrypto/OneCryptoLib.c
```diff
@@ -0,0 +1,6344 @@
...
+  @param  MinMajor          Minimum required major version for this function.
+  @param  MinMinor          Minimum required minor version for this function.
+
+**/
+#define CALL_CRYPTO_SERVICE(Function, Args, ErrorReturnValue, MinMajor, MinMinor)  \
+  do {                                                                 \
+    ONE_CRYPTO_PROTOCOL  *CryptoServices;                            \
+                                                                       \
+    CryptoServices = GetAndValidateCryptoProtocol (#Function, (MinMajor), (MinMinor)); \
+    if ((CryptoServices == NULL) || !ValidateCryptoFunction ((VOID *)CryptoServices->Function, #Function)) { \
```
makubacki: Casting a function pointer to `VOID *` is not portable will be a problem in some compilers, you will need to do something else.

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5476 @@
...
+  @brief Provides comprehensive cryptographic services for UEFI applications.
+
+  @{
+**/
+
+// =============================================================================
+// Protocol version: 1.0
+// =============================================================================
+#define VERSION_MAJOR  1ULL
+#define VERSION_MINOR  0ULL
```
makubacki: Because C essentially has a "global namespace", these macros need to be scoped to one crypto to prevent likely name collisions.

```suggestion
#define ONE_CRYPTO_VERSION_MAJOR  1ULL
#define ONE_CRYPTO_VERSION_MINOR  0ULL
```

CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -0,0 +1,5476 @@
...
+typedef struct _ONE_CRYPTO_PROTOCOL {
+  // ---------------------------------------------------------------------------
+  // Versioning
+  // Major.Minor.Revision
+  // Major - Breaking change to this structure
+  // Minor - Functions added to the end of this structure
+  //
+  // ---------------------------------------------------------------------------
+  UINT16                                            Major;
+  UINT16                                            Minor;
```
makubacki, line 5245: To a casual reader this is currently confusing because "Revision" is not mentioned again after this comment. Please provide further explanation of how revision is tracked/used.

CryptoPkg/Library/BaseCryptLibOnOneCrypto/Readme.md
```diff
@@ -0,0 +1,891 @@
...
+
+- **Version validation** (at runtime, using protocol major/minor)
+- **Type safety** (strong C API)
+- **Consistent error handling**
+- **Stable API** (insulates platform code from protocol changes)
+- **Automatic initialization and protocol lookup**
+
+## Usage
+
+### INF Example
```
makubacki: This section needs updated. For example, `BaseCryptLib|CryptoPkg/Library/BaseCryptLibOnOneCrypto/BaseCryptLibOnOneCrypto.inf` no longer exists.

makubacki (approved):

apop5 (approved):

PR #1583, opened 2025-12-01
[CHERRY-PICK] Making StandaloneMmIplPei compatible with supervisor

kuqin12:
## Description

Current implementation of the `StandaloneMmIplPei` has a few design differences compared to MM supervisor.

This will introduce more maintenance burden as we move to the next phase of MM supervisor.

This change is made to try to converge the 2 designs, at least from the perspective of IPL, in order to use the IPL as is and still conform to the MM supervisor modelity.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on QEMU Q35 and booted to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
+++ b/StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
@@ -1026,9 +1026,12 @@ CreateMmFoundationHobList (
   //
   // Build ACPI variable HOB
   //
-  HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
-  MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gEfiAcpiVariableGuid, FALSE);
-  UsedSize += HobLength;
+  if (PcdGetBool (PcdAcpiS3Enable)) {
+    // Only check on this variable when S3 needs it.
+    HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
+    MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gEfiAcpiVariableGuid, FALSE);
+    UsedSize += HobLength;
+  }
 
   if (FeaturePcdGet (PcdCpuSmmProfileEnable)) {
     //


StandaloneMmPkg/Drivers/MmCommunicationDxe/MmCommunicationDxe.c
```diff
@@ -454,19 +454,28 @@ MmCommunicationEntryPoint (
...
-  VOID               *Registration;
+  EFI_STATUS                     Status;
+  EFI_HANDLE                     Handle;
+  EFI_HOB_GUID_TYPE              *GuidHob;
+  MM_COMM_BUFFER                 *MmCommonBuffer;
+  UINTN                          Index;
+  VOID                           *Registration;
+  MM_COMM_BUFFER_UPDATE_PROTOCOL *MmCommBufferUpdate;
+
+  Status = gBS->LocateProtocol (&gMmCommBufferUpdateProtocolGuid, NULL, (VOID**)&MmCommBufferUpdate);
```
kuqin12: This is for comprehensiveness if/when we upstream this to edk2. The system should still work even if without this change.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
```diff
@@ -1023,12 +1023,13 @@ CreateMmFoundationHobList (
...
-  // Build ACPI variable HOB
-  //
-  HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
-  MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gEfiAcpiVariableGuid, FALSE);
-  UsedSize += HobLength;
+  // MU_CHANGE: Do not support S3
+  // //
+  // // Build ACPI variable HOB
+  // //
+  // HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
```
kuqin12: This will be guarded by the s3 PCD.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
```diff
@@ -1043,7 +1044,8 @@ CreateMmFoundationHobList (
   // Build resource HOB for unblocked region
   //
   HobLength = GetRemainingHobSize (*FoundationHobSize, UsedSize);
-  MmIplBuildResourceHobForUnblockedRegion (FoundationHobList + UsedSize, &HobLength);
+  // MmIplBuildResourceHobForUnblockedRegion (FoundationHobList + UsedSize, &HobLength);
+  MmIplCopyGuidHob (FoundationHobList + UsedSize, &HobLength, &gMmUnblockRegionHobGuid, TRUE);
```
kuqin12: We will rename the GUID...

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -914,7 +914,7 @@ MmIplDispatchMmDrivers (
   //
   Size   = sizeof (CommunicateHeader);
   Status = Communicate (NULL, &CommunicateHeader, &Size);
-  ASSERT_EFI_ERROR (Status);
+  // ASSERT_EFI_ERROR (Status);
```
kuqin12: Not sure what to do with this.

StandaloneMmPkg/Drivers/MmCommunicationDxe/MmCommunicationDxe.c
kuqin12: This is dropped.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
kuqin12: A different PR is created here: https://github.com/microsoft/mu_feature_mm_supv/pull/583

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: This is PRed as is. Will see if others have different opinions.

makubacki: For clarity, can you provide a simple comparison of design for this point?

StandaloneMmPkg/Drivers/StandaloneMmIplPei/MmFoundationHob.c
kuqin12: Guarded.

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
kuqin12: The issue behind this is that the supervised Standalone MM hosts this handler in supervisor channel. Thus to invoke this handler, we need to indicate the given MM communicate call is targeting supervisor handler pool (we have that wrapped in separate PPI and protocol).

The call here will issue an MMI with user mode data populated and will end up with EFI_NOT_FOUND error because there is not handler for driver dispatching in the user space.

The removal of the asserts here is basically telling this IPL to ignore the error, and it will be handled later through a supv specific module.

makubacki: I haven't followed the full code path, but would one of these options work?

- Make separate INF with the function that starts the call stack different for a "communicate" and "non-communicate" version of StandaloneMmIplPei.
- Setup a unique/appropriate status code return type that can bubble back up in this case and only assert if an error code and not that return type.

kuqin12: oh, i see :) then we will bury that in the ring3 broker layer. That is fine with me. Let me try that in a bit. Thanks~!

kuqin12: I ended up doing the second option, given the ring 3 broker is not dispatched yet. And the change will likely look like this: https://github.com/kuqin12/mu_feature_mm_supv/commit/348e63e0ebb10e7966184b5dc7b5d4e0b4d93cda

makubacki: I think that's a little more preferable anyway to reduce platform integration overhead by avoiding two IPL INFs to pick from.

kuqin12: Okay, this change is updated.

kuqin12:
not stale...

StandaloneMmPkg/Drivers/StandaloneMmIplPei/StandaloneMmIplPei.c
```diff
@@ -1028,7 +1029,12 @@ StandaloneMmIplPeiEntry (
   // Dispatch StandaloneMm drivers in MM
   //
   Status = MmIplDispatchMmDrivers ();
-  ASSERT_EFI_ERROR (Status);
+  // MU_CHANGE: Do not assert on dispatch request returned with not_ready as we may need to do it through other channels.
+  if (Status == EFI_NOT_READY) {
+    DEBUG ((DEBUG_WARN, "%a Failed with not ready, may need further intention\n", __func__));
+  } else {
+    ASSERT_EFI_ERROR (Status);
+  }
```
kuqin12: Change of plan on this one. The current supervisor will not be made to work with this IPL. Instead, we will make the next gen support this naturally.

kuqin12: The other change for hob creation needs to get through.

kuqin12: This was updated. Will make changes on the supervisor side shortly.

kuqin12:
> This pull request has been automatically been closed because it did not have any activity in 60 days and no follow up within 7 days after being marked stale. Thank you for your contributions.

Closing is right. This was superseded by the cherry-pick PR https://github.com/microsoft/mu_basecore/pull/1795

PR #1586, opened 2025-12-03
CryptoPkg: Standardize BaseCryptLib

Flickdm:
## Description

This PR does two things:

1. Add definition for GetCryptoProviderVersionString.
   - While BaseCryptLibOnProtocolPpi cannot support this (today) the intent is that this will be supported by
      OneCryptoPkg and BaseCryptLibOnProtocolPpi is being deprecated.
2. Remove ParallelHash256
  - Again BaseCryptLibOnProtocolPpi can technically support this, however the goal is remove it becuase
     its dependencies make the current implementation not possible for multiple phases and no dependecies
     could be found for it.

- [X] Impacts functionality?
   Possible - A search was performed but no ParallelHash256 dependencies were found and if found, platforms should move away from.
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

QemuQ35

## Integration Instructions

Use the traditional hashing functions


os-d (approved):
If you want to do your BaseCryptLib.h moving around of things, I would do that in this PR. i.e. you were moving a definition up or something.

makubacki (approved):
Please provide a bit more context in the `GetCryptoProviderVersionString()` interface about the content of the version string being returned. Ranging from "`<crypto provider name> <ver>`" to just "`<ver>`" to "completely opaque depending on crypto provider with no guarantee", etc. Just to make caller expectations clear in the interface.

CryptoPkg/Include/Library/BaseCryptLib.h
Flickdm, line 1: Add comment saying that BaseCryptLib.h can only be changed when there is consideration for OneCrypto

Flickdm: making this in the PR for BaseCryptLibOnOneCrypto

apop5 (approved):

PR #1588, opened 2025-12-09
MdeModulePkg/Variable: Add PCD to control RT cache allocation

makubacki:
## Description

Add `PcdMigrateVariableRuntimeCacheBufferAllocation` to allow platforms to choose between allocating variable runtime cache buffers directly in PEI as `EfiRuntimeServicesData` (original behavior) or allocating as `EfiBootServicesData` in PEI and migrating to `EfiRuntimeServicesData` in DXE (to prevent runtime memory fragmentation).

The PCD defaults to FALSE to preserve the original behavior.

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- Boot to OS on a system without the Mu MM Supervisor (PCD = TRUE)
- Boot to OS on a system with the Mu MM Supervisor (PCD = FALSE)
- Tested on client and server platforms that do not use the Mu MM Supervisor (PCD = FALSE)

## Integration Instructions

- If the platform uses the [Mu MM Supervisor](https://github.com/microsoft/mu_feature_mm_supv), then it is capable of unblocking buffers for MM access in DXE and should set the PCD (`PcdMigrateVariableRuntimeCacheBufferAllocation`) to `TRUE` to avoid memory bucket fragmentation which may affect hibernate stability.
- If not, it likely cannot unblock buffers for MM access after PEI and should retain original behavior (`PcdMigrateVariableRuntimeCacheBufferAllocation` set to `FALSE`), so no platform change is needed.

- Ensure the proper `MmUnblockMemoryLib` is linked for your platform usage.
  - If using the Mu MM Supervisor:
    - `VariableSmmRuntimeDxe` should use [MmSupervisorUnblockMemoryLibDxe](https://github.com/microsoft/mu_feature_mm_supv/blob/main/MmSupervisorPkg/Library/MmSupervisorUnblockMemoryLib) from MmSupervisorPkg
  - If using the edk2 Standalone MM code code:
    - `VariablePei` should use [UefiCpuPkg/Library/MmUnblockMemoryLib](https://github.com/tianocore/edk2/blob/master/UefiCpuPkg/Library/MmUnblockMemoryLib)
    - `VariableSmmRuntimeDxe` should use [MdePkg/Library/MmUnblockMemoryLib](https://github.com/tianocore/edk2/tree/master/MdePkg/Library/MmUnblockMemoryLib)

This is marked as a "breaking change" because it reverts the default behavior in Mu back to the edk2 behavior.

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
@@ -1613,168 +1613,172 @@ InitVariableCache (
       );
 
     CopyMem (&mVariableRtCacheInfo, VariableRuntimeCacheInfo, sizeof (VARIABLE_RUNTIME_CACHE_INFO));
-    //
-    // MU_CHANGE - Commented out old header initialization (moved to individual buffer relocations)
-    // InitVariableStoreHeader ((VOID *)(UINTN)mVariableRtCacheInfo.RuntimeHobCacheBuffer, AllocatedHobCacheSize);
-    // InitVariableStoreHeader ((VOID *)(UINTN)mVariableRtCacheInfo.RuntimeNvCacheBuffer, AllocatedNvCacheSize);
-    // InitVariableStoreHeader ((VOID *)(UINTN)mVariableRtCacheInfo.RuntimeVolatileCacheBuffer, AllocatedVolatileCacheSize);
-    //
 
-    //
-    // MU_CHANGE [BEGIN] - Relocate runtime cache buffers from boot services to runtime services
-    // This prevents runtime memory fragmentation by consolidating allocations in DXE
-    //
-    // Relocate CacheInfoFlag buffer from boot services to runtime services
-    //
-    if (mVariableRtCacheInfo.CacheInfoFlagBuffer != 0) {
-      Pages  = EFI_SIZE_TO_PAGES (sizeof (CACHE_INFO_FLAG));
-      Status = gBS->AllocatePages (
-                      AllocateAnyPages,
-                      EfiRuntimeServicesData,
-                      Pages,
-                      &TempCacheInfoBuffer
-                      );
-      if (EFI_ERROR (Status)) {
-        DEBUG ((DEBUG_WARN, "Failed to allocate runtime CacheInfoFlag buffer: %r\n", Status));
-        AllRtBufferAllocationsSucceeded = FALSE;
-      } else {
-        Status = MmUnblockMemoryRequest (TempCacheInfoBuffer, Pages);
-        if ((Status != EFI_UNSUPPORTED) && EFI_ERROR (Status)) {
-          DEBUG ((DEBUG_WARN, "Failed to unblock CacheInfoFlag buffer: %r\n", Status));
-          gBS->FreePages (TempCacheInfoBuffer, Pages);
-          TempCacheInfoBuffer             = 0;
+    // MU_CHANGE [BEGIN] - Allocate RT cache buffer in DXE instead of PEI to prevent fragmentation
+    if (PcdGetBool (PcdMigrateVariableRuntimeCacheBufferAllocation)) {
+      //
+      // Relocate runtime cache buffers from boot services to runtime services
+      // This prevents runtime memory fragmentation by consolidating allocations in DXE
+      //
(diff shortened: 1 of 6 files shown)


kuqin12:
I am sure this change works fine for this case. But we discussed this type of issue in general offline. Maybe we can try to use the method from this [change](https://github.com/tianocore/edk2/pull/5288) to implement PEI memory bin, which will resolve the fragmented memory issue for all.

makubacki:
> I am sure this change works fine for this case. But we discussed this type of issue in general offline. Maybe we can try to use the method from this [change](https://github.com/tianocore/edk2/pull/5288) to implement PEI memory bin, which will resolve the fragmented memory issue for all.

I think that should be a follow up change. It needs further review and testing and might have unanticipated repercussions both in general and with the Rust DXE Core. This change is meant to simply solve the one place this is an issue in basecoree code (the variable driver) more widespread impact to the way buckets are managed.

---

Is the main missing piece for that solution PEI logic to carve out the bucket region (with `gEfiMemoryTypeInformationGuid` as the memory owner) and allocation logic to route generic runtime requests to the bucket region? I'm not aware of that being available in PEI today.

kuqin12:
> Is the main missing piece for that solution PEI logic to carve out the bucket region (with `gEfiMemoryTypeInformationGuid` as the memory owner) and allocation logic to route generic runtime requests to the bucket region? I'm not aware of that being available in PEI today.

Yes, that is one of the topics we discussed with Kinney the other day. Although he prefers to stay with the method you did in this PR, allocating all special memory at the top of memory (once the variable is read) and reroute all subsequent allocations to those buckets should basically cover the bucketing in PEI.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
```diff
@@ -1613,168 +1613,172 @@ InitVariableCache (
...
+                        &TempCacheInfoBuffer
+                        );
+        if (EFI_ERROR (Status)) {
+          DEBUG ((DEBUG_WARN, "Failed to allocate runtime CacheInfoFlag buffer: %r\n", Status));
           AllRtBufferAllocationsSucceeded = FALSE;
+        } else {
+          Status = MmUnblockMemoryRequest (TempCacheInfoBuffer, Pages);
+          if ((Status != EFI_UNSUPPORTED) && EFI_ERROR (Status)) {
+            DEBUG ((DEBUG_WARN, "Failed to unblock CacheInfoFlag buffer: %r\n", Status));
+            gBS->FreePages (TempCacheInfoBuffer, Pages);
```
kuqin12, line 1640: Just so that I understand, after migration, the originally allocated region is still mapped from MM environment, correct?

makubacki:
> > Is the main missing piece for that solution PEI logic to carve out the bucket region (with `gEfiMemoryTypeInformationGuid` as the memory owner) and allocation logic to route generic runtime requests to the bucket region? I'm not aware of that being available in PEI today.
> 
> Yes, that is one of the topics we discussed with Kinney the other day. Although he prefers to stay with the method you did in this PR, allocating all special memory at the top of memory (once the variable is read) and reroute all subsequent allocations to those buckets should basically cover the bucketing in PEI.

Yeah, I don't think bucketing in PEI would be complicated and the details are abstracted/simplified for DXE through the single resource descriptor HOB allowed. It's a good direction overall. Given RT allocations usually need to occur in post-memory, a stable region of memory should be available to set the buckets up and maintain them from that point onward. Again, I'm just hesitant to move systems that immediately need a solution onto that right away given there is still some work in a few places needed.

I filed the Rust DXE Core support tracking issue here https://github.com/OpenDevicePartnership/patina/issues/1175.

MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
makubacki: Yes

kuqin12 (approved):

makubacki:
@apop5, do you have a concern with this change?

apop5 (approved):

apop5:
> @apop5, do you have a concern with this change?

No concerns. I forgot to approve earlier.

PR #1593, opened 2025-12-11
[CHERRY-PICK] StandaloneMmPkg: Split MmEvent to a separate Driver

K-Pavana:
## Description

Needed to integrate Intel ref code without the change due to a build issue.

---

Due to PEIM will do following MM notify event under API mode:

1. MM end of dxe notify Event
2. MM ready to lock notify Event
3. MM ready to boot notify Event
4. MM exit boot services notify Event

It will conflict with the notify event in MmCommunicationDxe.inf on edk2 bootloader under API mode, so split following MmEvent to MmCommunicationNotifyDxe.inf, and avoid run this driver under API mode.

Cc: Ray Ni <ray.ni@intel.com>
Cc: Star Zeng <star.zeng@intel.com>
Cc: Jiaxin Wu <jiaxin.wu@intel.com>
Cc: Dun Tan <dun.tan@intel.com>
Cc: Khor Swee Aun <swee.aun.khor@intel.com>
Cc: Sami Mujawar <sami.mujawar@arm.com>
Cc: Ard Biesheuvel <ardb+tianocore@kernel.org>
(cherry picked from commit e44cb970da6da149dc796a4782be570e101a7fed)

- [x] Impacts functionality?
- [ ] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Code that uses this change on a server platform

## Integration Instructions

Todo

### DIFF
--- a/StandaloneMmPkg/Drivers/MmCommunicationNotifyDxe/MmCommunicationNotifyDxe.c
+++ b/StandaloneMmPkg/Drivers/MmCommunicationNotifyDxe/MmCommunicationNotifyDxe.c
@@ -1,0 +1,265 @@
+/** @file
+  MmCommunicationNotifyDxe driver creates the notifications of
+  some protocols and event.
+
+  Copyright (c) 2025, Intel Corporation. All rights reserved.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#include "MmCommunicationNotifyDxe.h"
+
+//
+// PI 1.7 MM Communication Protocol 2 instance
+//
+EFI_MM_COMMUNICATION2_PROTOCOL  *mMmCommunication2;
+MM_COMM_BUFFER                  mMmCommonBuffer;
+EFI_SMM_ACCESS2_PROTOCOL        *mSmmAccess;
+BOOLEAN                         mSmmLocked = FALSE;
+BOOLEAN                         mEndOfDxe  = FALSE;
+
+//
+// Table of Protocol notification and GUIDed Event notifications that the Standalone Mm requires
+//
+MM_EVENT_NOTIFICATION  mMmEvents[] = {
+  //
+  // Declare protocol notification on DxeMmReadyToLock protocols.  When this notification is established,
+  // the associated event is immediately signalled, so the notification function will be executed and the
+  // DXE Mm Ready To Lock Protocol will be found if it is already in the handle database.
+  //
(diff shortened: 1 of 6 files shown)


makubacki (changes requested):
@K-Pavana. I've updated the integration instructions in the PR description for "N/A" to "Todo". There are definitely integration instructions and this is a breaking change.

Please update the integration instructions to explain:

1. What platforms are impacted by this change (e.g. those that use `MmCommunicationDxe` in StandaloneMmPkg)
2. What is expected of those platforms (e.g. add the new module and keep the original)
3. What is expected by platforms that use Mu MM Supervisor communication

kuqin12:
@makubacki do you understand the commit message? Where in PEIM are they doing these events?

makubacki:
> @makubacki do you understand the commit message? Where in PEIM are they doing these events?

Given it is talking about "API mode", I assume what it means is that:

"FSP has API mode and dispatch mode. Because dispatch mode is PI-compatible, it can register real DXE handlers. Because API mode is intended to work with non-PI bootloaders, all of its event handlers must be registered in PEI (a PEIM) because FSP hosts its own PEI core in API mode. However, nothing stops API mode from being used by a PI-compatible (edk2) wrapper that has DXE. In that case, we now have a DXE driver, `MmCommunicationDxe` that duplicates functionality with the FSP API mode PEIM. Let's avoid that by splitting the conflicting code to a separate driver that is not included in API mode."

StandaloneMmPkg/Drivers/MmCommunicationNotifyDxe/MmCommunicationNotifyDxe.c
```diff
@@ -0,0 +1,265 @@
...
+      EFI_ERROR_CODE | EFI_ERROR_UNRECOVERED,
+      (EFI_SOFTWARE_SMM_DRIVER | EFI_SW_EC_ILLEGAL_SOFTWARE_STATE)
+      );
+    ASSERT (FALSE);
+  }
+
+  //
+  // Lock the SMRAM (Note: Locking SMRAM may not be supported on all platforms)
+  //
+  mSmmAccess->Lock (mSmmAccess);
```
kuqin12, line 150: I know this is a cherry-pick, but where is the close call?

apop5:
Spoke with @K-Pavana 

We are okay to close this with 202511 being available

PR #1595, opened 2025-12-11
NetworkPkg: Handle UDP4 polling failures

makubacki:
## Description

`EfiDhcp4Start()` would spin forever if `DhcpSb->IoStatus` is not updated from `EFI_ALREADY_STARTED`.

This change captures the poll status (instead of ignoring it) and break out on failures so scenarios like unplug/media-loss return an error instead of hanging.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- PXE boot with USB NIC disconnected during boot

## Integration Instructions

- N/A

### DIFF
--- a/NetworkPkg/Dhcp4Dxe/Dhcp4Impl.c
+++ b/NetworkPkg/Dhcp4Dxe/Dhcp4Impl.c
@@ -824,7 +824,15 @@ EfiDhcp4Start (
 
   if (CompletionEvent == NULL) {
     while (DhcpSb->IoStatus == EFI_ALREADY_STARTED) {
-      DhcpSb->UdpIo->Protocol.Udp4->Poll (DhcpSb->UdpIo->Protocol.Udp4);
+      // MU_CHANGE [BEGIN] - Add Udp4 Polling break support
+      Status = DhcpSb->UdpIo->Protocol.Udp4->Poll (DhcpSb->UdpIo->Protocol.Udp4);
+      if (EFI_ERROR (Status) && (Status != EFI_TIMEOUT)) {
+        // Break out if the NIC goes away or poll fails so we don't spin forever.
+        DhcpSb->IoStatus = Status;
+        break;
+      }
+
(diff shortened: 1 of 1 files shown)


NetworkPkg/Dhcp4Dxe/Dhcp4Impl.c
```diff
@@ -824,7 +824,15 @@ EfiDhcp4Start (
 
   if (CompletionEvent == NULL) {
     while (DhcpSb->IoStatus == EFI_ALREADY_STARTED) {
-      DhcpSb->UdpIo->Protocol.Udp4->Poll (DhcpSb->UdpIo->Protocol.Udp4);
+      // MU_CHANGE [BEGIN] - Add Udp4 Polling break support
+      Status = DhcpSb->UdpIo->Protocol.Udp4->Poll (DhcpSb->UdpIo->Protocol.Udp4);
+      if (EFI_ERROR (Status)) {
```
cfernald: Is the intention to catch things like `EFI_DEIVCE_ERROR`? Should this have an exception for `EFI_TIMEOUT`?

makubacki: Yeah, the goal is to use the `Poll()` return code to prevent the system from hanging in the loop if polling encounters an error. I'll check if an exception for `EFI_TIMEOUT` should be added.

kuqin12 (approved):

NetworkPkg/Dhcp4Dxe/Dhcp4Impl.c
makubacki: @cfernald, I forgot that this PR is still open. It looks like an exclusion for `EFI_TIMEOUT` should be added. That's added in the latest push.

cfernald (approved):

PR #1602, opened 2025-12-17
[REBASE&FF] ReportStatusCode: Always cache protocol before usage

### LINKED ISSUE #1601: [Bug]: Potential TPL inversion in ReportStatusCodeLib (DXE and Runtime)

- [x] I have searched existing issues

### Current Behavior

When the system asserts, it is possible that some platform will invoke report status code, to log telemetry, to print a message, etc. However, the assert could have happened at any TPL.

With the current implementation inside basecore, if this is the first time an RSC invoked from inside a module, it will try to locate the protocol before dispatching to the listeners. If the TPL is already at HIGH, the locate call could have failed or double faulted.

### Expected Behavior

Rather than trying to locate on the fly, the library should have attempted to locate the protocol inside the constructor, and register a callback if this is not there. Then the global protocol will be populated whenever the protocol is ready.

If an RSC call is invoked before protocol var is populated, we will have to drop it, just like it failed to locate the protocol.

### Steps To Reproduce

Invoke secure boot variable write and delete multiple times, it will trigger the event log to be full and cause an assert, which will then fail when trying to log a telemetry entry.

### Build Environment

```markdown
- OS(s):Win11
- Tool Chain(s): VS2022
- Targets Impacted: X64 & Aarch64
```

### Version Information

```text
Top of release.
```

### Urgency

High

Someone else needs to fix it

No maintainer feedback needed

### Anything else?

This is needed on some physical platform shortly.

Javagedes:
## Description

This commit stops the library for querying the protocol database for the
status code protocol at the time of use of the library function. This
logic can result in a TPL inversion if the protocol has not yet been
located and the caller is running in a TPL higher then TPL_NOTIFY.

Instead, the protocol is now located during construction of the library
during driver startup. If the protocol is not found, then an on-protocol
install event is registered. A deconstructor is also added to close the
event in the case that the driver is unloaded before the protocol is
installed.

This PR is a rebase and fast forward as it includes a new test.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

1. The written unit test, which failed in the first commit, now passes after the second commit
2. Verified that the event callback is successfully executed in all drivers that run before the protocol is produced, including the DXE Core.
3. Verified the steps to reproduce specified in #1601 no longer reproduces.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
+++ b/MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
@@ -20,6 +20,7 @@
 #include <Guid/StatusCodeDataTypeDebug.h>
 
 EFI_STATUS_CODE_PROTOCOL  *mReportStatusCodeLibStatusCodeProtocol = NULL;
+EFI_EVENT                 mOnStatusCodeProtocolInstallEvent; // MU_CHANGE - Add constructor to cache protocol or register for protocol notification
 
 /**
   Locate the report status code service.
@@ -86,9 +87,8 @@ InternalReportStatusCode (
       (ReportDebugCodeEnabled () && (((Type) & EFI_STATUS_CODE_TYPE_MASK) == EFI_DEBUG_CODE)))
   {
     //
-    // If mReportStatusCodeLibStatusCodeProtocol is NULL, then check if Report Status Code Protocol is available in system.
+    // If mReportStatusCodeLibStatusCodeProtocol is NULL, the protocol has not yet been registered. Return immediately.
     //
-    InternalGetReportStatusCode ();
     if (mReportStatusCodeLibStatusCodeProtocol == NULL) {
       return EFI_UNSUPPORTED;
     }
@@ -611,3 +611,109 @@ ReportDebugCodeEnabled (
 {
   return (BOOLEAN)((PcdGet8 (PcdReportStatusCodePropertyMask) & REPORT_STATUS_CODE_PROPERTY_DEBUG_CODE_ENABLED) != 0);
 }
+
+// MU_CHANGE [BEGIN] - Add constructor to cache protocol or register for protocol notification
+
+/**
+  Notification function called when the Status Code Runtime protocol is installed. Locates and caches the protocol.
+
+    @param    Event           Not Used.
+    @param    Context         Not Used.
+
+   @retval   none
+ **/
+VOID
+EFIAPI
+OnStatusCodeRuntimeProtocolInstalledNotification (
+  IN EFI_EVENT  Event,
+  IN VOID       *Context
+  )
+{
+  InternalGetReportStatusCode ();
+
+  if (mReportStatusCodeLibStatusCodeProtocol != NULL) {
+    gBS->CloseEvent (mOnStatusCodeProtocolInstallEvent);
+  }
+}
+
+/**
+    The constructor attempts to locate and cache StatusCode Protocol. If not found, it will
+    register a notify event triggered
+
+    @param  ImageHandle   The firmware allocated handle for the EFI image.
+    @param  SystemTable   A pointer to the EFI System Table.
+
+    @retval EFI_SUCCESS   The constructor always returns EFI_SUCCESS.
+
+**/
+EFI_STATUS
+EFIAPI
+ReportStatusCodeLibConstructor (
+  IN EFI_HANDLE        ImageHandle,
+  IN EFI_SYSTEM_TABLE  *SystemTable
+  )
+{
+  EFI_STATUS  Status;
+  VOID        *Registration;
+
+  InternalGetReportStatusCode ();
+
+  //
+  // Register for protocol installation notification if the protocol is not found
+  //
+  if (mReportStatusCodeLibStatusCodeProtocol == NULL) {
+    ASSERT (gBS != NULL);
+
+    //
+    // Create the event
+    //
+    Status = gBS->CreateEvent (
+                    EVT_NOTIFY_SIGNAL,
+                    TPL_CALLBACK,
+                    OnStatusCodeRuntimeProtocolInstalledNotification,
+                    NULL,
+                    &mOnStatusCodeProtocolInstallEvent
+                    );
+    ASSERT_EFI_ERROR (Status);
+
+    //
+    // Register for protocol notifications on this event
+    //
+    Status = gBS->RegisterProtocolNotify (
+                    &gEfiStatusCodeRuntimeProtocolGuid,
+                    mOnStatusCodeProtocolInstallEvent,
+                    &Registration
+                    );
+    ASSERT_EFI_ERROR (Status);
+  }
+
+  return EFI_SUCCESS;
+}
+
+/**
+    Closes the protocol notification event, if still open, when the library is unloaded.
+
+    @param  ImageHandle   The firmware allocated handle for the EFI image.
+    @param  SystemTable   A pointer to the EFI System Table.
+
+    @retval EFI_SUCCESS   The constructor always returns EFI_SUCCESS.
+
+**/
+EFI_STATUS
+EFIAPI
+ReportStatusCodeLibDestructor (
+  IN EFI_HANDLE        ImageHandle,
+  IN EFI_SYSTEM_TABLE  *SystemTable
+  )
+{
+  EFI_STATUS  Status;
+
+  if (mReportStatusCodeLibStatusCodeProtocol == NULL) {
+    Status = gBS->CloseEvent (mOnStatusCodeProtocolInstallEvent);
+    ASSERT_EFI_ERROR (Status);
+  }
+
+  return EFI_SUCCESS;
+}
+
+// MU_CHANGE [END] - Add constructor to cache protocol or register for protocol notification
--- a/MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.c
+++ b/MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.c
@@ -1,0 +1,84 @@
+/** @file
+  UEFI based application for unit testing the DxeReportStatusCodeLib.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+#include <TestDxeReportStatusCodeLibTestApp.h>
+
+#define UNIT_TEST_NAME     "Dxe Report Status Code Lib Unit Test Application"
+#define UNIT_TEST_VERSION  "0.1"
+
+UNIT_TEST_STATUS
+EFIAPI
+TestReportStatusCodeDoesNotInvertTpl (
+  IN UNIT_TEST_CONTEXT  Context
+  )
+{
+  EFI_STATUS        Status;
+  EFI_TPL           OldTpl;
+  EFI_SYSTEM_TABLE  *SystemTable = (EFI_SYSTEM_TABLE *)Context;
+
+  OldTpl = SystemTable->BootServices->RaiseTPL (TPL_HIGH_LEVEL);
+
+  Status = ReportStatusCode (EFI_PROGRESS_CODE, 0x0);
+
+  SystemTable->BootServices->RestoreTPL (OldTpl);
+
+  if (EFI_ERROR (Status)) {
+    UT_LOG_ERROR ("ReportStatusCode returned error: %r", Status);
+    return UNIT_TEST_ERROR_TEST_FAILED;
+  }
+
+  return UNIT_TEST_PASSED;
+}
+
+EFI_STATUS
+EFIAPI
+UefiTestMain (
+  IN EFI_HANDLE        ImageHandle,
+  IN EFI_SYSTEM_TABLE  *SystemTable
+  )
+{
+  EFI_STATUS                  Status;
+  UNIT_TEST_FRAMEWORK_HANDLE  Framework;
+  UNIT_TEST_SUITE_HANDLE      NoTplInversionTestSuite;
+
+  Status                  = EFI_SUCCESS;
+  Framework               = NULL;
+  NoTplInversionTestSuite = NULL;
+
+  DEBUG ((DEBUG_INFO, "%a v%a\n", UNIT_TEST_NAME, UNIT_TEST_VERSION));
+
+  //
+  // Start setting up the test framework for running the tests.
+  //
+  Status = InitUnitTestFramework (&Framework, UNIT_TEST_NAME, gEfiCallerBaseName, UNIT_TEST_VERSION);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed in InitUnitTestFramework. Status = %r\n", Status));
+    Status = EFI_OUT_OF_RESOURCES;
+    goto Exit;
+  }
+
+  //
+  // Test Suite 1
+  //
+  Status = CreateUnitTestSuite (&NoTplInversionTestSuite, Framework, "No TPL Inversion Tests", "Dxe.ReportStatusCode.NoTplInversion", NULL, NULL);
+  if (EFI_ERROR (Status)) {
+    DEBUG ((DEBUG_ERROR, "Failed in CreateUnitTestSuite for No TPL Inversion Tests\n"));
+    Status = EFI_OUT_OF_RESOURCES;
+    goto Exit;
+  }
+
+  AddTestCase (NoTplInversionTestSuite, "Test ReportStatusCode TPL High", "TestReportStatusCodeTplHigh", TestReportStatusCodeDoesNotInvertTpl, NULL, NULL, SystemTable);
+
+  Status = RunAllTestSuites (Framework);
+
+Exit:
+  if (Framework != NULL) {
+    FreeUnitTestFramework (Framework);
+  }
+
+  return Status;
+}
--- a/MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.h
+++ b/MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.h
@@ -1,0 +1,24 @@
+/** @file
+  UEFI based application for unit testing the DxeReportStatusCodeLib.
+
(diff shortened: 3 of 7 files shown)


MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,69 @@ ReportDebugCodeEnabled (
...
+   @retval   none
+ **/
+VOID
+EFIAPI
+OnStatusCodeRuntimeProtocolInstalledNotification (
+  IN EFI_EVENT  Event,
+  IN VOID       *Context
+  )
+{
+  InternalGetReportStatusCode ();
```
apop5, line 632: Should the event get closed? I think gEfiStatusCodeRuntimeProtocolGuid is only installed once...

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,69 @@ ReportDebugCodeEnabled (
...
+  )
+{
+  EFI_EVENT   Event;
+  EFI_STATUS  Status;
+  VOID        *Registration;
+
+  InternalGetReportStatusCode ();
+
+  // If the Status Code Runtime Protocol is not found, register a notification event for when it is installed.
+  if ((mReportStatusCodeLibStatusCodeProtocol == NULL) && (gBS != NULL) && (gBS->CreateEvent != NULL) && (gBS->RegisterProtocolNotify != NULL)) {
```
apop5: if gBS is valid, then CreateEvent and RegisterProtocolNotify should also be valid

https://github.com/microsoft/mu_basecore/blob/release/202502/MdeModulePkg/Core/Dxe/DxeMain/DxeMain.c#L42

they are populated in the default structure (mBootServices )

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Ah yes. Thanks. Updated :)

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Fair enough! Updated :)

os-d (approved):
Can you take this to edk2, too?

Javagedes:
> Can you take this to edk2, too?

ya, I will. Figured we wanted this here first though

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,75 @@ ReportDebugCodeEnabled (
...
+
+    @param  ImageHandle   The firmware allocated handle for the EFI image.
+    @param  SystemTable   A pointer to the EFI System Table.
+
+    @retval EFI_SUCCESS   The constructor always returns EFI_SUCCESS.
+
+**/
+EFI_STATUS
+EFIAPI
+DxeReportStatusCodeLibConstructor (
```
makubacki: A convenience function sometimes used for this pattern is `EfiCreateProtocolNotifyEvent()` in UefiLib. I'm not necessarily suggesting that you use it, and you'd have to refactor some logic slightly to do so, but it's an option.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
makubacki: If there is a reinstall, are you sure you'd want to keep using the older protocol instance?

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,75 @@ ReportDebugCodeEnabled (
...
+                    NULL,
+                    &Event
+                    );
+
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_WARN, "DxeReportStatusCodeLibConstructor: CreateEvent failed: %r\n", Status));
+      return EFI_SUCCESS;
+    }
+
+    Status = gBS->RegisterProtocolNotify (
```
makubacki, line 682: Perhaps printing on error to make use of the return code would be useful here too.

MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.c
```diff
@@ -0,0 +1,84 @@
...
+    DEBUG ((DEBUG_ERROR, "Failed in CreateUnitTestSuite for No TPL Inversion Tests\n"));
+    Status = EFI_OUT_OF_RESOURCES;
+    goto EXIT;
+  }
+
+  AddTestCase (NoTplInversionTestSuite, "Test ReportStatusCode TPL High", "TestReportStatusCodeTplHigh", TestReportStatusCodeDoesNotInvertTpl, NULL, NULL, SystemTable);
+
+  Status = RunAllTestSuites (Framework);
+
+EXIT:
```
makubacki: nit: Convention would usually have this be `Exit`

MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.h
```diff
@@ -0,0 +1,24 @@
+/** @file
+  UEFI based application for unit testing the DxeReportStatusCodeLib.
+
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef _TEST_DXE_REPORT_STATUS_CODE_LIB_H_
+#define _TEST_DXE_REPORT_STATUS_CODE_LIB_H_
```
makubacki:
```suggestion
#ifndef TEST_DXE_REPORT_STATUS_CODE_LIB_H_
#define TEST_DXE_REPORT_STATUS_CODE_LIB_H_

```

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: As @apop5 mentioned, is this installed more than once? Or should I just be safe and do it anyway? Technically with the current implementation, it also only uses the first protocol found, regardless of future possible installations.

If that is the case, then I'd probably need to register the event callback even if we found a protocol during start up.

Javagedes: Another benefit if we do it this way, is I can just remove the `InternalGetReportStatusCode` from all library functions, as we can just assume (is it safe to assume this?) that the protocol will be cached if it exists.

makubacki: It is possible that it could be reinstalled simply because that's an option and we can't control all code that interacts with the protocol database out there.

The expectation is usually that what would you get from a current `LocateProtocol()` call is what would be used moving forward.

My understanding is that prior behavior would also freshly locate the protocol on a `ReportStatusCode()` invocation. If that's the case, it should continue to be supported.

makubacki: Though, in the current implementation, it will still update the global through `InternalGetReportStatusCode()`, right?

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: This was a good idea and I tried it out after your suggestion, but one of the things it does is signal the event once. Since this library constructor is also called for the DXE_CORE, the event list is not fully set up and we end up asserting.

Due to this, I will be closing this without implementing.

MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.c
Javagedes: Updated to `Exit`. Thanks!

MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.h
Javagedes: Updated. Thanks!

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Added an error check and log for the register function also. Thanks!

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Hmm That is a good point. `InternalGetReportStatusCode` returns early if the cached protocol is already set. So this library does not currently ever update the protocol if a new one is registered.

I can update the event callback to update it regardless, if you wish. 

But I cannot update `InternalGetReportStatusCode` or we will try and re-locate the protocol every time, which will make us continue to hit the current TPL inversion error.

makubacki: I'm a little concerned about the number of entries in the event database leaving this open too given it is linked against so many DXE modules. But I think that's technically the more robust behavior and I lean toward it if the reinstall case was previously supported to preserve that behavior.

makubacki: Actually, looking at the current code, it does not locate when the global is non-null, so I think closing is fine since that's the precedent.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,80 @@ ReportDebugCodeEnabled (
...
+                    &gEfiStatusCodeRuntimeProtocolGuid,
+                    Event,
+                    &Registration
```
kuqin12, line 685: I think we might need to make this `Registration` global and unregister is in a destructor. This is in case the driver unloads itself, then we will need leave a dangling pointer in the protocol DB.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,80 @@ ReportDebugCodeEnabled (
...
+  EFI_STATUS  Status;
+  VOID        *Registration;
+
+  InternalGetReportStatusCode ();
+
+  // If the Status Code Runtime Protocol is not found, register a notification event for when it is installed.
+  if ((mReportStatusCodeLibStatusCodeProtocol == NULL) && (gBS != NULL)) {
+    Status = gBS->CreateEvent (
+                    EVT_NOTIFY_SIGNAL,
+                    TPL_CALLBACK,
```
kuqin12, line 672: nit: we probably can make this run at a higher level. But this is just to extend the coverage of RSC calls.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
kuqin12, line 46: We need to remove this locate call. Otherwise, a call before our callback will still run into the same issue?

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
```diff
@@ -611,3 +611,80 @@ ReportDebugCodeEnabled (
...
+DxeReportStatusCodeLibConstructor (
+  IN EFI_HANDLE        ImageHandle,
+  IN EFI_SYSTEM_TABLE  *SystemTable
+  )
+{
+  EFI_EVENT   Event;
+  EFI_STATUS  Status;
+  VOID        *Registration;
+
+  InternalGetReportStatusCode ();
```
kuqin12, line 659: And instead, we need to move the locate here.

MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.c
```diff
@@ -0,0 +1,84 @@
+/** @file
+  UEFI based application for unit testing the DxeReportStatusCodeLib.
```
kuqin12, line 2: I like the tests :)

MdeModulePkg/MdeModulePkg.dsc
```diff
@@ -351,6 +351,15 @@
...
+  # MU_CHANGE [BEGIN] - Add constructor to cache protocol or register for protocol notification
+  MdeModulePkg/Library/DxeReportStatusCodeLib/TestDxeReportStatusCodeLibTestApp.inf {
+    <LibraryClasses>
+        UnitTestLib|UnitTestFrameworkPkg/Library/UnitTestLib/UnitTestLib.inf
+        UnitTestPersistenceLib|UnitTestFrameworkPkg/Library/UnitTestPersistenceLibNull/UnitTestPersistenceLibNull.inf
+        UnitTestResultReportLib|UnitTestFrameworkPkg/Library/UnitTestResultReportLib/UnitTestResultReportLibDebugLib.inf
+        UnitTestBootLib|UnitTestFrameworkPkg/Library/UnitTestBootLibNull/UnitTestBootLibNull.inf
+  }
+  # MU_CHANGE [END] - Add constructor to cache protocol or register for protocol notification
   MdeModulePkg/Library/RuntimeDxeReportStatusCodeLib/RuntimeDxeReportStatusCodeLib.inf
```
kuqin12, line 363: I think you will need to make the same change for this runtime instance.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: I mentioned this in the PR comment, but this change merely follows the requested change in the attached issue and does not fully stop the possibility of a TPL inversion.

As this PR mentions, it is still possible to have an inversion if a driver is running before any Status Code protocol has been globally installed, and calls `ReportStatusCode` while in a TPL higher than `TPL_NOTIFY`.

I'm more than happy to fundamentally change the logic of this to only set the global via the constructor & protocol install event, removing the LocateProtocol from `ReportStatusCode` all together.

If that is your preference, just let me know.

kuqin12: So the proposal in the issue is to remove the locate on the fly and move it to the constructor only, so that we do not run into the same issue. We are already changing the logic quite a bit and adding a brand new test, might as well fully prevent the inversion from happening.

Javagedes: Sounds good. Seems like I misunderstood the proposal. I will make the changes and re request reviews :)

kuqin12: Thanks a lot!

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Cached the `Event`, not the registration, so that I can call `gBS->CloseEvent(...)` in the deconstructor.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: It can run at any level up to TPL_NOTIFY. If you have a preference, I will update. Just let me know which one you want.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: `InternalGetReportStatusCode ();` is what does the locate. so this portion is already correct.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
Javagedes: Per this conversation and offline conversation, I am updating this library to no longer attempt to get the report status code protocol (if not found) during a `ReportStatusCode` call. This will completely prevent the possibility of TPL inversion.

After changes, we do an initial `GetProtocol` in the constructor. If it fails to find the protocol, we instead create an on protocol event callback to cache the protocol.

The event will only ever be called once (uninstalling itself).

If the library is unloaded and the event has not been called, we will close the event.

MdeModulePkg/MdeModulePkg.dsc
Javagedes: This library instance is the backing library for `DXE_CORE`, `DXE_DRIVER`, and `DXE_RUNTIME_DRIVER` (among multiple others), so this should suffice.

Javagedes:
I have re-requested all reviews, as I have made a fundamental change to the pull request.

MdeModulePkg/Library/DxeReportStatusCodeLib/ReportStatusCodeLib.c
makubacki: I assume the request was to prioritize setting the global here to preempt TPL_CALLBACK notifications.

makubacki (approved):

kuqin12 (approved):

kuqin12 (approved):

apop5 (approved):

PR #1605, opened 2025-12-19
MdePkg: StandaloneMmDriverEntryPoint remove unnecessary dependency on MmServicesTableLib

Flickdm:
## Description

This pull request refactors how the MM System Table is accessed within the `StandaloneMmDriverEntryPoint.c` file. Instead of passing the system table pointer around, it introduces a cached global pointer, simplifying function calls and improving code clarity. The unnecessary dependency on `MmServicesTableLib` is also removed.

Refactoring of MM System Table access:

* Introduced a global variable `mCachedMmSystemTable` to cache the MM System Table pointer during driver initialization and replaced all usages of direct `MmSystemTable` or `gMmst` with this cached pointer throughout the file (`StandaloneMmDriverEntryPoint.c`). [[1]](diffhunk://#diff-ad037888cff8416996203a49ee4dd3bca86496ace3031f7552030b0f5217f575L17-R25) [[2]](diffhunk://#diff-ad037888cff8416996203a49ee4dd3bca86496ace3031f7552030b0f5217f575R97-R123) [[3]](diffhunk://#diff-ad037888cff8416996203a49ee4dd3bca86496ace3031f7552030b0f5217f575L122-R143) [[4]](diffhunk://#diff-ad037888cff8416996203a49ee4dd3bca86496ace3031f7552030b0f5217f575L50-R56)

Dependency cleanup:

* Removed the unused `MmServicesTableLib` from both the includes in `StandaloneMmDriverEntryPoint.c` and the `LibraryClasses` section in the INF file (`StandaloneMmDriverEntryPoint.inf`). [[1]](diffhunk://#diff-ad037888cff8416996203a49ee4dd3bca86496ace3031f7552030b0f5217f575L17-R25) [[2]](diffhunk://#diff-8f18183779e42959206c0c46d0d1f455822450b0181da84d1b6eb2d3a53ad7c4L38)

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

QemuQ35

## Integration Instructions

N/A

### DIFF
--- a/MdePkg/Library/StandaloneMmDriverEntryPoint/StandaloneMmDriverEntryPoint.c
+++ b/MdePkg/Library/StandaloneMmDriverEntryPoint/StandaloneMmDriverEntryPoint.c
@@ -14,8 +14,15 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Protocol/LoadedImage.h>
 #include <Library/BaseLib.h>
 #include <Library/DebugLib.h>
-#include <Library/MmServicesTableLib.h>
+// #include <Library/MmServicesTableLib.h> // MU_CHANGE
 #include <Library/StandaloneMmDriverEntryPoint.h>
+
+// MU_CHANGE [BEGIN]
+//
+// Cached pointer to the MM System Table that is set during driver initialization.
+//
+EFI_MM_SYSTEM_TABLE  *mCachedMmSystemTable = NULL;
+// MU_CHANGE [END]
 
 /**
   Unloads an image from memory.
@@ -46,9 +53,12 @@ _DriverUnloadHandler (
   // library destructors.  If the unload handler returned an error, then the driver can not be
   // unloaded, and the library destructors should not be called
   //
+  // MU_CHANGE [BEGIN]
   if (!EFI_ERROR (Status)) {
-    ProcessLibraryDestructorList (ImageHandle, gMmst);
+    ProcessLibraryDestructorList (ImageHandle, mCachedMmSystemTable);
   }
+
+  // MU_CHANGE [END]
 
   //
   // Return the status from the driver specific unload handler
@@ -88,12 +98,19 @@ _ModuleEntryPoint (
   EFI_STATUS                 Status;
   EFI_LOADED_IMAGE_PROTOCOL  *LoadedImage;
 
+  // MU_CHANGE [BEGIN]
+
+  //
+  // Cache the System Table
+  //
+  mCachedMmSystemTable = MmSystemTable;
+
   if (_gMmRevision != 0) {
     //
     // Make sure that the MM spec revision of the platform
     // is >= MM spec revision of the driver
     //
-    if (MmSystemTable->Hdr.Revision < _gMmRevision) {
+    if (mCachedMmSystemTable->Hdr.Revision < _gMmRevision) {
       return EFI_INCOMPATIBLE_VERSION;
     }
   }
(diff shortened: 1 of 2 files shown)


makubacki:
@Flickdm, global service tables accessed through service tables libraries is a common pattern. Is there more to this change? Afaict, the main issue with the previous code was that it used `gMmst` instead of `MmSystemTable` in the `gMmst->MmHandleProtocol()` call in the constructor.

os-d:
> @Flickdm, global service tables accessed through service tables libraries is a common pattern. Is there more to this change? Afaict, the main issue with the previous code was that it used `gMmst` instead of `MmSystemTable` in the `gMmst->MmHandleProtocol()` call in the constructor.

This is for OneCrypto, from a Sean request that I echoed. If the entry point lib doesn't use the services table lib, then the multiphase binary doesn't have to link in the services table lib. This helps to reduce the dependency chain; if something in the future that OneCrypto was relying on added a call to gMmst, we currently wouldn't catch it until it broke in runtime. If the library isn't linked in, we'd catch it at build time.

os-d (approved):
Not everything (including INF) covered by MU_CHANGE tags and commenting out old lines instead of removing.

Flickdm:
> Not everything (including INF) covered by MU_CHANGE tags and commenting out old lines instead of removing.

I can make that change real quick

makubacki:
> This is for OneCrypto, from a Sean request that I echoed. If the entry point lib doesn't use the services table lib, then the multiphase binary doesn't have to link in the services table lib. This helps to reduce the dependency chain; if something in the future that OneCrypto was relying on added a call to gMmst, we currently wouldn't catch it until it broke in runtime. If the library isn't linked in, we'd catch it at build time.

Is that code allowed to call the same services from the `MmSystemTable` pointer passed to their constructors and the entry point?

os-d:
> Is that code allowed to call the same services from the `MmSystemTable` pointer passed to their constructors and the entry point?

In the MM part of the driver, yes. But that is only done when loaded by StMM Core (when it calls _ModuleEntryPoint() which calls the constructors). In the DXE side of the driver, definitely not, but that path doesn't exist because there is a custom entry point and so the library constructors aren't called. But that's the concern, that the DXE side should never accidentally call gMmst, which this change lets us enforce at build time.

MdePkg/Library/StandaloneMmDriverEntryPoint/StandaloneMmDriverEntryPoint.c
```diff
@@ -14,9 +14,16 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
...
 #include <Library/DebugLib.h>
-#include <Library/MmServicesTableLib.h>
+// #include <Library/MmServicesTableLib.h> // MU_CHANGE
 #include <Library/StandaloneMmDriverEntryPoint.h>
 
+// MU_CHANGE [BEGIN]
+//
+// Cached pointer to the MM System Table that is set during driver initialization.
+//
+EFI_MM_SYSTEM_TABLE  *mCachedMmSystemTable = NULL;
```
makubacki, line 24: nit: We don't normally include "cached" in the name of global pointer variables. I think `mMmSystemTable` is sufficient. But this is personal preference, feel free to keep this name if you like it.

makubacki (approved):
I'm fine to approve the change for the one crypto use case. A few points:

- This is unintuitive for the normal MM case. `MmServicesTableLib` already does exactly this and exposes the MM services table through a public symbol that is well established. This redundantly caches the same pointer in the same module when it is linked against `MmServicesTableLib` (which is nearly all MM modules).
- It is also unintuitive to say that a library class restricted to certain module types cannot use another library class of the same type for the purpose that the services exclusive to the module type cannot be used.

---

However, there is little practical downside to this change, and it is not a breaking change, I just want to raise those points that you'll likely encounter when pushing it more broadly.

PR #1609, opened 2026-01-06
BaseTools: Add support for preserving build ID

cfernald:
## Description

Adds an optional flag that will copy the build-id section from the input ELF file to the output firmware image. The Build ID is used for debugging and allows tools to match the firmware image with the correct debug symbols.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on Q35 & SBSA with GCC5 toolchain

## Integration Instructions

N/A

### DIFF
--- a/BaseTools/Source/C/GenFw/Elf64Convert.c
+++ b/BaseTools/Source/C/GenFw/Elf64Convert.c
@@ -304,6 +304,11 @@ IsBuildIdShdr (
 {
   Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
 
+  if (Namedr->sh_offset + Shdr->sh_name >= mFileBufferSize) {
+    assert(FALSE);
+    return FALSE;
+  }
+
   return (BOOLEAN) (strcmp((CHAR8*)mEhdr + Namedr->sh_offset + Shdr->sh_name, ELF_BUILD_ID_SECTION_NAME) == 0);
 }
 
--- a/BaseTools/Source/C/GenFw/Elf64Convert.c
+++ b/BaseTools/Source/C/GenFw/Elf64Convert.c
@@ -130,6 +130,8 @@ STATIC UINT32 mHiiRsrcOffset;
 STATIC UINT32 mRelocOffset;
 STATIC UINT32 mDebugOffset;
 STATIC UINT32 mExportOffset;
+STATIC UINT32 mBuildIdOffset;
+STATIC BOOLEAN mBuildIdFound;
 //
 // Used for RISC-V relocations.
 //
@@ -227,6 +229,10 @@ InitializeElf64 (
     ElfFunctions->WriteExport = WriteExport64;
   }
 
+  if (mBuildIdFlag) {
+    mCoffNbrSections ++;
+  }
+
   return TRUE;
 }
 
@@ -288,6 +294,17 @@ IsHiiRsrcShdr (
   Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
 
   return (BOOLEAN) (strcmp((CHAR8*)mEhdr + Namedr->sh_offset + Shdr->sh_name, ELF_HII_SECTION_NAME) == 0);
+}
+
+STATIC
+BOOLEAN
+IsBuildIdShdr (
+  Elf_Shdr *Shdr
+  )
+{
+  Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
+
+  return (BOOLEAN) (strcmp((CHAR8*)mEhdr + Namedr->sh_offset + Shdr->sh_name, ELF_BUILD_ID_SECTION_NAME) == 0);
 }
 
 STATIC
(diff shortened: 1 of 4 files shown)


BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -1233,18 +1284,41 @@ ScanSections64 (
...
     // Don't make a section of size 0.
     NtHdr->Pe32Plus.FileHeader.NumberOfSections--;
   }
 
+  //
+  // Add build-ID section if requested and found
+  //
+  if (mBuildIdFlag) {
+    if (mBuildIdFound) {
+      CreateSectionHeader (".bldid", mBuildIdOffset, mRelocOffset - mBuildIdOffset,
```
cfernald, line 1318: I'm on the fence here. It feels odd to change the section name from `.build-id` to `.bldid`. This is currently being done because of the direct section size name limit. Longer names can be used, but required introducing complexity to genfw. My initial thinking is that this image already doesn't conform to image standards so the rename is worth it to avoid adding complexity, though perhaps that is lazy rationalization. Thoughts?

os-d: Is changing the section name going to affect any existing ELF parsing tools that work or kind of work on the franken-image? i.e. is WinDbg going to look for .build-id when we tell it this in actually an ELF image? Presumably not or you wouldn't be going for .bldid. Will it affect gdb debugging this and seeing this .bldid?

I would tend to leave the section name as is if it isn't a huge amount of complexity. But, I don't have a strong feeling on it. I would suggest taking this to edk2 and getting feedback there (we should upstream this anyway).

os-d (approved):

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -1114,6 +1131,40 @@ ScanSections64 (
     }
   }
 
+  //
+  //  The build-ID section.
+  //
+  mBuildIdOffset = mCoffOffset;
```
Javagedes, line 1142: should we be gating this behind `mBuildIdFlag == TRUE` ?

Javagedes: I Should have "started a review" so I could have quietly deleted this comment. Looks like it's necessary regardless for the .rsrc section logic below.

Javagedes (approved):

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -1114,6 +1131,40 @@ ScanSections64 (
...
+          mCoffOffset += (UINT32) shdr->sh_size;
+          mCoffOffset = CoffAlign(mCoffOffset);
+          mBuildIdFound = TRUE;
+        }
+        break;
+      }
+    }
+
+    if (!mBuildIdFound) {
+      Warning (NULL, 0, 0, NULL, "Build ID section is not found in %s.", mInImageName);
```
makubacki, line 1169: In `InitializeElf64()` you had:

```c
  if (mBuildIdFlag) {
    mCoffNbrSections ++;
  }
```

Does `mCoffNbrSections` need to be decremented heere if the build-id section is not found?

BaseTools/Source/C/GenFw/Elf64Convert.c
```diff
@@ -290,6 +296,17 @@ IsHiiRsrcShdr (
...
 
+STATIC
+BOOLEAN
+IsBuildIdShdr (
+  Elf_Shdr *Shdr
+  )
+{
+  Elf_Shdr *Namedr = GetShdrByIndex(mEhdr->e_shstrndx);
+
+  return (BOOLEAN) (strcmp((CHAR8*)mEhdr + Namedr->sh_offset + Shdr->sh_name, ELF_BUILD_ID_SECTION_NAME) == 0);
```
makubacki, line 312: It feels like there should be additional validation to check bounds here. Is that covered somewhere else? For example, in other places something like this checks the offset against the file size.

```c
        if (Shdr->sh_offset + Shdr->sh_size > mFileBufferSize) {
          return FALSE;
        }
```

makubacki (approved):

BaseTools/Source/C/GenFw/Elf64Convert.c
makubacki: @cfernald, can you clarify what tools are expected to process the copied section?

`build-id` is mentioned in [elf docs](https://www.man7.org/linux/man-pages/man5/elf.5.html) and [gdb docs](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Separate-Debug-Files.html), so it'd probably be simpler to just preserve the name to reduce confusion.

cfernald: yeah, the reason I wasn't original concerned with that is because we are not generally following any sane mechanims for dwarf symbol resolution generally, so if we aren't following anything standard up to this point, why start now. That said, I can switch to the full redirected name.

cfernald: The only tool checking this would be our own extension.

cfernald: @makubacki @os-d , fun story. So I implemented string table support in genfw only to realize that the loader drops string table, which is apparently pretty standard. Meaning that even if we use `.build-id` it will just look like `/4` at runtime with no additional information. I'm going to leave this `.bldid`, obviously this is unfortunate, but considering no standard tools are going to make sense of this anyways I think it's better then the alternative of complicating genfw, adding even more file bloat, and having to change the EFI PE loader.

os-d: Makes sense to me. Thanks for trying.

BaseTools/Source/C/GenFw/Elf64Convert.c
cfernald: I'm doing it later on line 1318 for the actual image. Technically I could do it to the global, but all the other examples (e.g. HiiRsrc) do it this way so it's best to be consistent.

BaseTools/Source/C/GenFw/Elf64Convert.c
cfernald: Yeah, i frankly just copy-pasted the other header checks around this function, but better to be safe. Will add, though will make it an assert to avoid masking bad image issues.

os-d (approved):

PR #1614, opened 2026-01-15
Add StatusCodeprocessor to facilitate the parsing of Status Code

liqiqiii:
## Description

Add StatusCodeprocessor to facilitate the parsing of Status Code
A tool for parsing and decoding UEFI/EDK2 status codes from debug logs.
Supports both standard PI specification codes and platform-specific custom codes,
with automatic header discovery, macro resolution, and GUID-to-module name lookup.
For more details, please read README.md.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested with real hardware, both standard PIStatusCode and platform specific status code can be parsed correctly.


makubacki:
BaseTools may not the best place for this. I'd prefer we avoid coupling more collateral with BaseTools. BaseTools is not maintained very well (today) and it is nice to access simple scripts without having to pull a full mu_basecore/edk2 code tree. If there is consensus to keep it in BaseTools this is something that I think should go to edk2 first since it doesn't need to be a Mu-specific override (there is no time pressure to check it in).

makubacki:
Also, I don't have time to comment in detail, but the readme file appears very good at a high-level and this looks like a useful tool.

liqiqiii:
@makubacki Hi Michael, I have reopened a PR with edk2-pytool-extension style changes and update the readme too. https://github.com/tianocore/edk2-pytool-extensions/pull/1053. This PR closed.

PR #1615, opened 2026-01-15
CryptoPkg: Add BaseCryptInit(..)

Flickdm:
## Description

Add BaseCryptInit(..) to allow BaseCryptLib and its underlying crypto provider to be manually initialized

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built locally with OneCryptoPkg

## Integration Instructions

N/A

### DIFF
--- a/CryptoPkg/Include/Library/BaseCryptLib.h
+++ b/CryptoPkg/Include/Library/BaseCryptLib.h
@@ -4362,4 +4362,25 @@ GetCryptoProviderVersionString (
   IN OUT UINTN  *BufferSize
   );
 
+// =====================================================================================
+//    Initialization Functions
+// =====================================================================================
+
+/**
+  Manually initialize the cryptographic library.
+
+  Normally, the UEFI build system will take care of this through the libraries constructor.
+  However there are some use cases where its necessary to initialize the library
+  manually. What it means to "initialize" the library is provider-specific, but it
+  typically involves setting up accelerated instructions.
+
+  @retval EFI_SUCCESS  The library was initialized successfully.
+
+**/
+EFI_STATUS
+EFIAPI
+BaseCryptInit (
+  VOID
+  );
+
 #endif // __BASE_CRYPT_LIB_H__


os-d:
Can you remind me why this is a separate lib and not part of BaseCryptLib? Also, looking at your mu_crypto_release PR, I wasn't seeing the lib, is the lib getting implemented in OpensslPkg?

Flickdm:
> Can you remind me why this is a separate lib and not part of BaseCryptLib? 

I considered BaseCryptLib but I was trying to keep that file as "pure" as possible. With the rule that each function definition would map to a protocol and each protocol would be implemented.

Effectively if you're building against openssl directly - the phase specific loaders will call the correct constructor that calls the underlying function implicitly to initialize the CPU acceleration. Now in OneCrypto since the binary is defined as StandaloneMM or SupvMm it will initialize correctly. The only environment that cannot do that is DXE. Thus the need for in those environments to have a special way of initializing acceleration. 

Thus only OneCrypto really needs this capability

> Also, looking at your mu_crypto_release PR, I wasn't seeing the lib, is the lib getting implemented in OpensslPkg?

The implementation is in OpensslPkg 

https://github.com/microsoft/mu_crypto_release/pull/167/files

This PR has all the changes to get it working.

OpensslPkg/Library/BaseCryptSetupLib/BaseCryptSetupLib.c
OpensslPkg/Include/Library/OpensslLibSetup.h

os-d:
We discussed offline on consolidating this logic into BaseCryptLib and not introducing a new library for it. It follows a similar pattern as SerialPortInitialize.

CryptoPkg/Include/Library/BaseCryptLib.h
```diff
@@ -4362,4 +4362,25 @@ GetCryptoProviderVersionString (
...
   );
 
+// =====================================================================================
+//    Initialization Functions
+// =====================================================================================
+
+/**
+  Manually initialize the cryptographic library.
+
+  Normally, the UEFI build system will take care of this through the libraries driver entry.
```
os-d: Nit: constructor, not driver entry (I know what you meant)

os-d (approved):
BaseCryptLibNull and whatever other versions in CryptoPkg will need to include this function in this PR, even if just return success.

apop5 (approved):

PR #1618, opened 2026-01-22
[FF-A][TPM] Updates for ArmFfaLib to not use global variable pre-memory

Raymond-MS:
## Description

Updated the ARM FF-A libraries to remove global variables for the SEC version as well as to work with the necessary changes for the SEC version. SEC will now query all necessary information while other libraries can rely on the globals set in ArmFfaCommonInit.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built QEMU SBSA with TPM enabled, verified TPM communication and boot to UEFI shell.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -34,9 +34,6 @@
 
 #include "ArmFfaCommon.h"
 
-BOOLEAN  gFfaSupported;
-UINT16   gPartId;
-
 /**
   Convert EFI_STATUS to FFA return code.
 
@@ -188,22 +185,6 @@ ArmCallFfa (
   } else {
     ArmCallSvc ((ARM_SVC_ARGS *)FfaArgs);
   }
-}
-
-/**
-  Check FF-A support or not.
-
-  @retval TRUE                   Supported
-  @retval FALSE                  Not supported
-
-**/
-BOOLEAN
-EFIAPI
-IsFfaSupported (
-  IN VOID
-  )
-{
-  return gFfaSupported;
 }
 
 /**
@@ -837,15 +818,21 @@ ArmFfaLibMsgSendDirectReq (
 {
   EFI_STATUS    Status;
   ARM_FFA_ARGS  FfaArgs;
+  UINT16        PartId;
 
-  if ((DestPartId == gPartId) || (ImpDefArgs == NULL)) {
+  Status = ArmFfaLibGetPartId (&PartId);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if ((DestPartId == PartId) || (ImpDefArgs == NULL)) {
     return EFI_INVALID_PARAMETER;
   }
 
   ZeroMem (&FfaArgs, sizeof (ARM_FFA_ARGS));
 
   FfaArgs.Arg0 = ARM_FID_FFA_MSG_SEND_DIRECT_REQ;
-  FfaArgs.Arg1 = PACK_PARTITION_ID_INFO (gPartId, DestPartId);
+  FfaArgs.Arg1 = PACK_PARTITION_ID_INFO (PartId, DestPartId);
   FfaArgs.Arg2 = Flags;
   FfaArgs.Arg3 = ImpDefArgs->Arg0;
   FfaArgs.Arg4 = ImpDefArgs->Arg1;
@@ -892,6 +879,7 @@ ArmFfaLibMsgSendDirectReq2 (
   EFI_STATUS    Status;
   UINT64        Uuid[2];
   ARM_FFA_ARGS  FfaArgs;
+  UINT16        PartId;
 
   /*
    * Direct message request 2 is only supported on AArch64.
@@ -900,7 +888,12 @@ ArmFfaLibMsgSendDirectReq2 (
     return EFI_UNSUPPORTED;
   }
 
-  if ((DestPartId == gPartId) || (ImpDefArgs == NULL)) {
+  Status = ArmFfaLibGetPartId (&PartId);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if ((DestPartId == PartId) || (ImpDefArgs == NULL)) {
     return EFI_INVALID_PARAMETER;
   }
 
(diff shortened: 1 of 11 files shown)


MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
cfernald: Should these globals still be in "common" if common is used for SEC without global variable access? Seems like it would make it easy for someone to mess up and use them in the future.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -971,44 +967,17 @@ ArmFfaLibCommonInit (
...
-
-  gFfaSupported = FALSE;
 
-  Status = ArmFfaLibGetVersion (
-             ARM_FFA_MAJOR_VERSION,
-             ARM_FFA_MINOR_VERSION,
-             &CurrentMajorVersion,
-             &CurrentMinorVersion
-             );
+  Status = ArmFfaLibIsFfaSupported(&gFfaSupported);
```
cfernald: Similar to the globals comment. I think we need to be more consistent on what code does and doesn't use globals.

Raymond-MS: This common init is what initializes these globals. The use of them comes from the get functions which are implemented in their corresponding phase libraries. For SEC, the "get" functions always go through the FF-A calls to get the global information while all other phase implementations will use the globals.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
Raymond-MS: I think this is fine with how our current implementation works. I should have addressed all access issues within SEC with this change.

kuqin12: I do not think they should be here if we do not use them, especially you are already touching up all the instances.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
```diff
@@ -86,4 +86,34 @@ GetRxTxBufferMinSizeAndAlign (
...
+  @param[out] IsSupported  If FF-A is supported
+
+  @retval EFI_SUCCESS            FF-A is supported
+  @retval EFI_UNSUPPORTED        FF-A isn't supported
+
+**/
+EFI_STATUS
+EFIAPI
+ArmFfaLibIsFfaSupported (
+  OUT BOOLEAN  *IsSupported
```
kuqin12: This input seems redundant if your return code already indicate whether this is supported.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
kuqin12: Would it make better sense to have the common init to do the query, and return the values. Different instances can chose to cache them in global variables vs. discard?

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -248,26 +316,23 @@ ArmFfaLibRxTxUnmap (
...
   if (Buffers != NULL) {
     FreePages (Buffers, (PcdGet64 (PcdFfaTxRxPageCount) * 2));
   }
 
-  mTxBuffer = NULL;
-  mRxBuffer = NULL;
+  /*
+   * Invalidate the HOB.
+   */
+  ZeroMem (RxTxBufferAllocationHob, sizeof (ARM_FFA_RX_TX_BUFFER_INFO));
```
kuqin12: I know it is outside of this specific PR, but the same applies to PEI, correct? I think we need to zero the HOB if this is unmapped in PEI as well?

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
Raymond-MS: Removed the globals and moved them to their respective phase specific ArmFfaLibs.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
Raymond-MS: Updated to remove the globals.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.h
Raymond-MS: Fixed.

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: So I have been digging into this a little. My understanding is that if for whatever reason Unmap is called on the RX/TX buffer during the SEC phase the HOB should be invalidated to let the next phase know there are no mapped buffers. In the PEI phase, it uses the normal RX/TX map library which doesn't do this on unmap. Should we update the PEI library to use the SEC version of the RX/TX map library? As far as I can tell that is the only way we can make sure the HOB is invalidated when unmap is called. We wouldn't want to put that specific code into the normal RX/TX map library like SEC does.

MdeModulePkg/Library/ArmFfaLib/ArmFfaRxTxMap.c
```diff
@@ -189,11 +189,14 @@ ArmFfaLibRxTxUnmap (
   EFI_STATUS    Status;
   ARM_FFA_ARGS  FfaArgs;
   VOID          *Buffers;
+  UINT16        SourceId;
```
cfernald: nit: partition ID, for consistency

cfernald: nit: partition ID, for consistency

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
```diff
@@ -232,14 +282,32 @@ ArmFfaLibRxTxUnmap (
...
-  ARM_FFA_ARGS  FfaArgs;
-  VOID          *Buffers;
+  EFI_STATUS                 Status;
+  ARM_FFA_ARGS               FfaArgs;
+  VOID                       *Buffers;
+  EFI_HOB_MEMORY_ALLOCATION  *RxTxBufferAllocationHob;
+  UINT16                     PartId;
+
+  /*
+   * Rx/Tx Buffer are allocated with continuous pages.
```
cfernald:
```suggestion
   * Rx/Tx buffers are allocated with continuous pages.
```

cfernald (approved):

MdeModulePkg/Library/ArmFfaLib/ArmFfaSecRxTxMap.c
Raymond-MS: As per our conversation, added a callback that can invalidate the HOBs in PEI and SEC when Unmap is called.

kuqin12 (approved):

Raymond-MS:
These changes are being upstreamed to EDK2. Once they get merged, I will close this and open another commit pulling down the changes.

Raymond-MS:
These changes were upstreamed to EDK2. Will open a PR cherry-picking these changes down into MU_BASECORE. Closing as duplicate. This is no longer needed.

PR #1634, opened 2026-02-09
UefiCpuPkg: Make X64 Page Table Definitions Public

os-d:
## Description

Currently, X64 page table structure definitions live in a private header. This moves them to the public
CpuPageTableLib.h header file so that other code can use these standard architectural definitions.

The main use case is for test code to consume these definitions to run memory protection tests.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Using memory protection tests.

## Integration Instructions

Include `<Library/CpuPageTableLib.h>` to get access to these architectural definitions.


kuqin12 (approved):

apop5 (approved):

UefiCpuPkg/Include/Library/CpuPageTableLib.h
```diff
@@ -63,6 +63,177 @@ typedef enum {
   PagingModeMax
 } PAGING_MODE;
 
+#define IA32_PE_BASE_ADDRESS_MASK_40  0xFFFFFFFFFF000ull
+#define IA32_PE_BASE_ADDRESS_MASK_39  0xFFFFFFFFFE000ull
+
+#define REGION_LENGTH(l)  LShiftU64 (1, (l) * 9 + 3)
```
os-d: I'm going to leave this definition behind in the private header, it is a bit too generically named and isn't an arch definition, as such. More of a helper.

kuqin12 (approved):

UefiCpuPkg/Include/Library/CpuPageTableLib.h
makubacki, line 1: Do these definitions need to in a library header? If they are generic paging structure definitions, they could go in a header file in `MdePkg/Include`.

os-d: Yeah, I had thought about that as well. AArch64 has an `MdePkg/Include/AArch64/AArch64.h` which has similar kind of definitions: https://github.com/tianocore/edk2/blob/f0542ae07d5e5e4d311bf8ae33bf26b4b1acf9f4/MdePkg/Include/AArch64/AArch64.h.

But X64 doesn't have the same (and I guess you would duplicate to IA32). Do you want me to create an X64/X64.h header? Something else?

makubacki: I wasn't entirely sure what made the most sense within MdePkg/Include either. Feel free to go with what you'd like. You'll probably get a request to move it somewhere different anyway.

os-d: I will leave this here in Mu, for now then, since we suspect it will be asked to be moved. I will ask the MdePkg maintainers in the edk2 PR if they think it is better housed there and if so where.

PR #1636, opened 2026-02-11
NetworkPkg: Add timeout polling for PXE media detection

### LINKED ISSUE #1635: [Task] Consider timeout for `NetLibDetectMedia()` calls in `NetLibDetectMediaWaitTimeout()`
During PXE boot, [`EfiPxeLoadFile()`](https://github.com/microsoft/mu_basecore/blob/b09f1fd5f3bf28ee3ba6d3bb7c25406246564b78/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c#L2375) is called to retrieve the Network Bootstrap Program (NBP) over network.

To begin, the code needs to check that network media is present. So, `NetLibDetectMediaWaitTimeout()` is called:

https://github.com/microsoft/mu_basecore/blob/b09f1fd5f3bf28ee3ba6d3bb7c25406246564b78/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.c#L2410-L2418

Note a few things about this call:

1. The return status code is not captured/checked.
2. A timeout of `PXEBC_CHECK_MEDIA_WAITING_TIME` is given:
   https://github.com/microsoft/mu_basecore/blob/b09f1fd5f3bf28ee3ba6d3bb7c25406246564b78/NetworkPkg/UefiPxeBcDxe/PxeBcImpl.h#L74

Also note a change made recent in commit https://github.com/microsoft/mu_basecore/commit/e9e821dbbe0e23fd05eab19afd5bfc79fa94276a. It changed the initial value of `*MediaState` from `EFI_SUCCESS` to `EFI_NOT_READY`.

In addition, the goal of `NetLibDetectMediaWaitTimeout()`:

>   Detect media state for a network device. This routine will wait for a period of time at
  a specified checking interval when a certain network is under connecting until connection
  process finishs or timeout. If Aip protocol is supported by low layer drivers, three kinds
  of media states can be detected: EFI_SUCCESS, EFI_NOT_READY and EFI_NO_MEDIA, represents
  connected state, connecting state and no media state respectively. When function detects
  the current state is EFI_NOT_READY, it will loop to wait for next time's check until state
  turns to be EFI_SUCCESS or EFI_NO_MEDIA. If Aip protocol is not supported, function will
  call NetLibDetectMedia() and return state directly.

---

Looking at the implementation of `NetLibDetectMediaWaitTimeout()`, the following checks are performed.

1. Get the SNP handle from the `ServiceHandle` (controller handle) given
2. Check if `gEfiAdapterInformationProtocolGuid` (AIP) is installed on that handle
   - If not, call `NetLibDetectMedia()` as a fallback for media detection
     https://github.com/microsoft/mu_basecore/blob/b09f1fd5f3bf28ee3ba6d3bb7c25406246564b78/NetworkPkg/Library/DxeNetLib/DxeNetLib.c#L2717-L2733

   - Notice that in this path, `*MediaState` is only updated if the `NetLibDetectMedia (ServiceHandle, &MediaPresent);` call returns success. Otherwise, it returns with the value of `*MediaState` initialized at the beginning of the function (`EFI_NOT_READY`). Prior to commit e9e821d, this was a bug (fixed in that commit) as it returned `*MediaState` set to `EFI_SUCCESS` when no media was detected at all.
   - Also see the `Status` is an error and not checked in `NetLibDetectMedia()` as noted above.
   - In summary, this creates a return path where a platform does not install `gEfiAdapterInformationProtocolGuid` (on the SNP handle) and simply by `NetLibDetectMedia()` returning an error, network media is considered present.
3. If AIP is installed on the SNP handle, the flow continues. Then, `Aip->GetInformation()` is called:
   
https://github.com/microsoft/mu_basecore/blob/b09f1fd5f3bf28ee3ba6d3bb7c25406246564b78/NetworkPkg/Library/DxeNetLib/DxeNetLib.c#L2741-L2778

   - If that does not report an error, then the `MediaInfo->MediaState` from the AIP instance is returned.
   - If it does return an error:
     - And that error is `EFI_UNSUPPORTED`, then the same `NetLibDetectMedia()` call flow occurs with the same problem.
     - If that error is not `EFI_UNSUPPORTED`, then `*MediaState` is also not updated and returns the initial value.
   - In summary, this creates two potential return paths where `*MediaState` is not updated properly.
     1. If `Aip->GetInformation()` returns an error that is not `EFI_UNSUPPORTED`
     2. If `Aip->GetInformation()` returns `EFI_UNSUPPORTED` and `NetLibDetectMedia()` reports any error.
4 If `Aip->GetInformation()` does not return an error and the media state it reports is `EFI_NOT_READY` (or a timeout check succeeds), then proceed to polling on `Aip->GetInformation()` using the `Timeout` value given to the function.

---

So,

1. The 20 second value often given as the timeout (`Timeout` argument) to `NetLibDetectMediaWaitTimeout()` is not used for `NetLibDetectMedia()` flows only `Aip->GetInformation()` calls.
2. We know some NICs take at least 2+ seconds to report back media presence after initialization when going through the `NetLibDetectMedia()` flow. That function supports two methods for media presence detection through SNP to the underlying UNDI driver: (1) `GET_STATUS` command and (2) a flow to stop and restart SNP for drivers that do not support the `GET_STATUS` command.

---

Given that, the issue proposes potentially tightening up status code/error handling and applying a timeout to the `NetLibDetectMedia()` calls so they have time to report presence.

---

It is also recommended that platforms and drivers installing `gEfiAdapterInformationProtocolGuid` try to accurately report media presence when `Aip->GetInformation()` is called with `gEfiAdapterInfoMediaStateGuid`.

### LINKED ISSUE #1638: [Task] Prevent multiple SNP shutdown and reinitialization invocations
`NetLibDetectMediaWaitTimeout()` in `DxeNetLib` is called 15 places in the codebase. Multiple callers might call into the function for the same NIC such as `Dhcp4Dxe`, `PxeBcDxe`, etc. For example:

```
  Snp->undi.initialize()  <-- By SimpleNetworkDriverStart

  Snp->undi.initialize()   <-- By NetLibDetectMedia()
  Snp->undi.initialize()  8000h:6h
  Snp->undi.initialize()

  Snp->undi.initialize()  <-- By NetLibDetectMedia()
  Snp->undi.initialize()  8000h:6h
  Snp->undi.initialize()
```

Initialization might also occur multiple times in the same flow, for example, when `SnpUndi32Initialize()` first calls with `PXE_OPFLAGS_INITIALIZE_DETECT_CABLE` and if that fails, again with `PXE_OPFLAGS_INITIALIZE_DO_NOT_DETECT_CABLE`:

https://github.com/microsoft/mu_basecore/blob/c4a845c8d5ac6e970cc3fd5160844bee01f9fb0e/NetworkPkg/SnpDxe/Initialize.c#L254-L277

Collectively this can lead to SNP shutting down and reinitializing multiple times when it might not need to impacting stability and performance.

---

This issue tracks exploration of a solution to reduce the number of times shutdown and reinitialization occurs across the network stack.

makubacki:
## Description

Resolves #1635
Resolves #1638

Simplify the `NetLibDetectMediaWaitTimeout()` execution flow and add a timeout on non-AIP media detection to allow NICs sufficient time to initialize and report media state.

By default, network devices must properly report media presence. However, the previous implementation would check `Aip->GetInformation()` fall back to `Snp->GetStatus()` and then return media is present if SNP returned that it did not support media detection. Falsely reporting media presence over the years has led to even more UNDI drivers not supporting media presence properly because of this behavior.

This change solves the following problems:

1. Allows sufficient time for SNP to report media presence instead of simply returning
   - It was verified that simply checking status and returning before missed presence reported.
2. Makes it clear to platform owners when media detection is not able to be determined so it is simply set to "media is present" to satisfy NIC FW that does not support actual media detection.
3. Partially due to (1) (and TPL guards introduced in previous recent changes), prevent multiple SNP shutdown and reinitialization processes from occurring which thrash NIC stability and impact network boot performance.

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- NetworkPkg CI
- Physical Intel client platform PXE boot
  - Does not need NIC OPROM workaround
- Physical Intel server platform PXE boot
  - Uses NIC OPROM workaround

## Integration Instructions

- N/A

### DIFF
--- a/NetworkPkg/Library/DxeNetLib/DxeNetLib.c
+++ b/NetworkPkg/Library/DxeNetLib/DxeNetLib.c
@@ -2437,6 +2437,126 @@ NetLibGetMacString (
   return EFI_SUCCESS;
 }
 
+// MU_CHANGE [BEGIN]: Consider timeout for NetLibDetectMedia() calls in NetLibDetectMediaWaitTimeout()
+
+/**
+  Get media status from Simple Network Protocol (SNP).
+
+  This function calls EFI_SIMPLE_NETWORK_PROTOCOL.GetStatus() to refresh
+  the media state and returns the current MediaPresent field from the SNP
+  mode data.
+
+  @param[in]  Snp           The Simple Network Protocol instance.
+  @param[out] MediaPresent  The pointer to receive the media present status.
+
+  @retval EFI_SUCCESS       MediaPresent is returned successfully.
+  @retval Others            Error returned by Snp->GetStatus().
+
+**/
+STATIC
+EFI_STATUS
+NetLibGetMediaStatus (
+  IN  EFI_SIMPLE_NETWORK_PROTOCOL  *Snp,
+  OUT BOOLEAN                      *MediaPresent
+  )
+{
+  EFI_STATUS  Status;
+  UINT32      InterruptStatus;
+
+  Status = Snp->GetStatus (Snp, &InterruptStatus, NULL);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  *MediaPresent = Snp->Mode->MediaPresent;
+  return EFI_SUCCESS;
+}
+
+/**
+  Wait for media to become present on a network device within a given timeout.
+
+  This helper periodically queries the Simple Network Protocol instance for
+  media state. It returns immediately if media is already present or if the
+  timeout is zero. Otherwise, it waits in MEDIA_STATE_DETECT_TIME_INTERVAL
+  increments until media becomes present or the timeout expires.
+
+  @param[in]  Snp           The Simple Network Protocol instance to query.
+  @param[in]  Timeout       The maximum number of 100ns units to wait. Zero
+                            means detect once and return immediately.
+  @param[out] MediaPresent  The pointer to receive the media present state.
+                            The value should be set to FALSE for this function
+                            to check for media presence.
+
+  @retval EFI_SUCCESS       Media is present or detected within the timeout.
+  @retval EFI_DEVICE_ERROR  Failed to create or use the timer event.
+  @retval EFI_TIMEOUT       The timeout expired before media became present.
+  @retval Others            Error returned by NetLibGetMediaStatus().
+
+**/
+STATIC
+EFI_STATUS
+NetLibWaitForMediaPresent (
+  IN  EFI_SIMPLE_NETWORK_PROTOCOL  *Snp,
+  IN  UINT64                       Timeout,
+  OUT BOOLEAN                      *MediaPresent
+  )
+{
+  EFI_STATUS  Status;
+  EFI_STATUS  TimerStatus;
+  EFI_EVENT   Timer;
+  INT64       TimeRemaining;
+
+  Status = NetLibGetMediaStatus (Snp, MediaPresent);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if (*MediaPresent || (Timeout == 0)) {
+    return EFI_SUCCESS;
+  }
+
+  Timer         = NULL;
+  TimeRemaining = Timeout;
+  Status        = gBS->CreateEvent (EVT_TIMER, TPL_CALLBACK, NULL, NULL, &Timer);
+  if (EFI_ERROR (Status)) {
+    return EFI_DEVICE_ERROR;
+  }
+
+  do {
+    Status = gBS->SetTimer (
+                    Timer,
+                    TimerRelative,
+                    MEDIA_STATE_DETECT_TIME_INTERVAL
+                    );
+    if (EFI_ERROR (Status)) {
+      gBS->CloseEvent (Timer);
+      return EFI_DEVICE_ERROR;
+    }
+
+    do {
+      TimerStatus = gBS->CheckEvent (Timer);
+      if (!EFI_ERROR (TimerStatus)) {
+        TimeRemaining -= MEDIA_STATE_DETECT_TIME_INTERVAL;
+        Status         = NetLibGetMediaStatus (Snp, MediaPresent);
+        if (EFI_ERROR (Status)) {
+          gBS->CloseEvent (Timer);
+          return Status;
+        }
+
+        if (*MediaPresent) {
+          gBS->CloseEvent (Timer);
+          return EFI_SUCCESS;
+        }
+      }
+    } while (TimerStatus == EFI_NOT_READY);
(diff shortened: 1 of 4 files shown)


makubacki:
Since `NetLibDetectMedia()` is in a public library header (`NetworkPkg/Include/Library/NetLib.h`), I will update this to leave that API unmodified. Add a new public API with a `Timeout` parameter and then have both call a common internal function where the original `NetLibDetectMedia()` implementation passes it a `Timeout` of `0`.

makubacki:
> Since `NetLibDetectMedia()` is in a public library header (`NetworkPkg/Include/Library/NetLib.h`), I will update this to leave that API unmodified. Add a new public API with a `Timeout` parameter and then have both call a common internal function where the original `NetLibDetectMedia()` implementation passes it a `Timeout` of `0`.

Pushed.

NetworkPkg/Library/DxeNetLib/DxeNetLib.c
```diff
@@ -2437,8 +2437,166 @@ NetLibGetMacString (
...
+STATIC
+EFI_STATUS
+NetLibSnpGetMediaStatus (
+  IN  EFI_SIMPLE_NETWORK_PROTOCOL  *Snp,
+  OUT BOOLEAN                      *MediaPresent
+  )
+{
+  EFI_STATUS  Status;
+  UINT32      InterruptStatus;
+
```
apop5, line 2466: NIT: should be check Snp and MediaPresent pointers?

apop5 (approved):

NetworkPkg/Library/DxeNetLib/DxeNetLib.c
os-d: I was going to make a similar comment, in that these are STATIC functions, so not too big a deal, but then if you go up a level to the callers, they are just doing the classic `ASSERT(!NULL);` in some cases, so no one is checking NULL

os-d (approved):

kuqin12 (approved):

NetworkPkg/Library/DxeNetLib/DxeNetLib.c
makubacki: It is a local helper, but it doesn't hurt to have the check either so I added it.

os-d (approved):

PR #1659, opened 2026-02-26
[CHERRY-PICK] MdeModulePkg: Allow dynamic generation of HEST ACPI table

os-d:
## Description

Introduce the HEST table generation protocol that allows platforms to build the table with multiple error source descriptors and install the table. The protocol provides two interfaces. The first interface allows for adding multiple error source descriptors into the HEST table. The second interface can then be used to dynamically install the fully populated HEST table. This allows multiple drivers and/or libraries to dynamically register error source descriptors into the HEST table.

Note: This is cherry-picked from https://git.gitlab.arm.com/arm-reference-solutions/edk2/-/commit/2dcd7973c295ec7e2a16b17126e6c5f5bbdfe98c not edk2. The owners are being worked with to upstream this to edk2.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical ARM64 platform.

## Integration Instructions

N/A.

### DIFF
--- a/MdeModulePkg/MdeModulePkg.dec
+++ b/MdeModulePkg/MdeModulePkg.dec
@@ -873,6 +873,11 @@
   gEdkiiGcdSyncCompleteProtocolGuid = { 0x650B7F40, 0x6564, 0x4FA9, { 0x93, 0x75, 0xCD, 0x6B, 0x52, 0x1B, 0x6E, 0x50 }}
   ## MU_CHANGE END
 
+  ## MU_CHANGE: START Cherry-pick HEST Driver
+  ## ACPI HEST table generation protocol
+  gHestTableProtocolGuid = { 0x705bdcd9, 0x8c47, 0x457e, { 0xad, 0x0d, 0xf7, 0x86, 0xf3, 0x4a, 0x0d, 0x63 } }
+  ## MU_CHANGE END
+
 [PcdsFeatureFlag]
   ## Indicates if the platform can support update capsule across a system reset.<BR><BR>
   #   TRUE  - Supports update capsule across a system reset.<BR>
--- a/MdeModulePkg/Universal/Apei/HestDxe/HestDxe.c
+++ b/MdeModulePkg/Universal/Apei/HestDxe/HestDxe.c
@@ -1,0 +1,365 @@
+/** @file
+  Dynamically builds and installs HEST ACPI table.
+
+  This driver implements protocol interfaces that can be used to create and
+  install HEST ACPI table. The interface allows one or more error source
+  producers to add the error source descriptors into the HEST table. And also
+  allow the resulting HEST table to be installed.
+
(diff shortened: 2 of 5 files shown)


MdeModulePkg/MdeModulePkg.dec
```diff
@@ -873,6 +873,11 @@
   gEdkiiGcdSyncCompleteProtocolGuid = { 0x650B7F40, 0x6564, 0x4FA9, { 0x93, 0x75, 0xCD, 0x6B, 0x52, 0x1B, 0x6E, 0x50 }}
   ## MU_CHANGE END
 
+  ## MU_CHANGE: START Cherry-pick HEST Driver
+  ## Arm Platform HEST table generation protocol
```
makubacki: Is this really specific to "Arm platforms"? If so, the protocol name should reflect that.

MdeModulePkg/Universal/Apei/HestDxe/HestDxe.c
```diff
@@ -0,0 +1,358 @@
...
+{
+  EFI_ACPI_6_4_HARDWARE_ERROR_SOURCE_TABLE_HEADER  *HestHeaderPtr;
+  HEST_DXE_DRIVER_DATA                             *HestDataList;
+
+  //
+  // Allocate memory for the head node of the list and HEST table header
+  // structure.
+  //
+  HestDataList                = AllocateZeroPool (sizeof (HEST_DXE_DRIVER_DATA));
+  HestDataList->HestTableData =
```
makubacki: Need to move the `(HestDataList == NULL)` before the dereference here.

makubacki (approved):
I know this is cherry-picked but I left some comments to fix and take as feedback. Looks like CI caught the library not being built in MdeModulePkg.dsc.

apop5 (approved):

MdeModulePkg/MdeModulePkg.dec
os-d: This I'd like to defer to the edk2 PR review. We have platforms using this currently (it existed in a previous branch of ours but was dropped because it was incorrectly believed unused), so I don't want to break them now and then potentially again if edk2 has a different request on naming.

cfernald (approved):

MdeModulePkg/MdeModulePkg.dec
makubacki: As I understand it, the comment is probably more the issue here than the protocol name. I'm just pointing out the two conflict right now.

os-d: Did a quick check, HEST is for any arch, it seems. Dropped the arm part of the comment.

kuqin12 (approved):

makubacki (approved):

PR #1666, opened 2026-03-02
MdeModulePkg: Add Device Control Protocol

eeshanl:
## Description

MdeModulePkg: Add Device Control Protocol

This protocol is used by the boot library to handoff control over a debug device to the OS.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on physical ARM platform.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Include/Protocol/DeviceControlProtocol.h
+++ b/MdeModulePkg/Include/Protocol/DeviceControlProtocol.h
@@ -1,0 +1,70 @@
+/** @file
+  Defines protocol and associated types for boot loader device control in the
+  UEFI environment.
+
+  Copyright (c) Microsoft Corporation.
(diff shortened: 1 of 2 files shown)


MdeModulePkg/Include/Protocol/DeviceControlProtocol.h
makubacki, line 1: Typically, public interfaces in MdeModulePkg would go to edk2 first. If the upstream doesn't agree with the change or requests a backward incompatible modification, that impact transfers to Mu consumers.

eeshanl:
closing, will not be going in mu

PR #1688, opened 2026-03-09
 [CHERRY-PICK] SecurityPkg: Fix Tcg2SubmitCommand in TPM field upgrade scenario [Rebase & FF]

liqiqiii:
Original commits from EDK2 is here: https://github.com/tianocore/edk2/commit/454e1fc40fe6535b5a89936c5377e33bbf965b8c
https://github.com/tianocore/edk2/commit/ba8296810c6c0f411f158955574c8ae8ffe283f0

There's a bug that after the device was forced shut down during the tpm
firmware update process, then the TPM will stay in field upgrade mode and
refuse to work properly. This will block TPM startup and cause the device
to bootloop.

Now in the TCG_DXE_DATA struct we don’t have a flag to indicate the
current TPM mode, and we will simply treat the upper Scenario with this
DEVICE_ERROR and set the TPMPresentFlag to FALSE.

Later in Tcg2SubmitCommand we will just return DEVICE_ERROR and skip the
actual TPM recovery.

We should have an extra flag that check for the TPM response code and if
it's TPM_RC_UPGRADE, we knew the device is in field upgrade mode and should
continue the workflow.

We should only return EFI_DEVICE_ERROR when both TPMPresentFlag and
TpmUpdateFlag are false.

The field upgrade is part of TCG spec, and the capsule update/recovery
is part of UEFI spec.
Make this PR to bring in the fix for this corner case.

The behavior will match the original design. Only difference is when TPMPresentFlag is false and response code is TPM_RC_UPGRADE.

### DIFF
--- a/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
+++ b/SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
@@ -1515,42 +1515,25 @@ Tcg2SubmitCommand (
                              &CurrentOutputBlockSize,
                              OutputParameterBlock
                              );
-  if (EFI_ERROR (Status)) {
-    return mTcgDxeData.BsCap.TPMPresentFlag ? Status : EFI_DEVICE_ERROR;
-  }
+  if (!mTcgDxeData.BsCap.TPMPresentFlag) {
+    // Special handling when TPM is thought to be absent
+    if ((CurrentOutputBlockSize >= sizeof (TPM2_RESPONSE_HEADER)) && !EFI_ERROR (Status)) {
+      // Command succeeded, check if it's actually TPM in update mode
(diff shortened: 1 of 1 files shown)


kuqin12:
I think you need to do `git cherry-pick <sha> -x` and then update the commit message title to start with `[CHERRY-PICK]`. Then we will be able to merge with rebase.

makubacki:
> I think you need to do `git cherry-pick <sha> -x` and then update the commit message title to start with `[CHERRY-PICK]`. Then we will be able to merge with rebase.

Also, cherry-pick each commit individually.

liqiqiii:
@kuqin12 @makubacki  Done. Thanks for the comment. First time doing this.

kuqin12 (approved):

makubacki (approved):

PR #1689, opened 2026-03-10
[2511] NetworkPkg/UefiPxeBcDxe: Add UDP Completion Tokens to PXEBC_PRIVATE_DATA

makubacki:
## Description

This change tracks UDP completion tokens in the `PXEBC_PRIVATE_DATA` structure, so they are persisted for cleanup in the Stop function in case it is called while a UDP operation is still pending.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- PXE boot with USB NIC disconnect to trigger `EfiPxeBcStop()` during UDP read

## Integration Instructions

- N/A

---

Note: A following PR may apply a similar change to the UDP write functions, but the issue was only observed in the UDP read functions, so this change is applied immediately for the time being.


makubacki:
Note: This change was already approved for 2502 in https://github.com/microsoft/mu_basecore/pull/1632. I lost track of merging it at the time. This just brings it to 2511 as well.

kuqin12 (approved):

apop5 (approved):

PR #1711, opened 2026-03-20
[SQUASH ON REBASE]fltused: Simplify and comment.

jaykrell:
## Description

The value and size of _fltused do not matter.
You just need a symbol.

Squash with 6c8a30ed32a26c6573a3bc0bef89ddc371228d82 on rebase.

- [ ] Impacts functionality? No.
- [ ] Impacts security? No.
- [ ] Breaking change? No.
- [ ] Includes tests? No, not needed.
- [ ] Includes documentation? No, not needed.

## How This Was Tested

Build.

## Integration Instructions

Nothing special.

### DIFF
--- a/MdePkg/Library/FltUsedLib/FltUsedLib.c
+++ b/MdePkg/Library/FltUsedLib/FltUsedLib.c
@@ -8,4 +8,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 // create a global to satisfy the compilers insertion of the _fltused
 // in reponse to detecting floating point type operations.
-int  _fltused = 0x9875;
+//
+// In some systems (not EDK2) this is used to tune static C runtime linking.
+// i.e. There are (larger) working and (small stub) non-working floating point functions,
+// and _fltused is in the object files with the working functions.
+char  _fltused;


os-d:
@jaykrell did you run into an issue with how it is currently defined?

jaykrell:
It wastes space, and worse, people will wonder what this 9875 values means.
It means nothing, and is unnecessary.

The ntdll definition is this:

// Make _fltused static in ntdll.lib instead of an import.
extern const char _fltused = 0;

I was chosing readonly over writable.

Windows has a variety of definitions, 0, 1, 9875, 9876.
The non-zero values are just random.

os-d:
> It wastes space, and worse, people will wonder what this 9875 values means. It means nothing, and is unnecessary.
> 
> The ntdll definition is this:
> 
> // Make _fltused static in ntdll.lib instead of an import. extern const char _fltused = 0;
> 
> I was chosing readonly over writable.
> 
> Windows has a variety of definitions, 0, 1, 9875, 9876. The non-zero values are just random.

I agree, I was just trying to understand if you had hit a specific issue with this or just a visual inspection. If visual inspection, fine, if a specific issue wanted to make sure this was the right fix and there weren’t pieces we were missing.

os-d (approved):

apop5:
> It wastes space, and worse, people will wonder what this 9875 values means.
> 
> It means nothing, and is unnecessary.
> 
> 
> 
> The ntdll definition is this:
> 
> 
> 
> // Make _fltused static in ntdll.lib instead of an import.
> 
> extern const char _fltused = 0;
> 
> 
> 
> I was chosing readonly over writable.
> 
> 
> 
> Windows has a variety of definitions, 0, 1, 9875, 9876.
> 
> The non-zero values are just random.

I am a little concerned about changing it from an int to a char.

If my memory is correct, the error message when it's not defined was "unresolved external int_fltused"

Do you still have the build error you were getting to verify?

jaykrell:
I was just reading history and source a little.

1. The type does not matter. There are no types at this level of the system.
i.e. linking.
The name is "_fltused", not "int_fltused".
This isn't C++. There's no typesafe linkage.
If it was type-sensitive, like in C++, the name would be something almost unstatable.
I.e. it'd be "?_fltused@@3HA"

I say almost unstatable, because there is the __identifier feature.
You can give any string to a symbol, really.

2. Observe what I showed, it is char in upstream Windows (with many values, 0, 1, 9875, 9876, etc.).

I made it that way years ago.
It was really much worse upstream.
Upstream exports it from ntdll.dll.
Which is kinda dumb.
 Someone made dynamic linking pay to resolve the symbol, not just static linking.
I changed it to be static in ntdll.lib. (libs can be a mix of static and import).
So "future" builds don't import it (future since years ago).
But it has to be kept exported for compatibility, perhaps.

3. Observe:
```
C:\s>type flt.c
float f1(float a)
{
 return a;
}

#if FIX
char _fltused;
#endif

C:\s>cl /LD flt.c /link /nod /noentry
Microsoft (R) C/C++ Optimizing Compiler Version 19.43.34810 for x64

flt.obj : error LNK2001: unresolved external symbol _fltused
flt.dll : fatal error LNK1120: 1 unresolved externals

C:\s>cl /DFIX /LD flt.c /link /nod /noentry

=> success
```
Then interestingly, you might wonder if LTCG brings type information
and strictness, as it is known to sometimes. But in this case it optimizes
away the reference, even if you don't request optimization.


I mean, the size and value are user-visible.
_Everything_ is user-visible.
I could assert or print it.
This is a breaking change, by an _extreme_ stretch of the imagination.
I've never seen anyone actually use the variable, just that the compiler references it and the linker must resolve it.
You could make it a function even, adding executable byte(s).

We could imagine putting:
`#if defined (_MSC_VER) && !defined(__clang__)`

around it..if we require/assume coherent builds. But then, such builds won't reference it anyway.

So much about so little, sorry.

MdePkg/Library/FltUsedLib/FltUsedLib.c
```diff
@@ -8,4 +8,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 
 // create a global to satisfy the compilers insertion of the _fltused
 // in reponse to detecting floating point type operations.
-int  _fltused = 0x9875;
```
makubacki, line 11: If the idea is to clean up the definition for simplicity, the codebase still defines `_fltused` in CryptoPkg/Library/IntrinsicLib/MemoryIntrinsics.c:

https://github.com/microsoft/mu_basecore/blob/f8278f79e839b51845e4f00013a6338bbc67c1a7/CryptoPkg/Library/IntrinsicLib/MemoryIntrinsics.c#L22-L24

That is assigned `1`, which continues the "randomly assigned value" pattern, but it is also of type `int`.

If this change modifies the type from `int` to `char` here, I think the differing types for the same symbol in the same codebase creates its own dose of confusion.

Your argument indicates the type does not matter. If that's the case, these should at least have the same type applying the argument for the type chosen consistently.

jaykrell: Agreed, they should be the same. Ideally remove the crypto one, to avoid risk of duplicate external error (which making them the same doesn't ward off).

makubacki:
Closing since MdePkg/Library/FltUsedLib is removed in https://github.com/microsoft/mu_basecore/pull/1736 to move to mu_plus

PR #1712, opened 2026-03-20
[SQUASH ON REBASE] Clang/Gcc: Move EnableInterruptsAndSleep out of EnableInterrupts.nasm.

jaykrell:
## Description

EnableInterrupts.nasm does not build for clang/gcc, so they cannot link to EnableInterruptsAndSleep.

Squash with 4969b55336aa45859a8f594fca5c175506bc81a8.

- [ ] Impacts functionality? No.
- [ ] Impacts security? No.
- [ ] Breaking change? No.
- [ ] Includes tests? No.
- [ ] Includes documentation? No.

## How This Was Tested

No. But the opposite. I cannot build HyperV-UEFI with Clang or Gcc because of this.

## Integration Instructions

Easy.

### DIFF
--- a/MdePkg/Library/BaseLib/X64/EnableInterruptsAndSleep.nasm
+++ b/MdePkg/Library/BaseLib/X64/EnableInterruptsAndSleep.nasm
@@ -1,15 +1,17 @@
+; MU_CHANGE - START
 ;------------------------------------------------------------------------------
 ;
 ; Copyright (c) 2006, Intel Corporation. All rights reserved.<BR>
+; Copyright (c) Microsoft Corporation.
 ; SPDX-License-Identifier: BSD-2-Clause-Patent
 ;
 ; Module Name:
 ;
-;   EnableInterrupts.Asm
+;   EnableInterruptsAndSleep.nasm
 ;
 ; Abstract:
 ;
-;   EnableInterrupts function
+;   EnableInterruptsAndSleep function
 ;
 ; Notes:
 ;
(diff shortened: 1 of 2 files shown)


os-d:
What is the build error you hit? This is built and used for clang/gcc virtual platforms in edk2

jaykrell:
Link failed. Unresolved external. For 64bit it is only in source file that is only built for MSFT.
Do you build .nasm files for gcc/clang even if marked MSFT?
Where is the use? How could it link?

os-d:
> Link failed. Unresolved external. For 64bit it is only in source file that is only built for MSFT. Do you build .nasm files for gcc/clang even if marked MSFT? Where is the use? How could it link?

Oh, I see, sorry, I was looking on my phone and missed it was MSFT only. I thought your statement of doesn’t build was that there was a build error not that the file isn’t included. This makes sense.

os-d (approved):

MdePkg/Library/BaseLib/X64/EnableInterruptsAndSleep.nasm
```diff
@@ -0,0 +1,34 @@
+; MU_CHANGE - START
+;------------------------------------------------------------------------------
+;
+; Copyright (c) 2006, Intel Corporation. All rights reserved.<BR>
```
apop5, line 4: Nit: this may be based on an intel copyright file, but there are Microsoft changes that should be reflected in adding a Microsoft copyright statement.

apop5 (approved):

MdePkg/Library/BaseLib/X64/EnableInterruptsAndSleep.nasm
jaykrell: This content was in a file, with only this mark.
Is that a mistake and the other file was missing a mark?
I don't know the math/science here.
Do we need both marks or just Microsoft?

jaykrell: I added. I realize that is what you said so hopefully it was deliberate and precise, or if not, I'll accept more nit picking.

apop5 (approved):

makubacki (approved):

cfernald (approved):

PR #1720, opened 2026-03-26
[CHERRY-PICK] Revert mu changes of variable driver in favor of edk2 commits

kuqin12:
## Description

EDK2 has merged the change originally checked in by Liqi. This change reverts his original commit and cherry-pick the edk2 commit.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on physical hardware.

## Integration Instructions

N/A

### DIFF
--- a/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
+++ b/MdeModulePkg/Universal/Variable/RuntimeDxe/VariableSmmRuntimeDxe.c
@@ -2025,13 +2025,8 @@ SmmVariableReady (
   //
   Status = GetVariablePayloadSize (&mVariableBufferPayloadSize);
   ASSERT_EFI_ERROR (Status);
-  if (mMmCommunication3 != NULL) {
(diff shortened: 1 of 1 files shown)


apop5 (approved):

os-d (approved):

makubacki (approved):

kuqin12:
well, seem that this change is too clean to be "Analyze"d?

kuqin12:
hmm. i think we close this one, and just handle it during the next integration. @apop5 thoughts?

apop5:
> hmm. i think we close this one, and just handle it during the next integration. @apop5 thoughts?

I think that is fine.

This would be a clean drop since the edk2 PR was merged without any other changes.

This is something we need to go back and revisit. why is the codeql workflow not triggering when the total delta of files is zero.

kuqin12:
> This would be a clean drop since the edk2 PR was merged without any other changes.
> 
Sounds good. Will drop this PR.
> This is something we need to go back and revisit. why is the codeql workflow not triggering when the total delta of files is zero.
>
Yeah. Although an empty PR is also odd. We can fix that when it is really needed.

PR #1722, opened 2026-03-26
NetworkPkg/Dhcp6Dxe: Prevent potential infinite hang in EfiDhcp6Stop()

makubacki:
## Description

When a USB network adapter is removed while a DHCPv6 lease is active, the USB stack calls DisconnectController() at TPL_CALLBACK. The network stack teardown eventually invokes EfiDhcp6Stop(), which sends a Release message and enters a while loop polling Udp6->Poll() for the server's Reply:

  while (Instance->UdpSts == EFI_ALREADY_STARTED) {
    Udp6->Poll (Udp6);
  }

This loop never terminates because:

1. The NIC hardware is physically gone
2. The timeout mechanism relies on Dhcp6OnTimerTick(), a timer event registered at TPL_CALLBACK. Since EfiDhcp6Stop() is already running at TPL_CALLBACK (called from DisconnectController()), the timer event does not fire. So, UdpSts is never updated, and the loop never exits.

Note: Dhcp6OnTimerTick() is not in the packet receive path.

Udp6->Poll() drives packet reception synchronously through the network stack (MNP -> IP6 -> UDP6 -> Dhcp6ReceivePacket()), and Dhcp6HandleReplyMsg() sets Instance->UdpSts = EFI_SUCCESS when a Release Reply arrives.

The timer event's role for Release messages is retransmission and to reach timeout. Which are blocked because of the TPL in this case.

This change removes the synchronous polling loop from EfiDhcp6Stop().

This matches how Dhcp4Dxe handles EfiDhcp4Stop() and EfiDhcp4Release(), which send the Release and immediately clean up without polling.

This is considered acceptable because:

- Dhcp6SendReleaseMsg() calls Dhcp6TransmitPacket(), which synchronously hands the packet to hardware by calling SNP->Transmit() before returning.

- The server acts on the Release immediately upon receipt per RFC 8415 Section 18.3.7. It does not wait for client acknowledgment.

- RFC 8415 Section 18.2.7 states:

  "implementations SHOULD retransmit one or more times, but MAY choose to terminate the retransmission procedure early"

- The Release Reply carries no information we (the client) need.

- Dhcp6CleanupSession(), called immediately after sending Release, invokes Dhcp6CleanupRetry(DHCP6_PACKET_STATEFUL) which iterates TxList and frees the Release TxCb and its packet. The same memory cleanup occurs whether or not a Reply was received.

---

For reference: [RFC 8415 Section 18.2.7](https://datatracker.ietf.org/doc/html/rfc8415#section-18.2.7)

---

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

- PR is in draft while platform testing occurs

## Integration Instructions

- N/A

### DIFF
--- a/NetworkPkg/Dhcp6Dxe/Dhcp6Impl.c
+++ b/NetworkPkg/Dhcp6Dxe/Dhcp6Impl.c
@@ -174,20 +174,26 @@ EfiDhcp6Stop (
   IN EFI_DHCP6_PROTOCOL  *This
   )
 {
-  EFI_TPL            OldTpl;
-  EFI_STATUS         Status;
-  EFI_UDP6_PROTOCOL  *Udp6;
-  DHCP6_INSTANCE     *Instance;
-  DHCP6_SERVICE      *Service;
+  // MU_CHANGE [BEGIN] - Prevent potential infinite loop in Dhcp6Stop()
+  EFI_TPL         OldTpl;
+  EFI_STATUS      Status;
+  DHCP6_INSTANCE  *Instance;
+
+  // EFI_UDP6_PROTOCOL  *Udp6;
+  // DHCP6_SERVICE   *Service;
+
+  // MU_CHANGE [END] - Prevent potential infinite loop in Dhcp6Stop()
 
   if (This == NULL) {
     return EFI_INVALID_PARAMETER;
   }
 
   Instance = DHCP6_INSTANCE_FROM_THIS (This);
-  Service  = Instance->Service;
-  Udp6     = Service->UdpIo->Protocol.Udp6;
-  Status   = EFI_SUCCESS;
+  // MU_CHANGE [BEGIN] - Prevent potential infinite loop in Dhcp6Stop()
+  // Service  = Instance->Service;
+  // Udp6     = Service->UdpIo->Protocol.Udp6;
+  // MU_CHANGE [END] - Prevent potential infinite loop in Dhcp6Stop()
+  Status = EFI_SUCCESS;
 
   //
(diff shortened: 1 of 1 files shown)


NetworkPkg/Dhcp6Dxe/Dhcp6Impl.c
```diff
@@ -241,7 +265,7 @@ EfiDhcp6Stop (
   Dhcp6CleanupSession (Instance, EFI_SUCCESS);
   gBS->RestoreTPL (OldTpl);
 
-  return Status;
+  return EFI_SUCCESS; // MU_CHANGE - Prevent potential infinite loop in Dhcp6Stop()
```
joschock: Should we consider still returning `Status` here in the event that e.g. `Dhcp6SendReleaseMsg` fails? Or should we enforce that `EfiDhcp6Stop` always succeeds no matter what?

makubacki: I don't have a strong opinion. I found 10 callers and they do not use the return value at all. To make the return value more useful, I updated it to `Status`.

makubacki: I'll close the thread based on this change. Feel free to reopen if needed.

os-d (approved):

apop5 (approved):

kuqin12 (approved):

joschock (approved):

PR #1725, opened 2026-03-28
Add Uefi Shell test app for the BootManagerPolicy Protocol

spbrogan:
## Description

Add Uefi Shell unit test for the BootManagerPolicy Protocol

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Built and run against QemuQ35 x64

## Integration Instructions

Add to the platform DSC with other UEFI shell based tests

### DIFF
--- a/MdePkg/Test/UnitTest/BootManagerPolicyProtocol/BootManagerPolicyProtocolTest.c
+++ b/MdePkg/Test/UnitTest/BootManagerPolicyProtocol/BootManagerPolicyProtocolTest.c
@@ -169,16 +169,16 @@ TestConnectDeviceClassConsole (
   EFI_STATUS              Status;
   BM_POLICY_TEST_CONTEXT  *Ctx;
 
-  Ctx    = (BM_POLICY_TEST_CONTEXT *)Context;
+  Ctx = (BM_POLICY_TEST_CONTEXT *)Context;
 
-  //logging statement so if system uefi log needs to be evaluated it is clear which operation was requested  
+  // logging statement so if system uefi log needs to be evaluated it is clear which operation was requested
   DEBUG ((DEBUG_INFO, "[%a] - Connect DeviceClass(Console) - Start\n", UNIT_TEST_APP_NAME));
 
   Status = Ctx->Protocol->ConnectDeviceClass (
                             Ctx->Protocol,
                             &gEfiBootManagerPolicyConsoleGuid
                             );
-  DEBUG ((DEBUG_INFO, "[%a] - Connect DeviceClass(Console) - End\n", UNIT_TEST_APP_NAME)); //log completed operation for clarity in logs
+  DEBUG ((DEBUG_INFO, "[%a] - Connect DeviceClass(Console) - End\n", UNIT_TEST_APP_NAME)); // log completed operation for clarity in logs
 
   UT_ASSERT_TRUE (IsValidConnectDeviceClassStatus (Status));
   UT_LOG_INFO ("ConnectDeviceClass(Console) returned %r\n", Status);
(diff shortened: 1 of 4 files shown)


MdePkg/Test/UnitTest/BootManagerPolicyProtocol/BootManagerPolicyProtocolTest.c
```diff
@@ -0,0 +1,583 @@
...
+{
+  EFI_STATUS              Status;
+  BM_POLICY_TEST_CONTEXT  *Ctx;
+
+  Ctx    = (BM_POLICY_TEST_CONTEXT *)Context;
+
+  //logging statement so if system uefi log needs to be evaluated it is clear which operation was requested  
+  DEBUG ((DEBUG_INFO, "[%a] - Connect DeviceClass(Console) - Start\n", UNIT_TEST_APP_NAME));
+
+  Status = Ctx->Protocol->ConnectDeviceClass (
```
apop5, line 177: This is exercising the protocol, but it is not verifying that there is any change in the system. Would something as simple as locating a handle buffer of all handles before/after and seeing if there is a difference be useful information? Something like "3 new handles exist after connecting"

Is would be testing system implementation, which is outside the scope of the protocol, but it would give confidence that the implementation of the protocol produces a change in system behavior.

MdePkg/Test/UnitTest/BootManagerPolicyProtocol/BootManagerPolicyProtocolTest.c
```diff
@@ -0,0 +1,583 @@
...
+  AddTestCase (
+    KnownGuidSuite,
+    "ConnectDeviceClass with Network GUID",
+    "BootManagerPolicy.ConnectDeviceClass.Network",
+    TestConnectDeviceClassNetwork,
+    LocateProtocolPreReq,
+    NULL,
+    &mTestContext
+    );
+  AddTestCase (
```
apop5, line 487: Connect all needs to be the last test. Running it before the other test cases will impact results. 

Ie connect all will connect all the device classes, resulting in different system behavior for connect storage class

os-d:
Take this to edk2?

MdePkg/Test/UnitTest/BootManagerPolicyProtocol/BootManagerPolicyProtocolTest.c
os-d: I think that would certainly be a better test, the current level of tests here are SCT level, just correctness of return status. But can we be sure ConnectAll wasn't called before this point? I don't think we can, right?

apop5: Agreed, we cannot make any assumptions about what the platform has previously connected.

I was looking at it as useful information coming out of the test. I wouldn't suggest making it fail the test if no new devices were connected.

spbrogan: The goal was basically SCT level test.  I don't love that we can't verify more but it is a starting point.  I don't think change in state is something I know programmatically since it is heavily dependent on starting state. 

I suppose we could call disconnect controller on something and then confirm that it is connected after calling the test.  That too seems like it will be fragile and error prone.  

I did add the debug comments so you could manually correlate the UEFI logs with the test case and see what happened.

spbrogan: I read this thread and I don't see anything actionable.  Will resolve once this is the last thing left.

MdePkg/Test/UnitTest/BootManagerPolicyProtocol/BootManagerPolicyProtocolTest.c
spbrogan: Agree.  moved to last test in suite and added comment.

apop5 (approved):

os-d (approved):

PR #1735, opened 2026-04-01
UefiCpuPkg: ArmMmuLib: Check if Block Split Following Page Alloc

os-d:
## Description

Currently, there is a bug in UpdateRegionMappingRecursive() when guard pages are enabled and a large page is being split.

The code checks whether the page table is a block or table and seeing that it is a block, allocates a new page table for the next level. However, when it does this, it will call an additional recursive call into the page table updating logic to make sure the new page table page is mapped. In addition, when guard pages are enabled, it will mark the guard page as RP. If the guard page is in the same block as we are already trying to split, the recursive call will split the block and mark the guard page as RP.

When we return to the original call, it will fill out the now orphaned page table but never install it into the page table hierarchy (and if it did, it would lose the guard page). This has been observed to cause a driver's code section to still have NX set on it and so crash when trying to execute.

This commit resolves the issue by checking if the block has already been split when we return from the new page table allocation. If it has, we simply update the existing table mapping instead of trying to split the block.

The allocated page table page cannot be immediately freed, because this might trigger the block to get re-merged, so a reference to it is held until the end of updating this level and subsequent levels, when it can be safely freed.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a physical ARM64 platform with NX and page guard enabled. System crashes before this change and boots with it.

## Integration Instructions

N/A.

### DIFF
--- a/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
+++ b/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
@@ -277,11 +277,14 @@ UpdateRegionMappingRecursive (
   VOID        *TranslationTable;
   EFI_STATUS  Status;
   BOOLEAN     NextTableIsLive;
+  VOID        *TableToFree; // MU_CHANGE: Fix Split During Alloc
 
   ASSERT (((RegionStart | RegionEnd) & EFI_PAGE_MASK) == 0);
 
   BlockShift = (Level + 1) * BITS_PER_LEVEL + MIN_T0SZ;
   BlockMask  = MAX_UINT64 >> BlockShift;
+
+  TableToFree = NULL; // MU_CHANGE: Fix Split During Alloc
 
   DEBUG ((
     DEBUG_VERBOSE,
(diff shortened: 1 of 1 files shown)


os-d:
I'm taking this to edk2 in parallel (which yes, just realized I forgot MU_CHANGE tags), but this breaking platforms currently, so also bringing into Mu.

cfernald (approved):

kuqin12 (approved):

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
```diff
@@ -432,6 +451,16 @@ UpdateRegionMappingRecursive (
     }
   }
 
+  // MU_CHANGE BEGIN: Fix Split During Alloc
+  // We may have left an orphaned page table page if we discovered a
+  // recursive call already split a block.
+  //
+  if (TableToFree != NULL) {
+    FreePages (TableToFree, 1);
+    TableToFree = NULL;
```
makubacki, line 461: Looks like dead code.

```suggestion
```

makubacki (approved):

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
cfernald: FWIW, I like this practice. It prevents use-after-free and the compiler should be able to easily optimize out any actual generated code.

makubacki: I like the practice too. It's just that in this case it is a local variable assignment that doesn't serve an effective purpose. If no tools will complain about it, then I don't have a strong preference on whether it stays or not.

os-d: Yeah, this is just paranoia because this is a recursive function (even though this is the end of the recursion) and against future changes to the function. It can be dropped, I don't have a big preference.

antklein (approved):

UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibCore.c
cfernald: I'd leave it. It's harmless and makes the code refactor-resistant.

cfernald (approved):

PR #1743, opened 2026-04-04
Fix numerous nvme issues - see each commit

### LINKED ISSUE #1738: [Bug]: NvmeDxe::NvmExpressMediaClear: off-by-one skips last LBA

- [x] I have searched existing issues

### Current Behavior

NvmExpressMediaClear: off-by-one skips last LBA

Why this is a real bug:

Media->LastBlock is the 0-based index of the last valid LBA. The < operator excludes it. The loop writes sectors 0 through LastBlock - 1, missing the final sector.

Consequence:
The final sector of the namespace is not overwritten during a media clear. For security-sensitive sanitization, this leaves residual data on media.

### Expected Behavior

Last block should be cleared

### Steps To Reproduce

Trigger path:

Namespace has Nsze = 100 → Media->LastBlock = 99.
Loop: SectorOffset from 0 to 98 → sector 99 is never written.
Why this is NOT a false positive:
The NVMe spec and EFI_BLOCK_IO_MEDIA specification define LastBlock as the LBA of the last block (inclusive). Correct iteration would use <=.

### Build Environment

```markdown
NA
```

### Version Information

```text
all
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1739: [Bug]: Nvme:: NvmExpressMediaClear: Incorrect buffersize

- [x] I have searched existing issues

### Current Behavior

NvmExpressMediaClear: BufferSize passed as 1 instead of block size → function is completely non-functional

Why this is a real bug:
WriteBlocks expects BufferSize in bytes. The call passes 1:

NvmeBlockIoWriteBlocks checks (BufferSize % BlockSize) != 0. Since 1 % 512 != 0 (for any standard block size ≥ 2), this always returns EFI_BAD_BUFFER_SIZE.

Consequence:
The NIST 800-88r1 media clear operation is entirely non-functional. No overwrite data is written to any sector. A caller relying on this for data sanitization receives no actual sanitization.

Minimal fix direction:
Change 1 to Device->Media.BlockSize.

### Expected Behavior

Block is overwritten

### Steps To Reproduce

Trigger path:

Caller invokes NvmExpressMediaClear with valid parameters.
Inner loop calls WriteBlocks with BufferSize = 1.
NvmeBlockIoWriteBlocks returns EFI_BAD_BUFFER_SIZE.
Loop continues (no error check), every iteration fails.
Function returns last error status. No data is ever written.
Why this is NOT a false positive:
The comment says "Write one sector per write" — the intent was BlockSize bytes. The EFI_BLOCK_IO_PROTOCOL specification defines BufferSize as bytes, not sectors. There is no code path where BufferSize=1 would pass the block-alignment check.

### Build Environment

```markdown
NA
```

### Version Information

```text
All
```

### Urgency

Medium

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

### LINKED ISSUE #1744: [Bug]: NvmeDxe:: Incorrect cleanup on failure case resulting in memory leak and protocol tracking

- [x] I have searched existing issues

### Current Behavior

NvmExpressDriverBindingStart: early return leaks Private and protocol opens

Why this is a real bug:
After allocating [Private](vscode-file://vscode-app/c:/Users/sebrogan/AppData/Local/Programs/Microsoft%20VS%20Code/07ff9d6178/resources/app/out/vs/code/electron-browser/workbench/workbench.html) via AllocateZeroPool (line ~1201) and successfully opening both gEfiDevicePathProtocolGuid and gEfiPciIoProtocolGuid BY_DRIVER, the function attempts PciIo->Attributes(EfiPciIoAttributeOperationGet, ...). If this call fails, the code executes return Status instead of goto Exit.

Why this is NOT a false positive:
The Exit: label at line ~1340 performs FreePool(Private), gBS->CloseProtocol(PciIo), and gBS->CloseProtocol(DevicePath). The direct return skips all of these. No other caller cleans up Private or closes the protocols.

Consequence:
Memory leak of NVME_CONTROLLER_PRIVATE_DATA. Two protocols remain opened BY_DRIVER on the controller handle, preventing other drivers from binding.

### Expected Behavior

Memory is freed
Protocols are closed

### Steps To Reproduce

NvmExpressDriverBindingStart is called.
OpenProtocol for DevicePath succeeds (opens BY_DRIVER).
OpenProtocol for PciIo succeeds (opens BY_DRIVER).
AllocateZeroPool for Private succeeds.
PciIo->Attributes(Get) returns an error.
return Status bypasses the Exit: label.

### Build Environment

```markdown
NA
```

### Version Information

```text
all
```

### Urgency

Low

I will fix it

No maintainer feedback needed

### Anything else?

_No response_

spbrogan:
## Description

Fix numerous nvme issues - see each commit.

fixes #1739, #1738, #1744

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Host based unit tests authored to show failure.
Host based unit tests passed after changes

## Integration Instructions

None.  Existing module.

### DIFF
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
@@ -366,7 +366,7 @@ NvmExpressMediaClear (
   // Per NIST 800-88r1, one or more pass of writes may be alteratively used.
   //
   for (TotalPassCount = 0; TotalPassCount < PassCount; TotalPassCount++) {
-    for (SectorOffset = 0; SectorOffset < Media->LastBlock; SectorOffset++ ) {
+    for (SectorOffset = 0; SectorOffset <= Media->LastBlock; SectorOffset++ ) {
       Status = Device->BlockIo.WriteBlocks (
                                  &Device->BlockIo,
                                  MediaId,
--- a/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
+++ b/MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
@@ -352,13 +352,17 @@ NvmExpressMediaClear (
   }
 
   //
-  // If an invalid buffer or buffer size is sent, the Media Clear operation
-  // cannot be performed as it requires a native WRITE command. The overwrite
-  // buffer must have granularity of a namespace block size.
+  // If an invalid buffer is sent, the Media Clear operation
+  // cannot be performed as it requires a native WRITE command.
   //
   if (SectorOwBuffer == NULL) {
     return EFI_INVALID_PARAMETER;
   }
+
+  //
+  // Since the parameters do not include OwBuffer Size, the size of the overwrite buffer is assumed to be equal
+  // to the block size of the media.
+  //
 
   Status = EFI_SUCCESS;
 
@@ -370,9 +374,9 @@ NvmExpressMediaClear (
       Status = Device->BlockIo.WriteBlocks (
                                  &Device->BlockIo,
                                  MediaId,
-                                 SectorOffset,  // Sector/LBA offset (increment each pass)
-                                 1,             // Write one sector per write
-                                 SectorOwBuffer // overwrite buffer
+                                 SectorOffset,                   // Sector/LBA offset (increment each pass)
+                                 (UINTN)Device->Media.BlockSize, // Write one block/sector at a time with overwrite buffer
+                                 SectorOwBuffer                  // overwrite buffer
                                  );
     }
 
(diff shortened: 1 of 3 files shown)


os-d:
Since this is not causing an active problem, can you take this directly to edk2? We can still review it there.

MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressMediaSanitize.c
```diff
@@ -352,27 +352,31 @@ NvmExpressMediaClear (
...
+  //
+
   Status = EFI_SUCCESS;
 
   //
   // Per NIST 800-88r1, one or more pass of writes may be alteratively used.
   //
   for (TotalPassCount = 0; TotalPassCount < PassCount; TotalPassCount++) {
-    for (SectorOffset = 0; SectorOffset < Media->LastBlock; SectorOffset++ ) {
+    for (SectorOffset = 0; SectorOffset <= Media->LastBlock; SectorOffset++ ) {
```
apop5, line 373: I think lastlba is zero based, not one based.

This should attempt to write past the last lba of the disk.

apop5: Double checked, your implementation is correct. Lastlba should be a valid lba.

apop5 (approved):

PR #1751, opened 2026-04-07
Support reblock

kuqin12:
## Description

Current method of migrating MM accessible buffers is leaving the previously unblocked buffers unblocked, which will pose the risk of failing DRTM validation.

This change adds a new interface to support reblock pages to mitigate the issue.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested using QEMU Q35 and booted to UEFI shell.

## Integration Instructions

Call `MmReblockMemoryRequest` after the old buffer is being replaced.

### DIFF
--- a/MdeModulePkg/Universal/Variable/Pei/Variable.c
+++ b/MdeModulePkg/Universal/Variable/Pei/Variable.c
@@ -1479,13 +1479,14 @@ BuildVariableRuntimeCacheInfoHob (
     //
     Buffer = AllocateRuntimePages (Pages);
     ASSERT (Buffer != NULL);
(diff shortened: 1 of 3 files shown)


apop5:
Would it be possible, at end of dxe/ready to bot, to have mm produce a debug list of unblocked regions? This would allow auditing of what is unblocked, and maybe enable tracking back any other regions that might need reblocked?

kuqin12:
Not an issue anymore. Will not introduce new changes to support a non-issue.

PR #1752, opened 2026-04-07
SecurityPkg: Remove Globals from HashLibTpm2PeilessSec

Raymond-MS:
## Description

Updated the HashLibTpm2PeilessSec library to no longer use globals in SEC. When enabled on SBSA, this was causing a hang as functions attempted to set RO memory. Changed the code to query the relevant information each time it was needed either through the TPM or Transfer List. Added new helper functions to Tpm2Help which removes the need to have a duplicate hash struct.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Built with the PeilessSecMeasureLib not set to the NULL instance on QEMU SBSA with the TPM enabled. Verified the library measure the FV and CRTM version.

## Integration Instructions

It is better for us to continue using the NULL instance of the PeilessSecMeasureLib as the time it took to measure the FV was ~15 mins on QEMU SBSA with TPM support.

### DIFF
--- a/SecurityPkg/Include/Library/Tpm2HelpLib.h
+++ b/SecurityPkg/Include/Library/Tpm2HelpLib.h
@@ -39,6 +39,7 @@ GetHashMaskFromAlgo (
   );
 
 // MU_CHANGE - [BEGIN]
+
 /**
   Get hash algorithm from mask.
 
@@ -73,8 +74,9 @@ GetHashInfoSize (
 UINT32
 EFIAPI
 GetHashMaskAtIndex (
-  IN UINT32 Index
+  IN UINT32  Index
   );
+
 // MU_CHANGE - [END]
 
 /**
--- a/SecurityPkg/Include/Library/Tpm2HelpLib.h
+++ b/SecurityPkg/Include/Library/Tpm2HelpLib.h
@@ -38,6 +38,45 @@ GetHashMaskFromAlgo (
   IN TPMI_ALG_HASH  HashAlgo
   );
 
+// MU_CHANGE - [BEGIN]
+/**
+  Get hash algorithm from mask.
+
+  @param[in] HashMask   Hash mask
+
+  @return Hash algorithm
+**/
+TPMI_ALG_HASH
+EFIAPI
+GetHashAlgoFromMask (
+  IN UINT32  HashMask
+  );
+
+/**
+  Get internal hash info size.
+
+  @return Hash info size
+**/
+UINTN
+EFIAPI
+GetHashInfoSize (
+  VOID
+  );
+
+/**
+  Get hash mask at specified index.
+
+  @param[in] Index   Index requested
+
+  @return Hash mask at the specified index
+**/
+UINT32
+EFIAPI
+GetHashMaskAtIndex (
+  IN UINT32 Index
+  );
+// MU_CHANGE - [END]
+
 /**
   Copy AuthSessionIn to TPM2 command buffer.
 
--- a/SecurityPkg/Include/Library/Tpm2HelpLib.h
+++ b/SecurityPkg/Include/Library/Tpm2HelpLib.h
@@ -1,4 +1,5 @@
 /** @file
+  MU_CHANGE
   Definitions for TPM2 helper functions
 
 Copyright (c), Microsoft Corporation.
@@ -38,8 +39,6 @@ GetHashMaskFromAlgo (
   IN TPMI_ALG_HASH  HashAlgo
   );
 
-// MU_CHANGE - [BEGIN]
-
 /**
   Get hash algorithm from mask.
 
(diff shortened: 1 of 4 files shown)


SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
```diff
@@ -83,6 +93,117 @@ GetTransferList (
...
+  TcgEfiSpecIdEventStruct = (TCG_EfiSpecIDEventStruct *)
+                            (EventLog + OFFSET_OF (TCG_PCR_EVENT, Event));
+
+  CopyMem (&NumberOfAlgorithms, TcgEfiSpecIdEventStruct + 1, sizeof (NumberOfAlgorithms));
+  DigestSize = (TCG_EfiSpecIdEventAlgorithmSize *)((UINT8 *)TcgEfiSpecIdEventStruct + sizeof (*TcgEfiSpecIdEventStruct) + sizeof (NumberOfAlgorithms));
+  DEBUG ((DEBUG_INFO, "%a: Transfer list TPM event log available\n", __func__));
+
+  // Update the supported hash bitmap based on the info from the TCG event log
+  for (Idx = 0; Idx < NumberOfAlgorithms; Idx++) {
+    *SupportedHashBitmap |= GetHashMaskFromAlgo (DigestSize[Idx].algorithmId);
```
cfernald, line 177: Should we be checking for a 0 return from GetHashMaskFromAlgo ?

SecurityPkg/Library/Tpm2HelpLib/Tpm2Help.c
```diff
@@ -104,6 +104,64 @@ GetHashMaskFromAlgo (
...
+{
+  UINTN  Index;
+
+  for (Index = 0; Index < sizeof (mHashInfo)/sizeof (mHashInfo[0]); Index++) {
+    if (mHashInfo[Index].HashMask == HashMask) {
+      return mHashInfo[Index].HashAlgo;
+    }
+  }
+
+  return 0;
```
cfernald: Elsewhere you are using `TPM_ALG_ERROR`

cfernald (approved):

SecurityPkg/Include/Library/Tpm2HelpLib.h
```diff
@@ -38,6 +38,47 @@ GetHashMaskFromAlgo (
   IN TPMI_ALG_HASH  HashAlgo
   );
 
+// MU_CHANGE - [BEGIN]
```
makubacki: Given this entire file is specific to Mu, you can just leave a MU_CHANGE tag at the top of the file. This is confusing as-is, because it makes it appear as if only these lines differ from an upstream file.

SecurityPkg/Include/Library/Tpm2HelpLib.h
```diff
@@ -38,6 +38,47 @@ GetHashMaskFromAlgo (
...
+/**
+  Get hash algorithm from mask.
+
+  @param[in] HashMask   Hash mask
+
+  @return Hash algorithm
+**/
+TPMI_ALG_HASH
+EFIAPI
+GetHashAlgoFromMask (
```
makubacki: If you're prefixing function names with "Tpm2" in https://github.com/tianocore/edk2/pull/11634, can you go ahead and do that here? It will be one less breaking change later.

SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
```diff
@@ -1,6 +1,11 @@
 /** @file
+  // MU_CHANGE
+  This library is the PeilessSec version of the HashLib. It will
+  initate a hash on each supported hash algorithm via the TPM or
```
makubacki:
```suggestion
  initiate a hash on each supported hash algorithm via the TPM or
```

SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
```diff
@@ -83,6 +93,117 @@ GetTransferList (
...
+    *SupportedHashBitmap = TpmHashBitmap;
+    DEBUG ((DEBUG_INFO, "%a: No Transfer List or TPM event log available\n", __func__));
+  }
+
+  *SupportedHashBitmap &= PcrHashBitmap;
+  if (*SupportedHashBitmap == 0x00) {
+    DEBUG ((DEBUG_ERROR, "%a: No supported Hash algorithm with event log Spec...!\n", __func__));
+  }
+
+  DEBUG ((DEBUG_INFO, "%a - Exit\n", __func__));
```
makubacki: We usually try to avoid messages like this, especially at `DEBUG_INFO` level.

SecurityPkg/Library/HashLibTpm2/HashLibTpm2PeilessSecLib.c
Raymond-MS: No, I don't think it is necessary as you would just OR in 0 which does nothing.

makubacki (approved):

PR #1763, opened 2026-04-13
[SQUASH ON REBASE] [REBASE & FF] ArmPlatformPkg: MemoryInitPeiLib: Update Resc Desc V2 HOB

os-d:
## Description

This PR has two commits:

Update Resc Desc V2 HOB
--
MemoryInitPeiLib attempts to search for a Resc Desc V2 HOB that describes system memory and if needed split it up and create a resc desc HOB v2 for the FD region that is not overlapping.

Currently, the change to produce a resc desc HOB v2 for system memory if not found does not correctly update it to split the HOB and will create an overlapping FD HOB.

The patina readiness tool catches this. This commit fixes this by searching through resc desc HOB v2s to find the system memory HOB v2 that covers system memory to fix up.

This should be squashed with b8960d8a908c98b566cb5204cc9a6eb8b3d1b11a and fb5e75cbf1c471de9a14233044c5e31ee3028a38 on rebase.

Loosen HOB Check
--
Currently, MemoryInitPeiLib checks if a resc desc HOB partially or wholly covers the system memory range defined by PcdSystemMemoryBase and PcdSystemMemorySize. If a HOB is not found, then the lib produces the system memory resc desc HOB.

However, it does not take into account that platforms have a small range defined in the PCDs and at runtime query the system memory range and produce a larger HOB. The current logic will not discover the larger HOB and will produce a HOB that conflicts with this one. Patina will mark this as an error; edk2 will silently choose one, but it may choose the smaller one.

This commit fixes this behavior by not producing a HOB in this library if one already exists that contains PcdSystemMemoryBase.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on a virtual platform where Patina is being integrated.

## Integration Instructions

N/A.

### DIFF
--- a/ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
+++ b/ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
@@ -139,16 +139,16 @@ MemoryPeim (
 
     // Search for System Memory Hob that contains the firmware
     NextHob.Raw = GetHobList ();
-    while ((NextHob.Raw = GetNextHob (EFI_HOB_TYPE_RESOURCE_DESCRIPTOR, NextHob.Raw)) != NULL) {
-      if ((NextHob.ResourceDescriptor->ResourceType == EFI_RESOURCE_SYSTEM_MEMORY) &&
-          (PcdGet64 (PcdFdBaseAddress) >= NextHob.ResourceDescriptor->PhysicalStart) &&
-          (FdTop <= NextHob.ResourceDescriptor->PhysicalStart + NextHob.ResourceDescriptor->ResourceLength))
+    while ((NextHob.Raw = GetNextHob (EFI_HOB_TYPE_RESOURCE_DESCRIPTOR2, NextHob.Raw)) != NULL) {
+      if ((NextHob.ResourceDescriptorV2->V1.ResourceType == EFI_RESOURCE_SYSTEM_MEMORY) &&
+          (PcdGet64 (PcdFdBaseAddress) >= NextHob.ResourceDescriptorV2->V1.PhysicalStart) &&
+          (FdTop <= NextHob.ResourceDescriptorV2->V1.PhysicalStart + NextHob.ResourceDescriptorV2->V1.ResourceLength))
       {
-        ResourceAttributes = NextHob.ResourceDescriptor->ResourceAttribute;
-        ResourceLength     = NextHob.ResourceDescriptor->ResourceLength;
-        ResourceTop        = NextHob.ResourceDescriptor->PhysicalStart + ResourceLength;
+        ResourceAttributes = NextHob.ResourceDescriptorV2->V1.ResourceAttribute;
+        ResourceLength     = NextHob.ResourceDescriptorV2->V1.ResourceLength;
+        ResourceTop        = NextHob.ResourceDescriptorV2->V1.PhysicalStart + ResourceLength;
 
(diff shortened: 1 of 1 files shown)


kuqin12 (approved):

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
makubacki, line 159: Given this is finding V2 resource descriptor HOBs with a memory attributes field, is it intentional to hardcode `EFI_MEMORY_WB`?

makubacki (approved):

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
os-d: This is producing the FD HOB with WB as the attribute, not checking the attribute of the system memory HOB

cfernald (approved):

ArmPlatformPkg/MemoryInitPei/MemoryInitPeiLib.c
makubacki: To clarify, the question is simply that the HOB found is of type `EFI_HOB_TYPE_RESOURCE_DESCRIPTOR2`:

```c
typedef struct {
  ///
  /// The HOB generic header. Header.HobType = EFI_HOB_TYPE_RESOURCE_DESCRIPTOR.
  ///
  EFI_HOB_RESOURCE_DESCRIPTOR    V1;
  ///
  /// The memory attributes (paging and caching) of the resource region.
  ///
  UINT64                         Attributes;
} EFI_HOB_RESOURCE_DESCRIPTOR_V2;
```

So, is the idea that the `Attributes` field is invalid? Not questioning just trying to understand the rationale.

os-d: No, this is weird logic, tbh.

What this function does is say: a resc desc HOB v2 that descs system memory may or may not have been produced already. If it has not been, I'll create one that describes all of system memory.

Now I know there is a resc desc HOB v2 that describes system memory. However, I am responsible for producing the FD HOB, so I need to bifurcate the system memory HOB to account for the space the FD HOB takes up so we don't have overlapping HOBs.

So then it checks if the FD region is on either edge or the middle of system memory and adjusts the base/length of the system memory HOB as needed and produces the FD HOB and a second system memory HOB if the FD region is not aligned to base or end of the system memory HOB.

This same strange "contract" exists in edk2 for resc desc HOB v1s. We just amended it to be v2 only and ignore the v1 system memory HOB and don't produce a v1 FD HOB.

makubacki: That all makes sense, the case that I had in mind conceptually was when a V2 resource descriptor HOB had already been produced with specific memory attributes and the FD falls in the middle and the attributes appear to not be preserved.

Otherwise, I understand that memory attributes didn't exist before and `EFI_MEMORY_WB` is reasonable here.

os-d: Oh I see what you are saying. Yes, that would make sense to preserve when creating the new system memory HOB below. I'll update to preserve the attribute from the existing HOB.

makubacki (approved):

makubacki (approved):

PR #1765, opened 2026-04-14
EmbeddedPkg/VirtualRealTimeClockLib: Fix RTC offset logic for negative TimeZone

capohuang:
## Description

[Problem]
A logic defect in VirtualRealTimeClockLib causes incorrect RTC calculation when Time->TimeZone is negative (e.g., -480 for UTC+8). The comparison EpochSeconds > (Time->TimeZone * SEC_PER_MIN) performs unsigned comparison due to C type promotion rules. This treats the negative offset as a very large unsigned integer, causing the code to skip necessary timezone compensation.
 
[Solution]
Updated the logic to ensure the timezone offset calculation and comparison handle signed values correctly. This allows the library to properly apply compensation for both positive and negative timezones.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
[Steps]
Set the system timezone to UTC+8 (which translates to a -480 offset in the UEFI variable).
Perform a shutdown or hibernate and cold boot.
Observe the RTC time in the UEFI shell and OS.

[Result]
The RTC correctly reflects the local time, and the signed/unsigned comparison issue is resolved.

## Integration Instructions
N/A

### DIFF
--- a/EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c
+++ b/EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c
@@ -55,6 +55,7 @@ LibGetTime (
   UINT64      Remainder;
   UINTN       EpochSeconds;
   UINTN       Size;
+  INT64       AdjustedEpochSeconds;
 
   if (Time == NULL) {
     return EFI_INVALID_PARAMETER;
@@ -150,7 +151,13 @@ LibGetTime (
 
     // Adjust for the correct time zone
     if (Time->TimeZone != EFI_UNSPECIFIED_TIMEZONE) {
-      EpochSeconds += Time->TimeZone * SEC_PER_MIN;
+      AdjustedEpochSeconds  = (INT64)EpochSeconds;
+      AdjustedEpochSeconds += (INT64)Time->TimeZone * (INT64)SEC_PER_MIN;
+      if (AdjustedEpochSeconds < 0) {
+        AdjustedEpochSeconds = 0;
+      }
+
+      EpochSeconds = (UINTN)AdjustedEpochSeconds;
     }
   }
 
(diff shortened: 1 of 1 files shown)


EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c
makubacki, line 1: Can you please fix the same problem in `LibGetTime()` at the beginning of this file?

https://github.com/microsoft/mu_basecore/blob/201039c2a82f458b672e2ee36d472c974723c219/EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c#L151-L154

EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c
```diff
@@ -250,10 +251,14 @@ LibSetTime (
   EpochSeconds = EfiTimeToEpoch (Time);
 
   // Adjust for the correct time zone, i.e. convert to UTC time zone
-  if (  (Time->TimeZone != EFI_UNSPECIFIED_TIMEZONE)
-     && (EpochSeconds > Time->TimeZone * SEC_PER_MIN))
-  {
-    EpochSeconds -= Time->TimeZone * SEC_PER_MIN;
+  if (Time->TimeZone != EFI_UNSPECIFIED_TIMEZONE) {
+    AdjustedEpochSeconds  = (INT64)EpochSeconds;
+    AdjustedEpochSeconds -= (INT64)Time->TimeZone * SEC_PER_MIN;
```
makubacki: Since `SEC_PER_MIN` is:

```c
#define SEC_PER_MIN   ((UINTN)    60)
```

C conversion rules are still going to convert `Time->TimeZone` to `UINT64` before the multiplication.

In that case, I think you'll still get the correct value with two's complement and modular arithmetic, but it would be cleaner and more in the spirit of the change to cast both to signed integers:


```suggestion
    AdjustedEpochSeconds -= (INT64)Time->TimeZone * (INT64)SEC_PER_MIN;
```

capohuang:
@microsoft-github-policy-service agree company="Microsoft"

EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c
capohuang: Fixed it.

EmbeddedPkg/Library/VirtualRealTimeClockLib/VirtualRealTimeClockLib.c
capohuang: Fixed it.

makubacki (approved):

cfernald (approved):

PR #1773, opened 2026-04-22
[Squash On Rebase] MdeModulePkg: Unconditionally ApplyMemoryProtectionPolicy

apop5:
## Description

When a non-NX_COMPAT EFI_APPLICATION (e.g. grubx64.efi) is loaded,
ActivateCompatibilityMode() sets mEnhancedMemoryProtectionActive = FALSE.
From then on, GetPermissionAttributeForMemoryType() returns 0 for every
memory type.

ApplyMemoryProtectionPolicy() would get OldAttributes and NewAttributes
based upon the memory type, not the actual memory settings, and if attributes 
matched, return early without a call to SetMemoryAttributes.  This could result
in the system providing memory with unintended attributes.

Skip the type-compare shortcut. Always call SetMemoryAttributes
when returning allocated memory.

Squash on Rebase with https://github.com/microsoft/mu_basecore/commit/b8374261aae751adae10f005b67bea2858d59358
(Implement Compatibility Mode for Enhanced Memory Protection)

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Physical platform booting grubx64.

## Integration Instructions
No integration necessary.

### DIFF
--- a/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
+++ b/MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
@@ -1586,7 +1586,10 @@ ApplyMemoryProtectionPolicy (
   IN  UINT64                Length
   )
 {
-  UINT64  OldAttributes;
+  // MU_CHANGE - Start - Enforce that the Core returns memory with expected attributes
+  // UINT64  OldAttributes;
+  // MU_CHANGE - End - Enforce that the Core returns memory with expected attributes
+
   UINT64  NewAttributes;
 
   //
@@ -1645,16 +1648,21 @@ ApplyMemoryProtectionPolicy (
   //
   NewAttributes = GetPermissionAttributeForMemoryType (NewType);
 
-  if (OldType != EfiMaxMemoryType) {
-    OldAttributes = GetPermissionAttributeForMemoryType (OldType);
-    if (OldAttributes == NewAttributes) {
-      // policy is the same between OldType and NewType
-      return EFI_SUCCESS;
-    }
-  } else if (NewAttributes == 0) {
-    // newly added region of a type that does not require protection
-    return EFI_SUCCESS;
-  }
+  // MU_CHANGE - Start - Enforce that the Core returns memory with expected attributes
+  // Ensure that the memory has the expected attributes by skipping the Edk2 checks.
+  // Compat mode, Memory Attributes Protocol, Cpu Arch are all able to
+  // change memory attributes so ensure the memory range attributes match expectations.
+  // if (OldType != EfiMaxMemoryType) {
+  //  OldAttributes = GetPermissionAttributeForMemoryType (OldType);
+  //  if (OldAttributes == NewAttributes) {
+  //    // policy is the same between OldType and NewType
+  //    return EFI_SUCCESS;
+  //  }
+  // } else if (NewAttributes == 0) {
+  //  // newly added region of a type that does not require protection
+  //  return EFI_SUCCESS;
+  // }
+  // MU_CHANGE - End - Enforce that the Core returns memory with expected attribute
 
   return gCpu->SetMemoryAttributes (gCpu, Memory, Length, NewAttributes);
 }


MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1645,16 +1645,30 @@ ApplyMemoryProtectionPolicy (
...
+  // Skip the shortcut while compatibility mode is active so stale XP bits can
+  // be cleared by the SetMemoryAttributes() call below.
+  if (IsEnhancedMemoryProtectionActive ()) {
+    if (OldType != EfiMaxMemoryType) {
+      OldAttributes = GetPermissionAttributeForMemoryType (OldType);
+      if (OldAttributes == NewAttributes) {
+        // policy is the same between OldType and NewType
+        return EFI_SUCCESS;
+      }
+    } else if (NewAttributes == 0) {
```
os-d: Wouldn't this be a problem when enhanced memory protections are enabled, also, but not for all types? E.g. if you had it on for BootServicesData and not for BootServicesCode and a memory region was allocated as BootServicesData, so gets NX on it, then is freed. Then, it gets allocated as BootServicesCode. Image protections are off, so it doesn't touch it. Then we try to execute NX.

There are various combinations of settings like that that I think could cause problems. What I'm driving at is that either on free we need to set back to a known state or on allocate we need to ensure we have a known state. Because we dropped RP on free memory, I think we should have the allocate side always set either NX or RWX, depending on GetPermissionAttributeForMemoryType, regardless of whether enhanced memory protections are active.

os-d:
Can you add a SQUASH ON REBASE with the commit hash that added the IsEnhancedMemoryProtectionsActive() call?

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
apop5: Taking a step back....

CpuArch protocol is available, and any driver can locate it and use it to change memory attributes.
Memory Attributes protocol is available, and any driver can locate it and change memory attributes.
Freeing memory will not necessarily change memory attributes.

Given these scenarios, I think it makes sense to say that the Core, when it is providing memory, that the core will ensure the memory attributes match expectations. So taking out the early exits entirely would be the preferred method. Will this have implications for performance?

os-d: Agreed. There are potential implications on performance, we have seen some ARM64 platforms be very slow to change attributes. On most platforms it should not matter, but I don't think we have any choice.

However, it wouldn't be any different than in the RP on free memory case (in fact, better, since we aren't changing on free as well, only on allocate). So I think we are okay.

os-d (approved):

makubacki (approved):

MdeModulePkg/Core/Dxe/Misc/MemoryProtection.c
```diff
@@ -1645,16 +1648,21 @@ ApplyMemoryProtectionPolicy (
...
-      return EFI_SUCCESS;
-    }
-  } else if (NewAttributes == 0) {
-    // newly added region of a type that does not require protection
-    return EFI_SUCCESS;
-  }
+  // MU_CHANGE - Start - Enforce that the Core returns memory with expected attributes
+  // Ensure that allocated memory has the expected attributes by skipping
+  // the checks. Compat mode, Memory Attributes Protocol, Cpu Arch are all able
+  // to change memory attributes so ensure the Allocates match expectations.
```
makubacki: nit: Sorry if I missed it but what does "Allocates" refer to here?

>  // to change memory attributes so ensure the Allocates match expectations.

apop5: ApplyMemoryProtectionPolicy is called through a couple of different paths, most of which stem from AllocatePages/FreePages.

I'll fix the wording.

makubacki: It was just the wording that threw me off. I was looking for a local variable named "Allocates" or something like that. Thanks!

MarcChen46 (approved):

os-d (approved):

PR #1788, opened 2026-05-08
SecurityPkg: Introduce Dynamic TCG Log Scaling

Raymond-MS:
## Description

Implemented dynamic TCG log scaling in Tcg2Dxe. When the log would become truncated it instead now dynamically scales doubling the size each time. An ERROR log is reported that an increase to your base log size should occur such that scaling is not necessary. This is a precaution against platforms that log a lot and the addition of new hashing algorithms for PQC. The log is allocated in BootServices memory. The ACPI log is created on ReadyToBoot with logs being added to both until they would need to scale. In this instance a truncation event is added to the ACPI log to indicate that the log is no longer valid and/or may not contain the entirety of the log. This ACPI log is allocated in NVS memory. If the ACPI table was already allocated at the time of the ACPI log creation, it is uninstalled and reinstalled with the updated LAML and LASA PCDs. Tests were added via TcgLogTest which includes a DXE driver and a UEFI shell UnitTest app. The DXE driver handles pre-ReadyToBoot tests while the TestApp handles post-ReadyToBoot tests as well as gathering the test results from the DXE driver. Markdown documents were created to detail the changes.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested via TcgLogTest included in the reference QEMU SBSA platform with TPM enabled. Confirmed the UnitTest results. Both tests report PASS.

## Integration Instructions

Include the TcgLogTest .inf's to your platform .dsc and .fdf files. You will need to include both the TcgLogTestDxe and TcgLogTestApp for full functionality.

### DIFF
--- a/MdePkg/Include/IndustryStandard/Tpm2Acpi.h
+++ b/MdePkg/Include/IndustryStandard/Tpm2Acpi.h
@@ -80,6 +80,36 @@ typedef struct {
   UINT8     Reserved[8];
 } EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA;
 
+// MU_CHANGE - [BEGIN]
+
+typedef struct {
+  EFI_ACPI_DESCRIPTION_HEADER    Header;
+  // Flags field is replaced in version 4 and above
+  //    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
+  //    BIT16~31: Reserved
+  UINT32                         Flags;
+  UINT64                         AddressOfControlArea;
+  UINT32                         StartMethod;
+  UINT8                          PlatformSpecificParameters[12]; // size up to 12
+  UINT32                         Laml;                           // Optional
+  UINT64                         Lasa;                           // Optional
+} EFI_TPM2_ACPI_TABLE_V4;
+
+typedef struct {
+  EFI_ACPI_DESCRIPTION_HEADER                               Header;
+  // Flags field is replaced in version 4 and above
+  //    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
+  //    BIT16~31: Reserved
+  UINT32                                                    Flags;
+  UINT64                                                    AddressOfControlArea;
+  UINT32                                                    StartMethod;
+  EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA    FfaParameters;
+  UINT32                                                    Laml; // Optional
+  UINT64                                                    Lasa; // Optional
+} EFI_TPM2_ACPI_TABLE_V5;
+
+// MU_CHANGE - [END]
+
 #define EFI_TPM2_ACPI_TABLE_ARM_FFA_PARAMETER_FLAG_NOTIFICATION_SUPPORT  BIT0
 
 #define EFI_TPM2_ACPI_TABLE_ARM_FFA_PARAMETER_ATTR_MEM_TYPE_MASK           0x3
--- a/MdePkg/Include/IndustryStandard/UefiTcgPlatform.h
+++ b/MdePkg/Include/IndustryStandard/UefiTcgPlatform.h
@@ -101,6 +101,11 @@
 #define FIRMWARE_DEBUGGER_EVENT_STRING      "UEFI Debug Mode"
 #define FIRMWARE_DEBUGGER_EVENT_STRING_LEN  (sizeof(FIRMWARE_DEBUGGER_EVENT_STRING) - 1)
 
+// MU_CHANGE
+// String logged as a NO_ACTION event to mark the ACPI-visible TCG
+// log as truncated when dynamic scaling occurs post ReadyToBoot.
+#define TCG_LOG_TRUNCATION_EVENT_STRING  "TCG Event Log Truncated"
+
 //
 // Set structure alignment to 1-byte
 //
--- a/SecurityPkg/SecurityPkg.dec
+++ b/SecurityPkg/SecurityPkg.dec
@@ -303,7 +303,7 @@
   ## MU_CHANGE - END - Add a new protocol to support Log-only events.
 
   ## MU_CHANGE - [BEGIN]
-  ## TCG Log Test protocol produced by TcgLogTestDxe.
+  ## Protocol used to test dynamic TCG log scaling functionality.
   # Tcg/TcgLogTest/TcgLogTest.h
   gTcgLogTestProtocolGuid            = {0xa3c12f80, 0x7d9e, 0x4b5a, { 0x91, 0xe4, 0x6c, 0xf8, 0x2d, 0xa1, 0xb7, 0x03 }}
   ## MU_CHANGE - [END]
--- a/SecurityPkg/SecurityPkg.dec
+++ b/SecurityPkg/SecurityPkg.dec
@@ -301,6 +301,13 @@
   # Include/Protocol/MuTcg2Protocol.h
   gMuTcg2ProtocolExGuid              = {0x227e7984, 0x1a77, 0x4762, { 0x96, 0x69, 0x57, 0x4c, 0xda, 0xd1, 0xa0, 0x1e }}
   ## MU_CHANGE - END - Add a new protocol to support Log-only events.
+
+  ## MU_CHANGE - [BEGIN]
+  ## TCG Log Test protocol produced by TcgLogTestDxe.
+  # Tcg/TcgLogTest/TcgLogTest.h
+  gTcgLogTestProtocolGuid            = {0xa3c12f80, 0x7d9e, 0x4b5a, { 0x91, 0xe4, 0x6c, 0xf8, 0x2d, 0xa1, 0xb7, 0x03 }}
+  ## MU_CHANGE - [END]
+
 [Ppis]
   ## The PPI GUID for that TPM physical presence should be locked.
   # Include/Ppi/LockPhysicalPresence.h
--- a/SecurityPkg/SecurityPkg.dsc
+++ b/SecurityPkg/SecurityPkg.dsc
@@ -248,6 +248,8 @@
   SecurityPkg/Library/PeiDxeTpmPlatformHierarchyLibNull/PeiDxeTpmPlatformHierarchyLib.inf
 
   SecurityPkg/Applications/TpmShellApp/TpmShellApp.inf  ## MU_CHANGE
+  SecurityPkg/Tcg/TcgLogTest/TcgLogTestDxe.inf  ## MU_CHANGE
+  SecurityPkg/Tcg/TcgLogTest/TcgLogTestApp.inf  ## MU_CHANGE
 
   #
   # TCG Storage.
--- a/SecurityPkg/SecurityPkg.dsc
+++ b/SecurityPkg/SecurityPkg.dsc
@@ -248,8 +248,16 @@
   SecurityPkg/Library/PeiDxeTpmPlatformHierarchyLibNull/PeiDxeTpmPlatformHierarchyLib.inf
 
   SecurityPkg/Applications/TpmShellApp/TpmShellApp.inf  ## MU_CHANGE
-  SecurityPkg/Tcg/TcgLogTest/TcgLogTestDxe.inf  ## MU_CHANGE
-  SecurityPkg/Tcg/TcgLogTest/TcgLogTestApp.inf  ## MU_CHANGE
+
+  ## MU_CHANGE - [BEGIN]
+  SecurityPkg/Tcg/TcgLogTest/TcgLogTestDxe.inf
+  SecurityPkg/Tcg/TcgLogTest/TcgLogTestApp.inf {
+    <LibraryClasses>
+      UnitTestLib|UnitTestFrameworkPkg/Library/UnitTestLib/UnitTestLib.inf
+      UnitTestPersistenceLib|UnitTestFrameworkPkg/Library/UnitTestPersistenceLibNull/UnitTestPersistenceLibNull.inf
+      UnitTestResultReportLib|UnitTestFrameworkPkg/Library/UnitTestResultReportLib/UnitTestResultReportLibDebugLib.inf
+  }
+  ## MU_CHANGE - [END]
 
   #
   # TCG Storage.
--- a/SecurityPkg/Tcg/Tcg2Dxe/README.md
+++ b/SecurityPkg/Tcg/Tcg2Dxe/README.md
@@ -1,0 +1,82 @@
+# Tcg2Dxe
+
+Tcg2Dxe is a DXE-phase UEFI driver that publishes the TCG2 protocol defined
+by the TCG EFI Protocol Specification. It's main responsibilites are to expose
+a standard interface to a TPM device, measure components and events into PCRs,
+support measured boot, and enable secure boot attestation.
+
+## Dynamic Event Log Scaling
+
+The TCG event log is initially allocated with a fixed size defined by a
+PCD: PcdTcgLogAreaMinLen. As firmware components log measured boot
+events the log fills up. Traditionally, when the log is full, subsequent events
+are dropped and the log is marked as truncated.
+
+Tcg2Dxe extends this behavior with **dynamic scaling**: when the log is about
+to overflow, the driver doubles its allocation, copies the existing log into
+the new buffer, and frees the old one. This allows the log to grow as needed
+and avoids losing events.
+
+### How It Works
+
+1. **Scaling check** — Before logging a TCG 2.0 event,
+   `TcgLogDynamicScalingNeeded` calculates whether the new event (plus a
+   reserved truncation marker) would exceed the current allocation
+   (`EventLogAreaStruct->Laml`). Space is reserved for the truncation marker
+   as long as the ACPI log has not yet been marked truncated.
+
+2. **Reallocation** — When scaling is needed, `TcgScaleEventLog` allocates a
+   new `EfiBootServicesData` region at twice the current size, copies the
+   existing log, updates the `Lasa`/`Laml` fields in the event log area
+   struct, and frees the old region.
+
+3. **Logging** — After scaling, the new event is logged into the resized buffer
+   via `TcgDxeLogEvent` inside a TPL-raised critical section.
+
+### Normal Log vs. ACPI Log vs. Final Events Log
+
+Tcg2Dxe maintains three distinct event log regions:
+
+| Log | Memory Type | Lifetime | Can Scale |
+| --- | ----------- | -------- | --------- |
+| **Normal log** | `EfiBootServicesData` | Available until `ExitBootServices` | Yes |
+| **ACPI log** | `EfiACPIMemoryNVS` | Persistent | No |
+| **Final Events log** | `EfiACPIMemoryNVS` | Persistent | No |
+
+- The **Normal log** is the main log copy which is returned via `GetEventLog`.
+  It can grow dynamically via scaling. Note that previous calls to `GetEventLog`
+  could contain stale data if the log was scaled after. It is recommended to
+  call `GetEventLog` each time access is required.
+- The **ACPI log** is created at `ReadyToBoot` by `GenerateAcpiLog`. It
+  allocates an `EfiACPIMemoryNVS` region equal to the **Normal log** size at that
+  point, copies the log contents, and updates the TPM2 ACPI table's
+  `LAML`/`LASA` fields so the OS can find it. If the TPM2 table was already
+  installed via Tcg2Acpi/Tcg2AcpiFfa, `GenerateAcpiLog` will uninstall and
+  reinstall the ACPI table with the updated LAML and LASA pointing to the
+  newly allocated NVS region.
+- The **Final Events log** (`EFI_TCG2_FINAL_EVENTS_TABLE`) records events
+  logged after `GetEventLog` has been called. It is installed as a UEFI
+  configuration table so the OS can discover events that occurred between its
+  call to `GetEventLog` and `ExitBootServices`. Because the **Final Events log**
+  does not scale, it can become truncated.
+
+After `ReadyToBoot`, every new event is also appended to the ACPI log. Because
+the ACPI log's NVS allocation is fixed (the OS may already be referencing its
+address), it cannot be reallocated.
+
+### Post-ReadyToBoot Truncation
+
+When dynamic scaling is triggered after `ReadyToBoot`:
+
+1. A `NO_ACTION` event with the payload `"TCG Event Log Truncated"` is
+   appended to the **ACPI log** to notify the OS that the ACPI-visible log is
+   now incomplete.
+2. The ACPI log is marked truncated (`EventLogTruncated = TRUE`) so no further
+   events are written to it and no additional space is reserved for the
+   truncation marker.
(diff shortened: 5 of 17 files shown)


SecurityPkg/Tcg/Tcg2Dxe/README.md
```diff
@@ -0,0 +1,82 @@
+# Tcg2Dxe
+
+Tcg2Dxe is a DXE-phase UEFI driver that publishes the TCG2 protocol defined
+by the TCG EFI Protocol Specification. It's main responsibilites are to expose
```
makubacki: It would be helpful for readers to link to exactly what you're talking about here.

```suggestion
by the [TCG EFI Protocol Specification](https://trustedcomputinggroup.org/resource/tcg-efi-protocol-specification/). It's main responsibilites are to expose
```

SecurityPkg/SecurityPkg.dec
```diff
@@ -301,6 +301,13 @@
   # Include/Protocol/MuTcg2Protocol.h
   gMuTcg2ProtocolExGuid              = {0x227e7984, 0x1a77, 0x4762, { 0x96, 0x69, 0x57, 0x4c, 0xda, 0xd1, 0xa0, 0x1e }}
   ## MU_CHANGE - END - Add a new protocol to support Log-only events.
+
+  ## MU_CHANGE - [BEGIN]
+  ## TCG Log Test protocol produced by TcgLogTestDxe.
```
makubacki: I know it's intended to be helpful, but implementation details like where a protocol is produced should be avoided in package-level definitions. People can easily find it in an INF search and the text here is not at risk of silently becoming inaccurate over time.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -134,6 +136,31 @@ VARIABLE_TYPE  mVariableType[] = {
...
 EFI_HANDLE  mImageHandle;
 
+// MU_CHANGE - [BEGIN]
+
+BOOLEAN                    mReadyToBoot = FALSE;
+TCG_EVENT_LOG_AREA_STRUCT  mAcpiEventLog;
+
+// String logged as a NO_ACTION event to mark the ACPI-visible TCG
+// log as truncated when dynamic scaling occurs post ReadyToBoot.
+#define TCG_LOG_TRUNCATION_EVENT_STRING  "TCG Event Log Truncated"
```
makubacki: Is this string defined in a specification?

Regardless, from a code maintainability perspective, this should reside in a header file.

Those in TCG specs (like PC PFP) are in MdePkg\Include\IndustryStandard\UefiTcgPlatform.h. Not saying this goes there, but that reside somewhere outside this specific DXE driver to follow conventions.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -134,6 +136,31 @@ VARIABLE_TYPE  mVariableType[] = {
...
+
+typedef struct {
+  EFI_ACPI_DESCRIPTION_HEADER    Header;
+  UINT32                         Flags;
+  UINT64                         AddressOfControlArea;
+  UINT32                         StartMethod;
+  UINT8                          PlatformSpecificParameters[12];
+  UINT32                         Laml;
+  UINT64                         Lasa;
+} EFI_TPM2_ACPI_TABLE_V4;
```
makubacki: This (`EFI_TPM2_ACPI_TABLE_V4`) would similarly be better in a header file per covention.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -2575,6 +2825,161 @@ MeasureSecureBootPolicy (
...
+  mAcpiEventLog.Lasa                  = AcpiLasa;
+  mAcpiEventLog.Laml                  = EventLogAreaStruct->Laml;
+  mAcpiEventLog.EventLogSize          = EventLogAreaStruct->EventLogSize;
+  mAcpiEventLog.LastEvent             = (UINT8 *)(UINTN)AcpiLasa + EventLogAreaStruct->EventLogSize;
+  mAcpiEventLog.EventLogStarted       = EventLogAreaStruct->EventLogStarted;
+  mAcpiEventLog.EventLogTruncated     = FALSE;
+  mAcpiEventLog.Next800155EventOffset = EventLogAreaStruct->Next800155EventOffset;
+
+  // Copy the data to the ACPI log.
+  CopyMem ((VOID *)(UINTN)AcpiLasa, (VOID *)(UINTN)EventLogAreaStruct->Lasa, (UINTN)mAcpiEventLog.Laml);
```
makubacki: Is it necessary to copy the full buffer here when the log size used is known (`EventLogAreaStruct->EventLogSize`)?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1133,6 +1186,151 @@ CopyDigestListBinToBuffer (
...
+  DEBUG ((DEBUG_INFO, "NewLasa: 0x%lx, NewLaml: 0x%x\n", NewLasa, NewLaml));
+
+  // Update the EventLogAreaStruct.
+  EventLogAreaStruct->Lasa = NewLasa;
+  EventLogAreaStruct->Laml = NewLaml;
+
+  // Once we reach ReadyToBoot, we do not want to free the old log regions, this is
+  // due to the ACPI table containing the old log pointer, we do not want to
+  // invalidate this memory.
+  gBS->FreePages (OldLasa, EFI_SIZE_TO_PAGES ((UINTN)OldLaml));
```
makubacki: I think this comment introduces more confusion than it helps since the free is unconditional here.

> "Once we reach ReadyToBoot, we do not want to free the old log regions".

Can you elaborate further in the comment to clarify it against the code?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1133,6 +1186,151 @@ CopyDigestListBinToBuffer (
...
+  if (NewEventSize > MAX_ADDRESS - NewEventHdrSize) {
+    return FALSE;
+  }
+
+  NewLogSize = NewEventHdrSize + NewEventSize;
+
+  // Reserve space for the truncation NO_ACTION_EVENT so that the log can never
+  // fill past the point where the truncation marker would not fit. We only need
+  // to continue to reserve space as long as the ACPI event log is not truncated.
+  if (!mAcpiEventLog.EventLogTruncated) {
```
makubacki: It looks like `GenerateAcpiLog()` does not create the ACPI event log until ReadyToBoot? In that case, it would seem that this should be guarded.


```suggestion
  if (mReadyToBoot && !mAcpiEventLog.EventLogTruncated) {
```

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1002,6 +1029,32 @@ TcgDxeLogEvent (
...
+               EventLogAreaStruct,
+               NewEventHdr,
+               NewEventHdrSize,
+               NewEventData,
+               NewEventSize
+               );
+
+    if (Status == EFI_OUT_OF_RESOURCES) {
+      EventLogAreaStruct->EventLogTruncated = TRUE;
+      return EFI_VOLUME_FULL;
```
makubacki: I'm not suggesting a change or anything, but is it necessary to return here versus continue to the FinalEventsTable logic if this fails for the ACPI event log?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1133,6 +1186,151 @@ CopyDigestListBinToBuffer (
...
+  if (NewLogSize + EventLogAreaStruct->EventLogSize > EventLogAreaStruct->Laml) {
+    DEBUG ((DEBUG_INFO, "  Laml       - 0x%lx\n", EventLogAreaStruct->Laml));
+    DEBUG ((DEBUG_INFO, "  NewLogSize - 0x%lx\n", NewLogSize));
+    DEBUG ((DEBUG_INFO, "  LogSize    - 0x%lx\n", EventLogAreaStruct->EventLogSize));
+    DEBUG ((DEBUG_ERROR, "Dynamic scaling required! Recommended to update your TCG log size!\n"));
+
+    // Log an error if we attempt to scale post ReadyToBoot.
+    if (mReadyToBoot) {
+      DEBUG ((DEBUG_ERROR, "Scaling post ReadyToBoot is invalid!\n"));
+    }
```
makubacki, line 1287: I might be missing something, but for:

```c
    if (mReadyToBoot) {
      DEBUG ((DEBUG_ERROR, "Scaling post ReadyToBoot is invalid!\n"));
    }
```

- Why wait until this far into the function to determine this?
- Why still return `TRUE` (scaling needed) if it is invalid?
- Isn't scaling post-ReadyToBoot considered valid (elsewhere)?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1133,6 +1186,151 @@ CopyDigestListBinToBuffer (
...
+      return FALSE;
+    }
+
+    NewLogSize += EventSize;
+  }
+
+  // Validate EventLogSize + NewLogSize doesn't cause an overflow.
+  if (NewLogSize > MAX_ADDRESS - EventLogAreaStruct->EventLogSize) {
+    return FALSE;
+  }
```
makubacki, line 1275: If this overflows, you might want to do something that just return `FALSE`.

makubacki (commented):
@Raymond-MS, I found the accompanying documentation helpful. I left some comments in a quick first pass, I might follow up when I have time to look in more detail.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: This is a good catch, I don't believe I meant to do that. It should definitely be EventLogAreaStruct-EventLogSize. Thanks! I did EventLogSize in the scaling of the log, should be doing the same here.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: The ACPI log is not created until ReadyToBoot but we want to make sure that the normal log has enough room for the truncation event as when we create the ACPI log we use the size of the current log as well as copying its contents. Without this check it is entirely possible that when we create the ACPI log region that the truncation event will not fit, having this check here until the ACPI log gets truncated guarantees that we always have enough room for the truncation event.

A change I could make would be to allocate the ACPI log with LAML + sizeof (TruncationEvent). Thoughts?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: This resulted from a conversation with Chris. The idea was to have an ASSERT (FALSE) here but that would prevent my tests from running to completion. Just adding a log was what we concluded on having.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: I think if this happens an ASSERT would be good. We should never really hit this.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: No, this is not a spec defined value. It's the data associated with the truncation event, I just made it a string. I will move it to UefiTcgPlatform.h. It's probably the best location for it.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Sounds good. I moved it to Tpm2Acpi.h.

SecurityPkg/SecurityPkg.dec
Raymond-MS: Sounds good. Updated.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Good catch. I am sorry, this is left over from when we only freed before ReadyToBoot. That changed when we decided to allocate the ACPI log at ReadyToBoot. Removed.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Good catch. This is a bug. We do not want to return here as it will cause the FinalEventsTable not to be logged to if the ACPI table gets truncated. Removed.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: I updated to just allocate extra memory for the truncation event.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: To add some more context, scaling post-ReadyToBoot is discouraged because we do not know when GetEventLog was last called. In this instance, a caller could have a stale "freed" pointer to the old log region. Chris had mentioned that scaling post-ReadyToBoot should not really occur and if it does, it is an indicator of some kind of failure.

This is similar to why the ACPI log gets marked truncated, there is no other way for a user of the ACPI log to know the contents are incomplete.

In general, we don't want scaling to ever occur which is why we have the DEBUG log of increasing your TCG log size is recommended. This implementation is really just a safety net in case there are instances where the log would become truncated, we would rather have it continue to work rather than failing outright.

makubacki: The idea to truncate the ACPI log with a marker and allow the `GetEventLog()` log to continue scaling seem reasonable, it is just that there is a logical contradiction between the debug message which states "Scaling post ReadyToBoot is invalid!" and the function the immediately returning `TRUE` (explicitly allowing the "invalid" action).

I think it would be more natural to:

1. Have this function just implement logic rather than policy feedback based on the scaling decision. So, the caller prints a message based on the result, if needed.
2. To update the message to be more precise/actionable. Like "Unexpected scaling needed post-ReadyToBoot <action>". It's unclear right now, why it's considered invalid but allowed and what the platform owner should do about it.

Raymond-MS: Sounds good, I can update the comment to be less confusing.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -2575,6 +2788,191 @@ MeasureSecureBootPolicy (
...
+    return EFI_NOT_FOUND;
+  }
+
+  // Walk installed ACPI tables looking for the TPM2 signature.
+  Index = 0;
+  while (TRUE) {
+    Status = AcpiSdt->GetAcpiTable (Index, &SdtHeader, &Version, &TableKey);
+    if (EFI_ERROR (Status)) {
+      // Reached the end of the table list without finding TPM2.
+      DEBUG ((DEBUG_WARN, "%a: TPM2 table not found\n", __func__));
```
Flickdm: NIT: All errors are considered "reached the end of the table list"

```
if (EFI_ERROR (Status)) {
  DEBUG ((DEBUG_WARN, "%a: GetAcpiTable failed at %u - %r\n", __func__, Index, Status));
  return EFI_NOT_FOUND;
}
```

MdePkg/Include/IndustryStandard/UefiTcgPlatform.h
```diff
@@ -101,6 +101,11 @@
 #define FIRMWARE_DEBUGGER_EVENT_STRING      "UEFI Debug Mode"
 #define FIRMWARE_DEBUGGER_EVENT_STRING_LEN  (sizeof(FIRMWARE_DEBUGGER_EVENT_STRING) - 1)
 
+// MU_CHANGE
+// String logged as a NO_ACTION event to mark the ACPI-visible TCG
+// log as truncated when dynamic scaling occurs post ReadyToBoot.
+#define TCG_LOG_TRUNCATION_EVENT_STRING  "TCG Event Log Truncated"
+
```
spbrogan, line 108: follow existing pattern to add len

spbrogan, line 108: Add comment that this is not yet spec defined.  do we have plans to take to tcg?

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
```diff
@@ -80,6 +80,36 @@ typedef struct {
...
+  // Flags field is replaced in version 4 and above
+  //    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
+  //    BIT16~31: Reserved
+  UINT32                                                    Flags;
+  UINT64                                                    AddressOfControlArea;
+  UINT32                                                    StartMethod;
+  EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA    FfaParameters;
+  UINT32                                                    Laml; // Optional
+  UINT64                                                    Lasa; // Optional
+} EFI_TPM2_ACPI_TABLE_V5;
```
spbrogan, line 109: is this really a version 5 table and is it really hard coded to FFA?  Seems like we should use unions and make a more flexible data structure.  What is the guidance for non FFA TPMs.  What table version do they use?   I also don't see a place for EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_SMC in the table.  Again, I assume there is a flavor of the table that has this.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
```diff
@@ -80,6 +80,36 @@ typedef struct {
...
+  // Flags field is replaced in version 4 and above
+  //    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
+  //    BIT16~31: Reserved
+  UINT32                         Flags;
+  UINT64                         AddressOfControlArea;
+  UINT32                         StartMethod;
+  UINT8                          PlatformSpecificParameters[12]; // size up to 12
+  UINT32                         Laml;                           // Optional
+  UINT64                         Lasa;                           // Optional
+} EFI_TPM2_ACPI_TABLE_V4;
```
spbrogan, line 96: need to combine this with EFI_TPM2_ACPI_TABLE

SecurityPkg/Tcg/Tcg2Dxe/README.md
```diff
@@ -0,0 +1,83 @@
...
+secure boot attestation.
+
+## Dynamic Event Log Scaling
+
+The TCG event log is initially allocated with a fixed size defined by a
+PCD: PcdTcgLogAreaMinLen. As firmware components log measured boot
+events the log fills up. Traditionally, when the log is full, subsequent events
+are dropped and the log is marked as truncated.
+
+Tcg2Dxe extends this behavior with **dynamic scaling**: when the log is about
```
spbrogan, line 16: probably important to call out here the different behaviors based on time of execution (pre ready to boot vs post).

SecurityPkg/SecurityPkg.dsc
```diff
@@ -249,6 +249,16 @@
 
   SecurityPkg/Applications/TpmShellApp/TpmShellApp.inf  ## MU_CHANGE
 
+  ## MU_CHANGE - [BEGIN]
+  SecurityPkg/Tcg/TcgLogTest/TcgLogTestDxe.inf
+  SecurityPkg/Tcg/TcgLogTest/TcgLogTestApp.inf {
+    <LibraryClasses>
+      UnitTestLib|UnitTestFrameworkPkg/Library/UnitTestLib/UnitTestLib.inf
+      UnitTestPersistenceLib|UnitTestFrameworkPkg/Library/UnitTestPersistenceLibNull/UnitTestPersistenceLibNull.inf
```
spbrogan, line 257: these don't look right for a functioning app

SecurityPkg/SecurityPkg.dec
```diff
@@ -301,6 +301,13 @@
   # Include/Protocol/MuTcg2Protocol.h
   gMuTcg2ProtocolExGuid              = {0x227e7984, 0x1a77, 0x4762, { 0x96, 0x69, 0x57, 0x4c, 0xda, 0xd1, 0xa0, 0x1e }}
   ## MU_CHANGE - END - Add a new protocol to support Log-only events.
+
+  ## MU_CHANGE - [BEGIN]
+  ## Protocol used to test dynamic TCG log scaling functionality.
```
spbrogan, line 306: this is a private protocol with visibility only to the test app and test dxe driver.  

I could see two options. 
1. just note that here
2.  don't create a package defined global for it and just include the guid define in the header file.

SecurityPkg/Tcg/TcgLogTest/TcgLogTestCommon.c
```diff
@@ -0,0 +1,499 @@
...
+#include <Protocol/Tcg2Protocol.h>
+#include <IndustryStandard/UefiTcgPlatform.h>
+
+#include "TcgLogTestCommon.h"
+
+#define TCG_LOG_TEST_PCR_INDEX      8
+#define TCG_LOG_TEST_EVENT_TYPE     EV_NO_ACTION
+#define TCG_LOG_TEST_EVENT_PAYLOAD  "TcgLogTestDxeEvent"
+
+STATIC CHAR8  mCommonEventPayload[] = TCG_LOG_TEST_EVENT_PAYLOAD;
```
spbrogan, line 25: this will probably cause a problem with initializer and compiler intrinsics that are not supported.

MdePkg/Include/IndustryStandard/UefiTcgPlatform.h
Raymond-MS: I believe that when we upstream, if this feature is adopted by the EDK2 community, it would be in our best interest to bring this to the TCG spec.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
Raymond-MS: I created a union of PlatformSpecificParameters, SmcParameters, and FfaParameters. All have the same length and allow backwards compatibility with the other versions of the table structure.

Raymond-MS:
Should be going with the V2 version here:

https://github.com/microsoft/mu_basecore/pull/1805

Closing as it's no longer needed.

PR #1792, opened 2026-05-11
MdeModulePkg: DxeIplPeim Move Decompress to MsCorePkg/Core/GuidedSectionExtractPeim

apop5:
## Description

  There are scenarios where PEIMs contained in compressed FVs are needed to complete
  the PEI stage (such as DxeIpl being in a compressed FV). GuidedSectionExtractPeim
  makes decompression available after memory is discovered instead of needing to
  wait for dxeipl to be given control. The functionality being remove here is entirely
  in GuidedSectionExtractPeimDisables, and requires platforms to include the PEIM
  out of hte MsCorePkg when they require decompression of Fvs.
  
  This change was split off the commit:
  https://github.com/microsoft/mu_basecore/commit/7a38833

- [X] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested
Physical platform encountered where multiple installations of Extract Guided section
PPIs triggered errors. 

## Integration Instructions
If a break is encountered, verify that the MsCorePkg's GuidedSectionExtractPeim is included
in the platforms DSC/FDF.

### DIFF
--- a/MdeModulePkg/Core/DxeIplPeim/DxeIpl.h
+++ b/MdeModulePkg/Core/DxeIplPeim/DxeIpl.h
@@ -15,9 +15,13 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Ppi/EndOfPeiPhase.h>
 #include <Ppi/MemoryDiscovered.h>
 #include <Ppi/ReadOnlyVariable2.h>
-#include <Ppi/Decompress.h>
+// MU_CHANGE [BEGIN] - Move Decompress to MsCorePkg/Core/GuidedSectionExtractPeim
+// #include <Ppi/Decompress.h>
+// MU_CHANGE [END] - Move Decompress to MsCorePkg/Core/GuidedSectionExtractPeim
 #include <Ppi/FirmwareVolumeInfo.h>
-#include <Ppi/GuidedSectionExtraction.h>
+// MU_CHANGE [BEGIN] - Move Decompress to MsCorePkg/Core/GuidedSectionExtractPeim
+// #include <Ppi/GuidedSectionExtraction.h>
+// MU_CHANGE [END] - Move Decompress to MsCorePkg/Core/GuidedSectionExtractPeim
 #include <Ppi/LoadFile.h>
 #include <Ppi/S3Resume2.h>
 #include <Ppi/RecoveryModule.h>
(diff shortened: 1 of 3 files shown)


vineelko (approved):

os-d (commented):
Can we make DxeIpl smarter instead? This change won't be accepted upstream, it would be nice if we could not carry this forever.

apop5:
> Can we make DxeIpl smarter instead? This change won't be accepted upstream, it would be nice if we could not carry this forever.

We can, but at the cost of size. We could make the DxeIpl smart enough to not install a Ppi for anything that is already available in the system, but then we end up carrying upto 2 copies of the decompress code, one from MsCorePkg and the other in DxeIpl.

os-d:
> > Can we make DxeIpl smarter instead? This change won't be accepted upstream, it would be nice if we could not carry this forever.
> 
> We can, but at the cost of size. We could make the DxeIpl smart enough to not install a Ppi for anything that is already available in the system, but then we end up carrying upto 2 copies of the decompress code, one from MsCorePkg and the other in DxeIpl.

Could the extraction logic be extracted (no pun intended) into a lib (if not already) and platforms using the MsCorePkg PEIM can link the null lib to DxeIpl?

apop5:
> > > Can we make DxeIpl smarter instead? This change won't be accepted upstream, it would be nice if we could not carry this forever.
> > 
> > 
> > We can, but at the cost of size. We could make the DxeIpl smart enough to not install a Ppi for anything that is already available in the system, but then we end up carrying upto 2 copies of the decompress code, one from MsCorePkg and the other in DxeIpl.
> 
> Could the extraction logic be extracted (no pun intended) into a lib (if not already) and platforms using the MsCorePkg PEIM can link the null lib to DxeIpl?

After offline discussion

1) Get this in for current platforms relying on it
2) Take GuidedSectionExtractPeim to edk2. And propose getting rid of code out of DxeIpl for publishing extraction. Deal with pushback from breaking change (requiring platforms to include a new PEIM) and come to agreement on when such a breaking change can be introduced into edk2.

makubacki:
> > Can we make DxeIpl smarter instead? This change won't be accepted upstream, it would be nice if we could not carry this forever.
> 
> We can, but at the cost of size. We could make the DxeIpl smart enough to not install a Ppi for anything that is already available in the system, but then we end up carrying upto 2 copies of the decompress code, one from MsCorePkg and the other in DxeIpl.

A platform does not have to link an actual decompression implementation to DxeIpl if it's not going to be used. Today's platforms should already not be doing that (and the ones I've checked do not link decompression code to DxeIpl).

> Could the extraction logic be extracted (no pun intended) into a lib (if not already) and platforms using the MsCorePkg PEIM can link the null lib to DxeIpl?

Decompression is a specific instance of section extraction. ExtractGuidedSectionLib already exists to abstraction section extraction itself.

makubacki:
> After offline discussion
> 
> 1. Get this in for current platforms relying on it
> 2. Take GuidedSectionExtractPeim to edk2. And propose getting rid of code out of DxeIpl for publishing extraction. Deal with pushback from breaking change (requiring platforms to include a new PEIM) and come to agreement on when such a breaking change can be introduced into edk2.

I have no problem with this PR being merged and separating section extraction linkage to a dedicated module instead of sharing responsibility with DXE IPL makes sense. That said, GuidedSectionExtractPeim needs some cleanup before it goes upstream.

makubacki (approved):

os-d (approved):

PR #1793, opened 2026-05-12
[Rebase & FF] .pytool: Image Validation Plugin: Drop Zero ImageBase Requirement

vineelko:
## Description

- Zero ImageBase requirement is dropped from Image validation plugin as it is not enforced by most PE loaders in UEFI.
- Patina in particular dropped support for producing .efi files with zero ImageBase
- This also improves the debugging experience when using WinDbg
---
- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Q35

## Integration Instructions

NA


Javagedes (commented):
Can you make this a REBASE&FF so that line endings are a different commit then the line changes?

vineelko:
The modified files had not been changed since 2022, and the repository added an `.editorconfig` file in the meantime. As a result, saving the files now enforces CRLF line endings.

.pytool/Plugin/ImageValidation/image_validation.cfg
```diff
@@ -1,70 +1,69 @@
-TARGET_ARCH:
```
os-d, line 1: Line endings changed, was it incorrect (LF) before and this fixes it?

makubacki (approved):

apop5 (approved):

PR #1803, opened 2026-05-19
ArmPkg: MmCommunicationDxe: Fix Comm Buffer Init

os-d:
## Description

Currently, MmCommunicationDxe expects the MM comm buffer to either be not added by the platform or added and allocated by the platform. However, not all platforms follow this pattern.

This commit makes the handling more generic by checking to see if a GCD descriptor exists that covers this range. If it does, the capabilities are updated as needed. If only a partial desc exists for this range, the driver will fail as that is a platform misconfiguration.

If the descriptor does not exist, the driver will add the buffer.

Then, for all successful cases, the attributes are updated on the buffer to what the driver expects them to be.

This will be taken to edk2 in parallel.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on an aarch64 platform that was producing the HOB but not having it allocated. This fixed it.

## Integration Instructions

Platforms must either not produce a resource descriptor HOB for the MM Communicate Buffer or must produce a HOB fully covering the region.

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -5,6 +5,9 @@
   SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
+
+#include <PiDxe.h>
+
 #include <Library/ArmLib.h>
 #include <Library/ArmFfaLib.h>
 #include <Library/ArmSmcLib.h>
@@ -23,6 +26,8 @@
 #include <IndustryStandard/ArmStdSmc.h>
 #include <IndustryStandard/ArmFfaSvc.h>
 #include <IndustryStandard/MmCommunicate.h>
+
+#define COMM_BUFFER_ATTRS (EFI_MEMORY_WB | EFI_MEMORY_XP | EFI_MEMORY_RUNTIME)
 
 //
 // Partition ID if FF-A support is enabled
@@ -697,6 +702,7 @@ MmCommunication2Initialize (
 {
   EFI_STATUS  Status;
   UINTN       Index;
+  EFI_GCD_MEMORY_SPACE_DESCRIPTOR  GcdDescriptor;
 
   // Initialize to make mm communication
   Status = InitializeCommunication ();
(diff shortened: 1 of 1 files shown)


ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -713,75 +719,56 @@ MmCommunication2Initialize (
 
   ASSERT (mNsCommBuffMemRegion.Length != 0);
 
-  Status = gDS->AddMemorySpace (
-                  EfiGcdMemoryTypeReserved,
+  Status = gDS->GetMemorySpaceDescriptor (
```
makubacki, line 722: Does this account for unallocated memory being represented as `EfiGcdMemoryTypeNonExistent`? It looks like this is assuming `EFI_NOT_FOUND` will be returned in that case.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -713,75 +719,56 @@ MmCommunication2Initialize (
...
+    ASSERT_EFI_ERROR (Status);
   }
 
   Status = gDS->SetMemorySpaceAttributes (
                   mNsCommBuffMemRegion.PhysicalBase,
                   mNsCommBuffMemRegion.Length,
-                  EFI_MEMORY_WB | EFI_MEMORY_XP | EFI_MEMORY_RUNTIME
+                  COMM_BUFFER_ATTRS
                   );
   if (EFI_ERROR (Status)) {
```
makubacki, line 783: Technically, `CleanAddedMemorySpace` is no longer called here only for "added memory spaces" since the call to `GetMemorySpaceDescriptor()` could have made it this far with a pre-existing descriptor. In that case, are you intending to remove the memory space unconditionally?

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -713,75 +719,56 @@ MmCommunication2Initialize (
...
-        DEBUG_ERROR,
-        "MmCommunicateInitialize: "
-        "Failed to remove existing MM-NS Buffer Memory Space - %r\n",
-        Status
-        ));
-      goto ReturnErrorStatus;
-    }
 
-    // Try to add the memory space again
+  if (EFI_ERROR (Status) || (GcdDescriptor.GcdMemoryType == EfiGcdMemoryTypeNonExistent)) {
```
makubacki, line 727: Since this module's primary responsibility is validating/initializing the comm buffer, it might also be worth a debug print (or something similar) when the descriptor type is unexpected like `EfiGcdMemoryTypeMemoryMappedIo`.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -713,75 +719,56 @@ MmCommunication2Initialize (
...
         ));
       goto ReturnErrorStatus;
     }
+  } else if (!(GcdDescriptor.BaseAddress <= mNsCommBuffMemRegion.PhysicalBase &&
+               (GcdDescriptor.BaseAddress + GcdDescriptor.Length) >=
+               (mNsCommBuffMemRegion.PhysicalBase + mNsCommBuffMemRegion.Length)))
+  {
+    DEBUG ((
+      DEBUG_ERROR,
+      "%a: Existing GCD memory space does not completely cover MM-NS Buffer Memory Space\n",
```
makubacki, line 751: nit: Other prints use `MmCommunicationInitialize` or `MmCommunicateInitialize`. This will be different since it will match the actual function name `MmCommunication2Initialize`. I think they should all be updated to use `%a` with `__func__`.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
os-d: Good point. I thought GetMemorySpaceDescriptor would return not found in that case, but it does not. I have updated it to handle that.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
os-d: Yeah, I chatted with @kuqin12 on the expectations here. I find it odd that this driver accepts that the memory might be added or might not be added. i.e. the contract isn't clear.

I can add a print and assert here if the region is not reserved

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
os-d: Good point. Tbh, I don't think there is much point in removing the memory space at all, if the platform relies on MM (which it presumably does if it includes this driver), it won't boot without mm communicate working. I'll remove CleanAddedMemorySpace.

os-d:
@makubacki I addressed the current feedback and re-tested, please re-review when you have a chance.

kuqin12 (approved):

makubacki (approved):

PR #1805, opened 2026-05-22
SecurityPkg: Introduce Dynamic TCG Log Scaling V2

Raymond-MS:
## Description

Implemented dynamic TCG log scaling in Tcg2Dxe. When the log would become truncated it instead now dynamically scales doubling the size each time. An ERROR log is reported that an increase to your base log size should occur such that scaling is not necessary. This is a precaution against platforms that log a lot and the addition of new hashing algorithms for PQC. The log is allocated in BootServices memory. Tests were added via TcgLogTest which includes a DXE driver and a UEFI shell UnitTest app. The DXE driver handles pre-ReadyToBoot tests while the TestApp handles post-ReadyToBoot tests as well as gathering the test results from the DXE driver. Markdown documents were created to detail the changes.

This version of dynamic scaling never sets the ACPI table LAML/LASA which means the table is never published with the log information. As such the only way to access the event log is through the Tcg2Protocol published by Tcg2Dxe. The LAML/LASA fields are OPTIONAL and when not set are removed from the table.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested via TcgLogTest included in the reference QEMU ARM VIRT platform with TPM enabled. Confirmed the UnitTest results. All tests report PASS.

## Integration Instructions

Include the TcgLogTest .inf's to your platform .dsc and .fdf files. You will need to include both the TcgLogTestDxe and TcgLogTestApp for full functionality.

### DIFF
--- a/MdePkg/Include/IndustryStandard/Tpm2Acpi.h
+++ b/MdePkg/Include/IndustryStandard/Tpm2Acpi.h
@@ -24,18 +24,24 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #define EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_5  16
 #define EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE             EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_5
 
-typedef struct {
-  EFI_ACPI_DESCRIPTION_HEADER    Header;
-  // Flags field is replaced in version 4 and above
-  //    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
-  //    BIT16~31: Reserved
-  UINT32                         Flags;
-  UINT64                         AddressOfControlArea;
+// MU_CHANGE - [BEGIN]
+
+// Common fields shared across all TPM2 ACPI table revisions.
+// Flags field is replaced in version 4 and above:
+//    BIT0~15:  PlatformClass      This field is only valid for version 4 and above
+//    BIT16~31: Reserved
+//
+#define EFI_TPM2_ACPI_TABLE_COMMON_FIELDS \
+  EFI_ACPI_DESCRIPTION_HEADER    Header; \
+  UINT32                         Flags; \
+  UINT64                         AddressOfControlArea; \
   UINT32                         StartMethod;
-  // UINT8                       PlatformSpecificParameters[];  // size up to 16
-  // UINT32                      Laml;                          // Optional
-  // UINT64                      Lasa;                          // Optional
+
+typedef struct {
+  EFI_TPM2_ACPI_TABLE_COMMON_FIELDS
 } EFI_TPM2_ACPI_TABLE;
+
+// MU_CHANGE - [END]
 
 #define EFI_TPM2_ACPI_TABLE_START_METHOD_ACPI                                         2
 #define EFI_TPM2_ACPI_TABLE_START_METHOD_TIS                                          6
@@ -64,7 +70,8 @@ typedef struct {
   UINT32    Interrupt;
   UINT8     Flags;
   UINT8     OperationFlags;
-  UINT8     Reserved[2];
+  UINT8     Attributes; // MU_CHANGE
+  UINT8     Reserved;   // MU_CHANGE
   UINT32    SmcFunctionId;
 } EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_SMC;
 
@@ -79,6 +86,48 @@ typedef struct {
   UINT16    PartitionId;
   UINT8     Reserved[8];
 } EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA;
+
+// MU_CHANGE - [BEGIN]
+
+typedef struct {
+  EFI_TPM2_ACPI_TABLE_COMMON_FIELDS
+
+  // StartMethodSpecificParameters is variable in size and LAML/LASA are
+  // optional fields. It is the user's responsibility to access the
+  // Header.Length field to determine what is accessible in the table.
+  union {
+    UINT8                                                     PlatformSpecificParameters[EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_4];
+    EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_SMC    SmcParameters;
+  } StartMethodSpecificParameters;
+
+  UINT32    Laml; // Optional
+  UINT64    Lasa; // Optional
+} EFI_TPM2_ACPI_TABLE_V4;
+
+typedef struct {
+  EFI_TPM2_ACPI_TABLE_COMMON_FIELDS
+
+  // StartMethodSpecificParameters is variable in size and LAML/LASA are
+  // optional fields. It is the user's responsibility to access the
+  // Header.Length field to determine what is accessible in the table.
+  union {
+    UINT8                                                     PlatformSpecificParameters[EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_5];
+    EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_SMC    SmcParameters;
+    EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA    FfaParameters;
+  } StartMethodSpecificParameters;
+
+  UINT32    Laml; // Optional
+  UINT64    Lasa; // Optional
+} EFI_TPM2_ACPI_TABLE_V5;
+
+typedef struct {
+  EFI_TPM2_ACPI_TABLE_COMMON_FIELDS
+  UINT8     PlatformSpecificParameters[EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE];
+  UINT32    Laml; // Optional
+  UINT64    Lasa; // Optional
+} EFI_TPM2_ACPI_TABLE_TEMPLATE;
+
+// MU_CHANGE - [END]
 
 #define EFI_TPM2_ACPI_TABLE_ARM_FFA_PARAMETER_FLAG_NOTIFICATION_SUPPORT  BIT0
 
--- a/SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
+++ b/SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
@@ -1,0 +1,20 @@
+/** @file
+  Defines the GUID used to signal that the TCG event log has been dynamically
+  scaled. Consumers may register a notification callback on this event group
+  to react to the scaling event.
+
+  Copyright (c), Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef TCG2_EVENT_LOG_SCALED_H_
+#define TCG2_EVENT_LOG_SCALED_H_
+
+#define TCG2_EVENT_LOG_SCALED_GUID \
+  { \
+    0x9b4f7c2a, 0x1d3e, 0x4a8b, { 0x9c, 0x6f, 0x5e, 0x8d, 0x2a, 0x1b, 0x4c, 0x7f } \
+  }
+
+extern EFI_GUID  gTcg2EventLogScaledGuid;
+
+#endif // TCG2_EVENT_LOG_SCALED_H_
--- a/SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c
+++ b/SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c
@@ -74,7 +74,11 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 //
 #define MAX_PRS_INT_BUF_SIZE  (15*4)
 
-#pragma pack(1)
+// MU_CHANGE - [BEGIN]
+
+#if 0
+
+  #pragma pack(1)
 
 typedef struct {
   EFI_ACPI_DESCRIPTION_HEADER    Header;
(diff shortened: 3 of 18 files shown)


MdePkg/Include/IndustryStandard/Tpm2Acpi.h
```diff
@@ -80,6 +87,41 @@ typedef struct {
...
 } EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA;
 
+// MU_CHANGE - [BEGIN]
+
+typedef struct {
+  EFI_TPM2_ACPI_TABLE_COMMON_FIELDS
+  union {
+    UINT8                                                     PlatformSpecificParameters[EFI_TPM2_ACPI_TABLE_START_METHOD_SPECIFIC_PARAMETERS_MAX_SIZE_REVISION_4];
+    EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_SMC    SmcParameters;
+    EFI_TPM2_ACPI_START_METHOD_SPECIFIC_PARAMETERS_ARM_FFA    FfaParameters;
```
cfernald: isn't FF-A only in version 5+?

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
```diff
@@ -80,6 +87,41 @@ typedef struct {
...
+  } StartMethodSpecificParameters;
+
+  UINT32    Laml; // Optional
```
cfernald, line 103: I'm not against it exactly, but if these arre option then that probably means that the ACPI table header size may or may not be large enough to include them. Additionally, I suspect that the array above may not be fully included either since it is "up to" that size. Though it's a pain, that is a reason leaving them as a comment can actually be safer because the caller cannot assume from the struct they are present. Maybe a least add a comment about size checks.

cfernald (approved):

SecurityPkg/Tcg/Tcg2Acpi/Tcg2Acpi.c
```diff
@@ -810,13 +806,14 @@ PublishTpm2 (
...
+      sizeof (EFI_TPM2_ACPI_TABLE_TEMPLATE)
       );
   }
 
   // MU_CHANGE [END]
 
-  mTpm2AcpiTemplate.Header.Revision = PcdGet8 (PcdTpm2AcpiTableRev);
+  mTpm2AcpiTemplate.Header.Signature = EFI_ACPI_5_0_TRUSTED_COMPUTING_PLATFORM_2_TABLE_SIGNATURE; // MU_CHANGE
+  mTpm2AcpiTemplate.Header.Revision  = PcdGet8 (PcdTpm2AcpiTableRev); // MU_CHANGE
   DEBUG ((DEBUG_INFO, "Tpm2 ACPI table revision is %d\n", mTpm2AcpiTemplate.Header.Revision));
```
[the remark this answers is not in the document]
Raymond-MS: The spec mentions that the ACPI table must be measured before any modifications. If the table was measured after modifications, it is possible that the values would be inconsistent per boot which would make the PCR measurement unstable which goes against the point of the measurement. Won't fix.

SecurityPkg/Tcg/Tcg2AcpiFfa/Tcg2AcpiFfa.c
```diff
@@ -350,60 +344,70 @@ PublishTpm2 (
...
+      &mTpm2AcpiTemplate,
+      sizeof (EFI_TPM2_ACPI_TABLE_TEMPLATE)
+      );
+  }
 
-  mTpm2AcpiTemplate.Header.Revision = PcdGet8 (PcdTpm2AcpiTableRev);
+  mTpm2AcpiTemplate.Header.Signature = EFI_ACPI_5_0_TRUSTED_COMPUTING_PLATFORM_2_TABLE_SIGNATURE;
+  mTpm2AcpiTemplate.Header.Length    = sizeof (EFI_TPM2_ACPI_TABLE_V5);
+  mTpm2AcpiTemplate.Header.Revision  = PcdGet8 (PcdTpm2AcpiTableRev);
   DEBUG ((DEBUG_INFO, "Tpm2 ACPI table revision is %d\n", mTpm2AcpiTemplate.Header.Revision));
```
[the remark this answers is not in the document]
Raymond-MS: I answered this in the previous comment relating to this. Won't fix.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
Raymond-MS: Good catch, updated the V4 struct.

MdePkg/Include/IndustryStandard/Tpm2Acpi.h
Raymond-MS: Will add a comment detailing this so it is less confusing.

Flickdm (approved):

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1133,6 +1294,158 @@ CopyDigestListBinToBuffer (
...
+  if (EventLogAreaStruct == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  // If the log was scaled the maximum number of times, mark it as truncated.
+  if (EventLogAreaStruct->ScaleCount >= TCG_EVENT_LOG_MAX_SCALE_COUNT) {
+    DEBUG ((DEBUG_ERROR, "%a: Scale limit reached (%u), truncating log\n", __func__, TCG_EVENT_LOG_MAX_SCALE_COUNT));
+    EventLogAreaStruct->EventLogTruncated = TRUE;
+    return EFI_OUT_OF_RESOURCES;
+  }
```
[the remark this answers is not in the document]
Raymond-MS: Good catch, updated.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -28,6 +30,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Protocol/MuTcg2Protocol.h> // MU_CHANGE - Add a new protocol to support Log-only events.
 #include <Protocol/TrEEProtocol.h>
 #include <Protocol/ResetNotification.h>
+#include <Protocol/AcpiTable.h>                  // MU_CHANGE
+#include <Protocol/AcpiSystemDescriptionTable.h> // MU_CHANGE
```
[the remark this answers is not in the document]
Raymond-MS: Good catch, removed.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.inf
```diff
@@ -107,6 +108,8 @@
   gEfiMpServiceProtocolGuid                          ## SOMETIMES_CONSUMES
   gEfiVariableWriteArchProtocolGuid                  ## NOTIFY
   gEfiResetNotificationProtocolGuid                  ## CONSUMES
+  gEfiAcpiTableProtocolGuid                          ## SOMETIMES_CONSUMES  # MU_CHANGE
+  gEfiAcpiSdtProtocolGuid                            ## SOMETIMES_CONSUMES  # MU_CHANGE
```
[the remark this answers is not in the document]
Raymond-MS: Good catch, removed.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1043,6 +1196,10 @@ TcgDxeLogEvent (
   return Status;
 }
 
+// MU_CHANGE - [BEGIN]
+
+#if 0
+
 /**
   Get TPML_DIGEST_VALUES compact binary buffer size.
```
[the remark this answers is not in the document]
Raymond-MS: This is necessary for tracking MU_CHANGES. I am unable to just remove the code. This leaves some indicator of where the code existed such that it is easier to track what happened. Won't fix. Will get removed when upstreamed.

SecurityPkg/Tcg/TcgLogTest/TcgLogTestDxe.c
```diff
@@ -0,0 +1,258 @@
...
+  if (!IsTestEnabled ()) {
+    DEBUG ((DEBUG_INFO, "%a: Test not enabled, skipping\n", __func__));
+    return EFI_SUCCESS;
+  }
+
+  DEBUG ((DEBUG_INFO, "%a: Test enabled, running pre-ReadyToBoot test\n", __func__));
+
+  // Clear the enable variable so we don't re-run on the next boot.
+  TcgLogTestEnable (&mTcgLogTestProtocol, FALSE);
+
```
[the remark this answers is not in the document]
Raymond-MS: Good catch, updated the logging.

cfernald (approved):

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1133,6 +1292,159 @@ CopyDigestListBinToBuffer (
...
+    EventLogAreaStruct->EventLogTruncated = TRUE;
+    return EFI_VOLUME_FULL;
+  }
+
+  NewLaml = EventLogAreaStruct->Laml * 2;
+  if (NewLaml <= EventLogAreaStruct->Laml) {
+    DEBUG ((DEBUG_ERROR, "%a: Laml overflow (0x%lx * 2)\n", __func__, EventLogAreaStruct->Laml));
+    EventLogAreaStruct->EventLogTruncated = TRUE;
+    return EFI_OUT_OF_RESOURCES;
+  }
```
makubacki, line 1338: I was intuitively expecting this to double and check against `TcgLogDynamicScalingNeeded()` `TCG_EVENT_LOG_MAX_SCALE_COUNT` number of times. For example, if the doubled size is not large enough and the max scale count has not been reached, I would expect the scale to automatically double again here.

Am I misunderstanding or is there a reason that's not done?

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -1905,9 +2242,11 @@ SetupEventLog (
...
         (mTcgDxeData.FinalEventsTable[Index])->NumberOfEvents = 0;
 
-        mTcgDxeData.FinalEventLogAreaStruct[Index].EventLogFormat        = mTcg2EventInfo[Index].LogFormat;
-        mTcgDxeData.FinalEventLogAreaStruct[Index].Lasa                  = Lasa + sizeof (EFI_TCG2_FINAL_EVENTS_TABLE);
-        mTcgDxeData.FinalEventLogAreaStruct[Index].Laml                  = PcdGet32 (PcdTcg2FinalLogAreaLen) - sizeof (EFI_TCG2_FINAL_EVENTS_TABLE);
+        mTcgDxeData.FinalEventLogAreaStruct[Index].EventLogFormat = mTcg2EventInfo[Index].LogFormat;
+        mTcgDxeData.FinalEventLogAreaStruct[Index].Lasa           = Lasa + sizeof (EFI_TCG2_FINAL_EVENTS_TABLE);
+        // MU_CHANGE
+        // Reserve space in the FinalEventLog for the EV_NO_ACTION truncation marker event.
+        mTcgDxeData.FinalEventLogAreaStruct[Index].Laml                  = PcdGet32 (PcdTcg2FinalLogAreaLen) - sizeof (EFI_TCG2_FINAL_EVENTS_TABLE) - GetTruncationEventSize ();
```
makubacki, line 2263: Given the PCD can be configured per-platform, you should verify the subtraction won't underflow first (e.g. `PcdGet32 (PcdTcg2FinalLogAreaLen) >= (sizeof (EFI_TCG2_FINAL_EVENTS_TABLE) + GetTruncationEventSize ()`).

SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
```diff
@@ -0,0 +1,20 @@
+/** @file
+  Defines the GUID used by Tcg2Dxe to signal that the TCG event log has been
+  dynamically scaled. Consumers may register a notification callback on this
+  event group to react to the scaling event.
```
makubacki: nit: It's best not to couple implementation details in other modules with generic public headers like this. You would probably remember to come back and update this in a refactor but someone else would not. The same point can be made without implementation specifics.


```suggestion
  Defines the GUID used to signal that the TCG event log has been dynamically
  scaled. Consumers may register a notification callback on this event group
  to react to the scaling event.
```

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -942,6 +960,130 @@ TcgCommLogEvent (
...
+}
+
+/**
+  Compute the total size in bytes of the EV_NO_ACTION truncation marker event
+  (header for the active PCR banks + truncation payload).
+
+  Used to reserve headroom in the Final Events log so the marker is always
+  guaranteed to fit when truncation occurs.
+
+  @retval  Size in bytes of the truncation marker event.
```
makubacki: nit: [`retval`](https://www.doxygen.nl/manual/commands.html#cmdretval) is technically used for a return value with a name and [`return`](https://www.doxygen.nl/manual/commands.html#cmdreturn) for one without a name.


```suggestion
  @return  Size in bytes of the truncation marker event.
```

makubacki (approved):
I didn't look super closely at the test code.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: There is no good reason. This should normally never happen unless the PCD was set so small that a log happens to be larger than the double length. With a default size of 0x10000 this should never happen. However, I updated the code to handle this situation such that we don't fail if this happens and instead handle it somewhat gracefully.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Similar to scaling multiple times at once, this shouldn't really happen if the PCD is set properly. However, this is still a potential issue. Good catch. I will update to take this into account.

SecurityPkg/Include/Guid/Tcg2EventLogScaled.h
Raymond-MS: Sounds good, updated.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
Raymond-MS: Fixed.

cfernald (approved):

PR #1812, opened 2026-06-04
[REBASE & FF] Add per-streamId isolation for SmmuDxe

eeshanl:
### Add per-StreamID isolation for SmmuDxe

Replaces SmmuDxe's previous "shared page table root" model with **per-StreamID stage-2 isolation** in `ArmPkg` and adds the supporting plumbing in `MdeModulePkg` and `EmbeddedPkg` so non-PCIe and handle-less DMA agents can participate.

Every DMA-capable device gets its own stage-2 page-table root and VMID, mappings made for one device are invisible to any other, and translation related faults are surfaced immediately via GIC EVTQ/GERR ISRs.

#### Commits

1. **Revert "Add `gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid` ..."** - superseded by a `RegisterNonDiscoverableMmioDevice` parameter in the next commit.
2. **`MdeModulePkg`: Add `UniqueId` to `NonDiscoverableDeviceRegistrationLib`** - `RegisterNonDiscoverableMmioDevice` now takes a platform `UniqueId`, surfaced as `PciIo->GetLocation()` = `(Seg=0xFF, Bus=UniqueId>>5, Dev=UniqueId&0x1F, Func=0)`. **Breaking change.**
3. **`MdeModulePkg`: Add `SetAttributeById` to IoMmu Protocol** - `(IommuBase, DmaId, Mapping, IoMmuAccess)` entry point for handle-less callers. Protocol revision bumped to `0x00010001`. `IoMmuLib` NULL-checks; `IoMmuLibNull` is a no-op.
4. **`MdeModulePkg`: Add `Handle` to `NON_DISCOVERABLE_PCI_DEVICE`** - passes the controller handle (instead of `NULL`) to `IoMmuSetAttribute`, so the IOMMU producer can resolve the StreamID via `HandleProtocol → PciIo → GetLocation → IORT`.
5. **`EmbeddedPkg/DmaLib`: Use `IoMmuSetAttributeById` with `IommuBase`/`DmaId`**  - `DmaMap` takes two new params, forwarded to `SetAttributeById`. `IommuBase == 0` is the explicit opt-out. **Breaking change.**
6. **`ArmPkg/SmmuDxe`: Add per-StreamID isolation for each DMA capable device** -
   - `SmmuStreamContext` holds `{PageTableRoot, Vmid}` per StreamID per SMMU; both allocated lazily and the STE promoted `INVALID → STAGE_2_TRANSLATE` via break-before-make on first `SetAttribute*`.
   - Aliases share the primary's root + VMID one page-table update covers DMA from all StreamIDs of one logical device.
   - `DeviceHandleToStreamId`: real PCIe → IORT RC; `Seg=0xFF` → platform `SMMU_NC_DEVICE_ENTRY` HOB → IORT Named Component.
   - GIC EVTQ/GERR ISRs surface faults immediately; opportunistic drain on every `SetAttribute*`.
   - ExitBootServices is per-SMMU: bypass for SMMUs with RMR streams, abort for the rest.
   - `SMMU_CONFIG` gains `NcDeviceListSize/Offset`; new `SMMU_NC_DEVICE_ENTRY {UniqueId, ObjName[32]}` maps platform UniqueIds → IORT NC names.

- [ ] Impacts functionality?
- [x] Impacts security?
- [x] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

## How This Was Tested

Tested on SBSA (https://github.com/OpenDevicePartnership/patina-qemu/pull/261), QemuArmVirtPkg, and physical arm platform.

## Integration Instructions

1. **Build the IORT.** Include one SMMUv3 node, the necessary ITS / Root Complex nodes for PCIe, and one `EFI_ACPI_IORT_TYPE_NAMED_COMP` node per non-discoverable DMA-capable device. Each Named Component's ID mappings must produce the StreamIDs the device emits and reference the owning SMMUv3 node.
2. **(Optional) Register any non-PCI/NonDiscoverable devices with a stable `UniqueId`** via `RegisterNonDiscoverableMmioDevice (UniqueId, ...)`. SmmuDxe reconstructs that `UniqueId` from `PciIo->GetLocation()` later, so values must be unique across the platform.
3. **Build the `SMMU_CONFIG` HOB** in your platform `SecPlatformSmmuConfigLib`. Append (in any order) the IORT blob, the optional NC device lookup table for NonDiscoverable devices, and the optional SMMU disable list, then point the offsets/sizes in the `SMMU_CONFIG` header at them:

   ```c
   STATIC CONST SMMU_NC_DEVICE_ENTRY  NcDeviceTable[] = {
     // { UniqueId, IORT NC ObjectName }
     { 0x1, "USB"       },
     { 0x2, "SATA_AHCI" },
   };

   SmmuConfig = AllocateZeroPool (sizeof (SMMU_CONFIG)
                                  + sizeof (IortData)
                                  + sizeof (NcDeviceTable));
   SmmuConfig->VersionMajor       = CURRENT_SMMU_CONFIG_VERSION_MAJOR;
   SmmuConfig->VersionMinor       = CURRENT_SMMU_CONFIG_VERSION_MINOR;
   SmmuConfig->IortSize           = sizeof (IortData);
   SmmuConfig->IortOffset         = sizeof (SMMU_CONFIG);
   SmmuConfig->NcDeviceListSize   = sizeof (NcDeviceTable);
   SmmuConfig->NcDeviceListOffset = sizeof (SMMU_CONFIG) + sizeof (IortData);

   CopyMem ((UINT8 *)SmmuConfig + SmmuConfig->IortOffset,
            &IortData,    sizeof (IortData));
   CopyMem ((UINT8 *)SmmuConfig + SmmuConfig->NcDeviceListOffset,
            NcDeviceTable, sizeof (NcDeviceTable));

   BuildGuidDataHob (&gSmmuConfigHobGuid,
                     SmmuConfig,
                     sizeof (SMMU_CONFIG)
                     + sizeof (IortData)
                     + sizeof (NcDeviceTable));
   ```

4. **Do not install the IORT yourself** SmmuDxe installs it as an ACPI table after parsing.
5. **Disable specific SMMUs (optional).** Append a `UINT64[]` of SMMU base addresses to put in bypass mode and point `SmmuDisabledListOffset` / `SmmuDisabledListSize` at it.
6. **Handle-less DMA agents (optional).** Firmware components that have no UEFI device handle (and therefore are not in the IORT) call `IoMmuLib::IoMmuSetAttributeById (IommuBase, DmaId, ...)` directly. Pass `IommuBase == 0` to opt out of IOMMU programming entirely.

See [`ArmPkg/Drivers/SmmuDxe/README.md`](ArmPkg/Drivers/SmmuDxe/README.md) for the end-to-end flow diagrams, HOB layout, and the per-StreamID isolation walkthrough.

#### Breaking changes

- **`RegisterNonDiscoverableMmioDevice` gains a leading `UniqueId` parameter.** All existing call sites must be updated to pass a value that is unique across the platform.
- **`gEdkiiNonDiscoverableDeviceUniqueIdProtocolGuid` is removed.** Any out-of-tree consumer of that protocol must switch to the new `UniqueId` parameter on `RegisterNonDiscoverableMmioDevice`.
- **`DmaMap` gains `IommuBase` + `DmaId` parameters.** Pass `0/0` to keep pre-existing behaviour (no IOMMU programming); supply real values to opt into per-StreamID isolation for handle-less DMA agents.
- **`EDKII_IOMMU_PROTOCOL_REVISION` bumped to `0x00010001`.** The new `SetAttributeById` field is optional. Legacy producers leave it `NULL`, and consumers that call it through `IoMmuLib::IoMmuSetAttributeById` get `EFI_UNSUPPORTED` and must gracefully fall back.


MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
```diff
@@ -10,9 +10,6 @@
 
 #include <Protocol/DriverBinding.h>
 
-#define MAX_NON_DISCOVERABLE_PCI_DEVICE_ID  (32 * 256)
```
Raymond-MS: Removal of this code should have MU_CHANGE tags and not be fully removed.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
```diff
@@ -148,11 +145,6 @@ NonDiscoverablePciDeviceStart (
   NON_DISCOVERABLE_PCI_DEVICE  *Dev;
   EFI_STATUS                   Status;
 
-  ASSERT (mUniqueIdCounter < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID);
```
Raymond-MS: Same here with MU_CHANGE tags.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
```diff
@@ -187,8 +183,6 @@ NonDiscoverablePciDeviceStart (
     goto CloseProtocol;
   }
 
-  Dev->UniqueId = mUniqueIdCounter++;
```
Raymond-MS: Same here with MU_CHANGE tags.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -127,4 +127,32 @@ IoMmuSetAttribute (
   IN UINT64      IoMmuAccess
   );
 
+/**
```
Raymond-MS, line 130: This should contain MU_CHANGE tags

os-d:
@eeshanl I see this is in draft, can you let me know when it is ready to review?

MdeModulePkg/Library/IoMmuLib/IoMmuLib.c
```diff
@@ -180,6 +180,45 @@ IoMmuSetAttribute (
   return mIoMmuProtocol->SetAttribute (mIoMmuProtocol, DeviceHandle, Mapping, IoMmuAccess);
 }
 
+/**
```
Raymond-MS, line 187: This should have MU_CHANGE tags.

MdeModulePkg/Library/IoMmuLib/IoMmuLib.c
```diff
@@ -252,5 +291,12 @@ IoMmuLibInit (
     }
   }
 
+  // Check if version is atleast EDKII_IOMMU_PROTOCOL_REVISION, if not, return unsupported
```
Raymond-MS: This should have MU_CHANGE tags.

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
```diff
@@ -131,3 +131,25 @@ IoMmuSetAttribute (
 {
   return EFI_SUCCESS;
 }
+
```
Raymond-MS, line 134: This should have MU_CHANGE tags.

EmbeddedPkg/Drivers/NonCoherentIoMmuDxe/NonCoherentIoMmuDxe.c
```diff
@@ -142,6 +142,8 @@ NonCoherentIoMmuMap (
            DmaOperation,
            HostAddress,
            NumberOfBytes,
+           0,
```
Raymond-MS: This should have MU_CHANGE tags

EmbeddedPkg/Include/Library/DmaLib.h
```diff
@@ -60,6 +60,10 @@ typedef enum {
   @param  HostAddress           The system memory address to map to the DMA controller.
   @param  NumberOfBytes         On input the number of bytes to map. On output the number of bytes
                                 that were mapped.
+  @param  IommuBase             Base MMIO address of the IOMMU that owns the DMA agent's DmaId.
```
Raymond-MS, line 64: This should have MU_CHANGE tags

EmbeddedPkg/Library/CoherentDmaLib/CoherentDmaLib.c
```diff
@@ -59,6 +70,7 @@ DmaMap (
   VOID                   *IoMmuHostAddress;
   EDKII_IOMMU_OPERATION  IoMmuOperation;
   UINT64                 IoMmuAttribute;
+  COHERENT_MAP_INFO      *Map;
```
Raymond-MS, line 73: This should have a MU_CHANGE tag

EmbeddedPkg/Library/CoherentDmaLib/CoherentDmaLib.c
```diff
@@ -95,6 +114,7 @@ DmaMap (
     default:
       DEBUG ((DEBUG_ERROR, "%a - Invalid operation %d\n", __func__, Operation));
       ASSERT (FALSE);
+      FreePool (Map); // MU_CHANGE
```
Raymond-MS: This tag is within a block MU_CHANGE

EmbeddedPkg/Library/CoherentDmaLib/CoherentDmaLib.c
```diff
@@ -103,36 +123,43 @@ DmaMap (
              IoMmuHostAddress,
              NumberOfBytes,
              DeviceAddress,
-             Mapping
+             &Map->IoMmuContext // MU_CHANGE
```
Raymond-MS: Same here.

EmbeddedPkg/Library/CoherentDmaLib/CoherentDmaLib.c
```diff
@@ -103,36 +123,43 @@ DmaMap (
              IoMmuHostAddress,
              NumberOfBytes,
              DeviceAddress,
-             Mapping
+             &Map->IoMmuContext // MU_CHANGE
              );
   if (EFI_ERROR (Status)) {
     DEBUG ((DEBUG_ERROR, "%a - IoMmuMap failed.\n", __func__));
     ASSERT (FALSE);
+    FreePool (Map); // MU_CHANGE
```
Raymond-MS: Same here.

EmbeddedPkg/Library/NonCoherentDmaLib/NonCoherentDmaLib.c
```diff
@@ -392,13 +398,14 @@ DmaMap (
     goto FreeMapInfo;
   }
 
-  Status = IoMmuSetAttribute (
-             NULL,
+  Status = IoMmuSetAttributeEx (
```
Raymond-MS: This section needs a MU_CHANGE tag

eeshanl:
> @eeshanl I see this is in draft, can you let me know when it is ready to review?

@os-d Feel free to review if you would like. It is more-or-less ready for review, just doing a last pass through to clean up the code.

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: fixed commented out removed lines and added MU_CHANGE tags

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: fixed, commented out removed lines and added MU_CHANGE tags

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: done

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: not adding here since this file already does not exist in edk2 yet

MdeModulePkg/Library/IoMmuLib/IoMmuLib.c
eeshanl: not adding here since this file already does not exist in edk2 yet

MdeModulePkg/Library/IoMmuLib/IoMmuLib.c
eeshanl: not adding here since this file already does not exist in edk2 yet

MdeModulePkg/Library/IoMmuLibNull/IoMmuLibNull.c
eeshanl: not adding here since this file already does not exist in edk2 yet

EmbeddedPkg/Include/Library/DmaLib.h
eeshanl: done

EmbeddedPkg/Library/CoherentDmaLib/CoherentDmaLib.c
eeshanl: done

EmbeddedPkg/Library/NonCoherentDmaLib/NonCoherentDmaLib.c
eeshanl: done

EmbeddedPkg/Drivers/NonCoherentIoMmuDxe/NonCoherentIoMmuDxe.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+  @param [out]  PageTableRoot   Receives the root.
+  @param [out]  OutCtx          Receives the per-stream context
+                                (PageTableRoot + VMID) for this StreamID.
+
+  @retval EFI_SUCCESS            Success.
+  @retval EFI_OUT_OF_RESOURCES   Allocation failed.
+  @retval Other                  STE promotion failure.
+**/
+EFI_STATUS
+SmmuStreamContextEnsurePageTable (
```
os-d: nit: consider having the same naming as your earlier fn (GetOrCreate)

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+    return EFI_SUCCESS;
+  }
+
+  // Allocate a per-stream VMID if this context doesn't have one yet.
+  // 0 is reserved as "unassigned". The width depends on IDR0.VMID16:
+  // 8-bit -> 1..0xFF, 16-bit -> 1..0xFFFF.
+  if (Ctx->Vmid == 0) {
+    UINT16  MaxVmid;
+
+    MaxVmid = SmmuInfo->Vmid16Supported ? 0xFFFF : 0xFF;
```
os-d: nit: MAX_UINT[8|16]

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+  }
+
+  Status = SmmuStreamContextGetOrCreate (SmmuInfo, StreamId, &Ctx);
+  if (EFI_ERROR (Status)) {
+    return Status;
+  }
+
+  if (Ctx->PageTableRoot != NULL) {
+    *PageTableRoot = Ctx->PageTableRoot;
+    *OutCtx        = Ctx;
```
os-d: nit: If PageTableRoot is a member of the ctx struct, what is the value of returning PageTableRoot separately?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+  Ctx->PageTableRoot = NewRoot;
+  *PageTableRoot     = NewRoot;
+  *OutCtx            = Ctx;
+  return EFI_SUCCESS;
+}
+
+/**
+  Bind a StreamID's context to share an existing primary stream's stage-2
+  page-table root and VMID. Used when a single device exposes multiple
+  StreamIDs  that must all see the same translations.
```
os-d:
```suggestion
  StreamIDs that must all see the same translations.
```

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+  if ((AliasCtx->PageTableRoot == PrimaryCtx->PageTableRoot) &&
+      (AliasCtx->Vmid == PrimaryCtx->Vmid))
+  {
+    return EFI_SUCCESS;
+  }
+
+  // If this alias context had previously been promoted with its own root,
+  // that's a configuration error: the same StreamID cannot legitimately
+  // serve as a primary and an alias at the same time.
+  if ((AliasCtx->PageTableRoot != NULL) &&
```
os-d: Does same page table root and different vmid matter? I think probably not

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+  AliasCtx->Vmid          = PrimaryCtx->Vmid;
+  AliasCtx->PageTableRoot = PrimaryCtx->PageTableRoot;
+
+  Status = SmmuV3PromoteSteToTranslate (
+             SmmuInfo,
+             AliasStreamId,
+             PrimaryCtx->Vmid,
+             PrimaryCtx->PageTableRoot
+             );
+  if (EFI_ERROR (Status)) {
```
os-d, line 285: Do you want to add some asserts in some of these failure cases so they are easier to catch?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -92,17 +328,19 @@ UpdateMapping (
       }
 
       ZeroMem ((VOID *)NewPage, EFI_PAGE_SIZE);
+      ArmDataSynchronizationBarrier ();
```
os-d, line 379: I think this one should go after the bits are written to the entry and just before it is written into the table. We want to ensure that all writes are complete before we install it into the page table

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -92,17 +328,19 @@ UpdateMapping (
       }
 
       ZeroMem ((VOID *)NewPage, EFI_PAGE_SIZE);
+      ArmDataSynchronizationBarrier ();
 
       Entry                   = (PAGE_TABLE_ENTRY)(UINTN)NewPage | PAGE_TABLE_ACCESS_FLAG | PAGE_TABLE_DESCRIPTOR | PAGE_TABLE_ENTRY_VALID_BIT;
       Current->Entries[Index] = Entry; // valid entry
+      ArmDataSynchronizationBarrier ();
```
os-d: I think this one is not needed because we are not at the leaf level, so the translation is still invalid. We're going to then update the leaf level and that's where we want to make sure the valid translation has completed before exiting out. @cfernald ?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -114,8 +352,11 @@ UpdateMapping (
       Entry                  |= Flags | PAGE_TABLE_ACCESS_FLAG | PAGE_TABLE_DESCRIPTOR | PAGE_TABLE_ENTRY_VALID_BIT;
       Current->Entries[Index] =  Entry;
     } else {
-      Current->Entries[Index] = Current->Entries[Index] & ~PAGE_TABLE_ENTRY_VALID_BIT; // only invalidate leaf entry
+      Entry                   = Current->Entries[Index] & ~PAGE_TABLE_ENTRY_VALID_BIT;
+      Current->Entries[Index] = Entry; // only invalidate leaf entry
     }
+
+    ArmDataSynchronizationBarrier ();
```
os-d: This one may need to be updated to follow what the MMU updating does: https://github.com/tianocore/edk2/blob/1f8832058746a4832bcda9f844a9ef718b72d18d/UefiCpuPkg/Library/ArmMmuLib/AArch64/ArmMmuLibReplaceEntry.S#L52-L68.

- Write invalid entry (break)
- Data barrier to ensure entry is finished writing invalid
- Flush TLB for VA
- Data barrier to ensure that completes (this is different since you are sending to SMMU)
- Write updated entry
- DSB + ISB (make sure new entry is seen and speculative execution doesn't pull something incorrect)

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -160,33 +405,27 @@ UpdatePageTable (
...
       DEBUG ((DEBUG_ERROR, "%a: Failed to update page table mapping\n", __func__));
+      gBS->RestoreTPL (OldTpl);
       goto End;
     }
 
-    CurPhysicalAddress += EFI_PAGE_SIZE;
-  }
+    gBS->RestoreTPL (OldTpl);
+
+    ArmDataSynchronizationBarrier ();
```
os-d: This would be dropped if we change the ordering as mentioned above.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -160,33 +405,27 @@ UpdatePageTable (
...
+      if (EFI_ERROR (Status)) {
+        DEBUG ((DEBUG_ERROR, "%a: Failed to invalidate TLB for address 0x%llx\n", __func__, CurPhysicalAddress));
+        goto End;
       }
     }
+
+    CurPhysicalAddress += EFI_PAGE_SIZE;
   }
 
   SpeculationBarrier ();
```
os-d: This would then also not be needed

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
```diff
@@ -148,10 +152,13 @@ NonDiscoverablePciDeviceStart (
...
   EFI_STATUS                   Status;
 
-  ASSERT (mUniqueIdCounter < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID);
-  if (mUniqueIdCounter >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID) {
-    return EFI_OUT_OF_RESOURCES;
-  }
+  // MU_CHANGE [BEGIN]: mUniqueIdCounter limit check no longer needed; the
+  // platform supplies UniqueId via NON_DISCOVERABLE_DEVICE.
+  // ASSERT (mUniqueIdCounter < MAX_NON_DISCOVERABLE_PCI_DEVICE_ID);
+  // if (mUniqueIdCounter >= MAX_NON_DISCOVERABLE_PCI_DEVICE_ID) {
```
os-d, line 165: Should a different unique check occur now to ensure platform misconfiguration didn't occur? I'm not sure if this driver keeps a list of registered devices.

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -127,4 +127,32 @@ IoMmuSetAttribute (
...
+  Set the R/W access attributes for Mapping in the Page Table for a caller
+  that explicitly specifies (IommuBase, DmaId) instead of an EFI_HANDLE.
+
+  Intended for firmware-internal DMA agents that have no UEFI device
+  handle (and therefore cannot be resolved via the IORT). The caller is
+  responsible for providing the correct IommuBase and DmaId.
+
+  @param [in]  IommuBase     Base MMIO address of the IOMMU that owns DmaId.
+  @param [in]  DmaId     DMA identifier emitted by the calling DMA agent (e.g. StreamID on Arm SMMU, RequesterID on VT-d).
```
os-d: nit: match spacing

MdeModulePkg/Include/Library/IoMmuLib.h
```diff
@@ -127,4 +127,32 @@ IoMmuSetAttribute (
...
+
+  @retval EFI_SUCCESS            Success.
+  @retval EFI_NOT_READY          The IoMmu protocol is not ready.
+  @retval EFI_UNSUPPORTED        The IoMmu protocol does not implement
+                                 SetAttributeEx.
+  @retval Other                  Other errors as defined by the IoMmu protocol.
+**/
+EFI_STATUS
+EFIAPI
+IoMmuSetAttributeEx (
```
os-d: nit: edk2 doesn't typically use the Ex naming, consider IoMmuSetAttributeForId or something.

os-d (commented):
I did not finish my review, but I am out Mon/Tues, so I wanted to leave the first set of comments

ArmPkg/Drivers/SmmuDxe/IoMmu.c
cfernald: I think this is correct, because we are about to add this page to the structure but haven't yet put any valid entry in it, we need to be sure that the memory zeroing occurs before its put in the structure.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
cfernald: I agree. I don't think this one is required.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
cfernald: I don't think we actually need to follow break-before-make as it is formally defined here, since we never "change" an entry, but only add new entries or set existing block entries to invalid. So I believe we can finish all the operations before doing any TLBI or barriers. The only barriers I believe we need in updating the page tables are to prevent adding uninitialized data to the structure (such as after the zero) 

Image

cfernald: @eeshanl , if we intend on keeping this semantic, it might be worth adding a check that we never actually "change" either by bailing if already valid, or checking that the valid entry is identical, or both.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
os-d: Right, my point is that we then write the standalone entry after this, then install the page table. So might as well put the barrier after the write to the standalone entry, we then have ensured the zeroing and the writing has finished. The writing being finished needs to occur before DMA and that should be handled by barriers around the TLB invalidation later, but still, if we are going to put one here I would guarantee both before installing it.

EmbeddedPkg/Library/NonCoherentDmaLib/NonCoherentDmaLib.c
cfernald: I dont actually see these. Perhaps they are in the wrong commit.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,241 @@ typedef struct IOMMU_MAP_INFO {
...
+{
+  EFI_STATUS           Status;
+  SMMU_STREAM_CONTEXT  *Ctx;
+  PAGE_TABLE           *NewRoot;
+
+  if ((SmmuInfo == NULL) || (PageTableRoot == NULL) || (OutCtx == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status = SmmuStreamContextGetOrCreate (SmmuInfo, StreamId, &Ctx);
```
cfernald: Why is this secondary structure needed? Why not get this info from the stream table so there is a single source of truth? I think it could help simplify the overall flow as well.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -160,33 +405,27 @@ UpdatePageTable (
   PhysicalAddressEnd = ALIGN_VALUE (PhysicalAddress + Bytes, EFI_PAGE_SIZE);
 
   while (CurPhysicalAddress < PhysicalAddressEnd) {
-    Status = UpdateMapping (Root, CurPhysicalAddress, CurPhysicalAddress, Flags, Valid);
+    OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
```
cfernald: do this before the loop. As it stands you do this redundantly each iteration

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,256 @@ typedef struct IOMMU_MAP_INFO {
...
+    if (SmmuInfo->NextVmid == 0) {
+      // wrapped past the max
+      DEBUG ((DEBUG_ERROR, "%a: VMID space exhausted on SMMU 0x%llx\n", __func__, SmmuInfo->SmmuBase));
+      ASSERT (FALSE);
+      return EFI_OUT_OF_RESOURCES;
+    }
+
+    Ctx->Vmid = SmmuInfo->NextVmid;
+    if (SmmuInfo->NextVmid == MaxVmid) {
+      SmmuInfo->NextVmid = 0; // mark exhausted; next allocation will fail above
```
cfernald: Consider a #define for the exhausted value, for reasability.

EmbeddedPkg/Library/NonCoherentDmaLib/NonCoherentDmaLib.c
eeshanl: it actually already had a MU_CHANGE tag when i first introduced IoMmuLib to this file. So not needed to have a nested  set of MU_CHANGE tags

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done, good catch. removed extra PageTableRoot param.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: no same page table root and a different vmid would not matter

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,256 @@ typedef struct IOMMU_MAP_INFO {
...
+  Ctx = (SMMU_STREAM_CONTEXT *)AllocateZeroPool (sizeof (SMMU_STREAM_CONTEXT));
+  if (Ctx == NULL) {
+    DEBUG ((DEBUG_ERROR, "%a: Failed to allocate SMMU_STREAM_CONTEXT for SmmuBase=0x%llx StreamId 0x%x\n", __func__, SmmuInfo->SmmuBase, StreamId));
+    ASSERT (FALSE);
+    return EFI_OUT_OF_RESOURCES;
+  }
+
+  Ctx->StreamId      = StreamId;
+  Ctx->PageTableRoot = NULL;
+  InsertTailList (&SmmuInfo->StreamContextList, &Ctx->Link);
```
cfernald: Not sure if it's possible in practice, but it seems there is a potential race condition with these structures and interrupts, similar to the regular page table update.

Considering these are now per-stream-id, I doubt it will come up though, but consider at least checking for contention.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: removed

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: keeping this one per discussion with @cfernald

MdeModulePkg/Bus/Pci/NonDiscoverablePciDeviceDxe/NonDiscoverablePciDeviceDxe.c
eeshanl: Yes, i added in a check that will LocateHandleBuffer and loop through to check if any duplicates are found

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: done

MdeModulePkg/Include/Library/IoMmuLib.h
eeshanl: done, renamed to IoMmuSetAttributeById

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done, removed.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: yes good call, removed. i refer directly to the STE's for source of truth.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: We no longer have this context list, removed it. However I still need to review all of the changes for potential race conditions

ArmPkg/Drivers/SmmuDxe/IoMmu.c
cfernald: oops

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -102,7 +353,7 @@ UpdateMapping (
 
   // leaf level
   if (Current != 0) {
-    Index = PAGE_TABLE_INDEX (VirtualAddress, Level, mIoMmu->SmmuInfo->OutputAddressWidth, mIoMmu->SmmuInfo->TranslationStartingLevel, mIoMmu->SmmuInfo->PageTableRootConcatenated);
+    Index = PAGE_TABLE_INDEX (VirtualAddress, Level, SmmuInfo->OutputAddressWidth, SmmuInfo->TranslationStartingLevel, SmmuInfo->PageTableRootConcatenated);
 
     if (Valid && ((Current->Entries[Index] & PAGE_TABLE_ENTRY_VALID_BIT) != 0)) {
```
cfernald: move this after calculating new entry, check if the same, and return fauilure if not. Just return without setting otherwise

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -160,37 +416,29 @@ UpdatePageTable (
...
+    Status = UpdateMapping (SmmuInfo, Root, CurPhysicalAddress, CurPhysicalAddress, Flags, Valid);
     if (EFI_ERROR (Status)) {
       DEBUG ((DEBUG_ERROR, "%a: Failed to update page table mapping\n", __func__));
+      gBS->RestoreTPL (OldTpl);
       goto End;
     }
 
+    gBS->RestoreTPL (OldTpl);
+
+    ArmDataSynchronizationBarrier ();
```
cfernald: move out of loop

eeshanl: done, moved to top of SmmuV3TLBInvalidateAll() function before issueing any TLB commands.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done, added locks around STE manipulation logic in the IoMmuSetAttribute path.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done, moved to one line below

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Done added the check., updated readme, and added comments around this check.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: moved the DSB to be the first thing we do inside SmmuV3TLBInvalidateAll()

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
```diff
@@ -338,10 +377,15 @@ SmmuV3BuildStreamTableEntry (
...
   StreamEntry->S2Tg   = SMMUV3_STREAM_TABLE_ENTRY_S2TG_4KB;
   StreamEntry->S2Aa64 = 1;                                                                                // AArch64 S2 translation tables
-  StreamEntry->S2Ttb  = (UINT64)(UINTN)SmmuInfo->PageTableRoot >> SMMUV3_STREAM_TABLE_ENTRY_S2TTB_OFFSET; // Page table root address
+  if (PageTableRoot != NULL) {
+    StreamEntry->S2Ttb = (UINT64)(UINTN)PageTableRoot >> SMMUV3_STREAM_TABLE_ENTRY_S2TTB_OFFSET;  // Page table root address
+  } else {
+    StreamEntry->S2Ttb = 0;  // For abort STEs, S2Ttb is set to 0 so any access will fault since it is not a valid page-table root.
+  }
+
   if ((Idr0.S1p == 1) && (Idr0.S2p == 1)) {
```
os-d: This needs to be changed, right? We don’t care about S1p and S2p is checked for now

os-d (approved):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: we set S2PTW Protected Table Walk if both stages are supported.

"For an STE configured for translation at both stages, a stage 1 translation table walk access or CD fetch access
made to a stage 2 page with any Device type is terminated and recorded as a stage 2 Permission fault if this field is
set."

I think it is probably harmless to keep for robustness and future updates.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -102,19 +384,28 @@ UpdateMapping (
...
+
+      // Break-before-make does not apply here because we are only switching between invalid/valid, no other Entry bits are changing.
+      // If the entry is already valid, it must have the same PA and flags to be considered a match; otherwise it's an error because we don't expect multiple mappings for the same VA.
+      if ((Current->Entries[Index] & PAGE_TABLE_ENTRY_VALID_BIT) != 0) {
+        DEBUG ((DEBUG_INFO, "%a: Page already mapped with valid Entry. VirtualAddress = 0x%llx PhysicalAddress=0x%llx\n", __func__, VirtualAddress, PhysicalAddress));
+        if (Current->Entries[Index] != Entry) {
+          DEBUG ((DEBUG_ERROR, "%a: Page already mapped with different PA or flags. OldEntry = 0x%llx NewEntry = 0x%llx\n", __func__, Current->Entries[Index], Entry));
+          Status = EFI_DEVICE_ERROR;
+          goto End;
+        }
```
cfernald, line 403: might as well skip setting it if it is already identical

cfernald (approved):

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,281 @@ typedef struct IOMMU_MAP_INFO {
...
+  OUT UINT16     *OutVmid
+  )
+{
+  UINT16  MaxVmid;
+
+  MaxVmid = SmmuInfo->Vmid16Supported ? MAX_UINT16 : MAX_UINT8;
+  if (SmmuInfo->NextVmid == SMMU_VMID_RESERVED) {
+    // wrapped past the max
+    DEBUG ((DEBUG_ERROR, "%a: VMID space exhausted on SMMU 0x%llx\n", __func__, SmmuInfo->SmmuBase));
+    ASSERT (FALSE);
```
makubacki: We don't enforce this often but in cases like this, it gives additional context to directly assert on the condition.


```suggestion
    ASSERT (SmmuInfo->NextVmid != SMMU_VMID_RESERVED);
```

cfernald (approved):

MdeModulePkg/Library/IoMmuLib/IoMmuLib.c
```diff
@@ -252,5 +291,12 @@ IoMmuLibInit (
     }
   }
 
+  // Check if version is atleast EDKII_IOMMU_PROTOCOL_REVISION, if not, return unsupported
+  if ((mIoMmuProtocol != NULL) && (mIoMmuProtocol->Revision < EDKII_IOMMU_PROTOCOL_REVISION)) {
+    DEBUG ((DEBUG_ERROR, "%a: IoMmuProtocol revision 0x%llx is less than the required revision 0x%llx\n", __func__, mIoMmuProtocol->Revision, EDKII_IOMMU_PROTOCOL_REVISION));
```
kuqin12: Do you really want to assert here? Since this protocol is rev'd, would it be better to just disable the new interface?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,281 @@ typedef struct IOMMU_MAP_INFO {
...
+    return Status;
+  }
+
+  Status = SmmuV3AllocateVmid (SmmuInfo, &NewVmid);
+  if (EFI_ERROR (Status)) {
+    gBS->RestoreTPL (OldTpl);
+    return Status;
+  }
+
+  NewRoot = SmmuV3AllocatePageTableRoot ();
```
makubacki, line 177: This (`SmmuV3AllocatePageTableRoot()`) is currently calling `AllocatedAlignedPages()` at `TPL_HIGH_LEVEL` which is not allowed.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -160,37 +456,27 @@ UpdatePageTable (
   PhysicalAddressEnd = ALIGN_VALUE (PhysicalAddress + Bytes, EFI_PAGE_SIZE);
 
   while (CurPhysicalAddress < PhysicalAddressEnd) {
-    Status = UpdateMapping (Root, CurPhysicalAddress, CurPhysicalAddress, Flags, Valid);
+    OldTpl = gBS->RaiseTPL (TPL_HIGH_LEVEL);
+    Status = UpdateMapping (SmmuInfo, Root, CurPhysicalAddress, CurPhysicalAddress, Flags, Valid);
```
makubacki, line 461: This (`UpdateMapping()`) is currently calling `AllocatePages()` at `TPL_HIGH_LEVEL` which is not allowed.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -770,6 +849,158 @@ SmmuV3CmdQueueUpdateCachedConsumer (
...
+    DEBUG ((DEBUG_ERROR, "%a: IOMMU_CONFIG/SMMU_INFO structure is NULL\n", __func__));
+    return;
+  }
+
+  for (SmmuIndex = 0; SmmuIndex < mIoMmu->SmmuCount; SmmuIndex++) {
+    SmmuInfo = &mIoMmu->SmmuInfo[SmmuIndex];
+    if (SmmuInfo->Enabled) {
+      if ((Source == SmmuInfo->EvtqIrqNum) || (Source == SmmuInfo->GerrIrqNum)) {
+        SmmuV3LogErrors (SmmuInfo);
+        ASSERT (FALSE);
```
makubacki: Do you intent to unconditionally assert here? `SmmuV3LogErrors()`'s return value is based on whether errors were found. Should the assert be conditionalized on that? In general, is this worth halting an ASSERT-enabled boot over versus logging alone.

ArmPkg/Drivers/SmmuDxe/README.md
```diff
@@ -181,6 +196,311 @@ For direct mappings (no bounce buffer), only the mapping information structure i
...
+  encoded in its `S2Ttb` / `S2Vmid` fields (the STE is the single source of truth). Otherwise it allocates a fresh
+  VMID from `SMMU_INFO->NextVmid`, allocates a new root, and promotes the STE. VMID `0` is reserved as "unassigned".
+- VMID width comes from `IDR0.VMID16`: 8-bit SMMUs use values `1..0xFF`, 16-bit SMMUs use `1..0xFFFF`. SmmuDxe never
+  reuses a VMID for a different root within the same boot.
+- `SmmuV3StreamAlias` does the inverse: given the **primary's** `(Root, Vmid)`, it promotes the alias STE in
+  place to point at the same root with the same `S2VMID`. If the alias STE already encodes the same `(Root, Vmid)`
+  this is a no-op. That's how multiple StreamIDs share one mapping - they all resolve through the same VMID tag, so
+  a single TLB invalidation by `(VMID, IPA)` covers every alias.
+- `UpdatePageTable` issues `SmmuV3TLBInvalidateAddress (SmmuInfo, Vmid, IPA)` on unmap. Because the VMID is unique
+  per root, that invalidation can't accidentally evict another device's translations.
```
makubacki: Is this comment stale? I don't see `SmmuV3TLBInvalidateAddress()` called anywhere right now and it looks like `UpdatePageTable()` calls `SmmuV3TLBInvalidateAll (SmmuInfo, Vmid)`.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
```diff
@@ -1451,3 +1738,407 @@ SmmuV3ParseIort (
...
+  Node = (EFI_ACPI_6_0_IO_REMAPPING_NODE *)((UINT8 *)Iort + Iort->NodeOffset);
+  for (NodeIdx = 0; NodeIdx < Iort->NumNodes; NodeIdx++) {
+    if (Node->Type == EFI_ACPI_IORT_TYPE_NAMED_COMP) {
+      NcNode   = (EFI_ACPI_6_0_IO_REMAPPING_NAMED_COMP_NODE *)Node;
+      NodeName = (CONST CHAR8 *)((UINT8 *)NcNode + sizeof (EFI_ACPI_6_0_IO_REMAPPING_NAMED_COMP_NODE));
+      NameMax  = Node->IdReference - sizeof (EFI_ACPI_6_0_IO_REMAPPING_NAMED_COMP_NODE);
+
+      if ((NameMax > 0) && (AsciiStrnCmp (NodeName, WantedName, NameMax) == 0) &&
+          (AsciiStrLen (WantedName) < NameMax))
+      {
```
makubacki: You might want to also check that `Node->IdReference` is greater than `sizeof (EFI_ACPI_6_0_IO_REMAPPING_NAMED_COMP_NODE)` before accepting `NameMax` since the calculation is based on ACPI table data.


```suggestion
      if ((Node->IdReference > sizeof (EFI_ACPI_6_0_IO_REMAPPING_NAMED_COMP_NODE)) &&
          (NameMax > 0) && (AsciiStrnCmp (NodeName, WantedName, NameMax) == 0) &&
          (AsciiStrLen (WantedName) < NameMax))
      {
```

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -476,6 +763,169 @@ IoMmuAllocateBuffer (
...
+          AliasEntry->StreamId,
+          PrimaryStreamId,
+          TargetSmmu->SmmuBase,
+          Status
+          ));
+        ASSERT_EFI_ERROR (Status);
+        return Status;
+      }
+
+      StreamIdCount++;
```
makubacki: `StreamIdCount` is being incremented, but I don't see it read anywhere. Is it needed?

ArmPkg/Drivers/SmmuDxe/IoMmu.c
```diff
@@ -35,6 +34,281 @@ typedef struct IOMMU_MAP_INFO {
...
+{
+  EFI_STATUS                 Status;
+  SMMUV3_STREAM_TABLE_ENTRY  *AliasSte;
+  PAGE_TABLE                 *ExistingRoot;
+  UINT16                     ExistingVmid;
+  EFI_TPL                    OldTpl;
+
+  if ((SmmuInfo == NULL) || (PrimaryRoot == NULL) || (PrimaryVmid == 0)) {
+    DEBUG ((DEBUG_ERROR, "%a: Invalid parameter\n", __func__));
+    ASSERT (FALSE);
```
makubacki: Again, minor but this would produce a cleaner assert message.

```suggestion
    ASSERT_EFI_ERROR (EFI_INVALID_PARAMETER);
```

I won't leave any more comments about assert messages.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: Yes, if this interrupt triggers, then we already know an SMMU fault has occured in the event queue, or a global error has been set. So, we want to halt regardless.

makubacki: Can you please leave a brief comment to explain that? First thing someone is going to do when the assert is hit is come to this line of code.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done, fixed up all the asserts throughout the SmmuDxe/ to produce cleaner assert messages.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: good catch, removed.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: done

ArmPkg/Drivers/SmmuDxe/README.md
eeshanl: Fixed the readme to document SmmuV3TLBInvalidateAll instead.

ArmPkg/Drivers/SmmuDxe/SmmuV3Util.c
eeshanl: I have removed the ASSERT (FALSE) from the IsrHandler as SmmuV3LogErrors already asserted if an error was found. Added comment on SmmuV3LogErrors to call out that it will assert on error.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Fixed all TPL locks to be more granular and made sure we do not allocate within a locked region.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: Fixed all TPL locks to be more granular and made sure we do not allocate within a locked region.

ArmPkg/Drivers/SmmuDxe/IoMmu.c
eeshanl: done

MdeModulePkg/Library/IoMmuLib/IoMmuLib.c
eeshanl: Changed so that if any of the following conditions are not met:

`
    mSetAttributeByIdSupported = ((mIoMmuProtocol != NULL) &&
                                  (mIoMmuProtocol->Revision >= EDKII_IOMMU_PROTOCOL_REVISION) &&
                                  (mIoMmuProtocol->SetAttributeById != NULL)
                                  );
`

IoMmuSetAttributeById will return EFI_UNSUPPORTED.

makubacki (approved):

kuqin12 (approved):

ArmPkg/Drivers/SmmuDxe/SmmuDxe.c
eeshanl: removed.

cfernald (approved):

PR #1817, opened 2026-06-08
Fixing getting policy from inside the notification

kuqin12:
## Description

Policy service supports registering a callback based on the event type. However, the callback will be dispatched while the atomic lock was acquired, causing the handler not able to getting the updated data.

This change converts the lock acquisition function to a soft acquire when performing get policy function.

A new test case is also added to cover such usage.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [x] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This was tested on QEMU AArch64 platform and was able to get policy after the change.

## Integration Instructions

N/A

### DIFF
--- a/PolicyServicePkg/PolicyService/DxeMm/UnitTest/DxeMmPolicyUnitTest.c
+++ b/PolicyServicePkg/PolicyService/DxeMm/UnitTest/DxeMmPolicyUnitTest.c
@@ -41,6 +41,13 @@ NOTIFICATION_TRACKER  Notifications[100]      = { 0 };
 UINT32                NotificationsCount      =  0;
 UINT32                NotifyCountUpdateRemove =  0;
 
+UINT8    GetPolicyExpectedData[32]  = { 0 };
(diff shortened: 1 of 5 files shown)


cfernald (approved):

PolicyServicePkg/PolicyService/DxeMm/UnitTest/DxeMmPolicyUnitTest.c
```diff
@@ -150,6 +180,44 @@ GenericNotify (
   }
```
kuqin12, line 180: @cfernald hmm, this is failing the test, which makes sense. It is trying to update the notify routine and set other policy. Is this supported?

kuqin12:
Closing with this github issue: https://github.com/microsoft/mu_basecore/issues/1818. Will patch the caller to be "query while using" instead.

PR #1827, opened 2026-06-19
[Rebase & FF] Fix Partition ID being used for FFA_RUN calls

kuqin12:
## Description

The current implementation of FFA run invocation always uses the original partition ID as the input for FFA_RUN. However, this is problematic when the partitions are chain loading the direct request 2 functions and the partitions are interruptible. In that case, the caller should FFA_RUN to the partition specified in w1.

Then the w1 information was eaten in the case of direct request calls, because we only exposed the payload to the callers.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on ARM virt and physical platform.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationPei/MmCommunicationPei.c
+++ b/ArmPkg/Drivers/MmCommunicationPei/MmCommunicationPei.c
@@ -235,7 +235,8 @@ SendFfaMmCommunicate (
 
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the StMM SP since it is UP.
-    Status = ArmFfaLibRun (mStMmPartId, 0x00, NULL);
+    // MU_CHANGE: Use the source partition ID of the FFA header for the ffa_run call.
+    Status = ArmFfaLibRun (GET_SOURCE_PARTITION_ID (CommunicateArgs.Header.x1), 0x00, &CommunicateArgs);
   }
 
   return Status;
--- a/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
+++ b/MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
@@ -741,12 +741,16 @@ ArmFfaLibRun (
   ArmCallFfa (&FfaArgs);
 
   Status = FfaArgsToEfiStatus (&FfaArgs);
-  if (EFI_ERROR (Status)) {
-    return Status;
-  }
+  // MU_CHANGE: Copy the FFA header to the direct message arguments
+  // if (EFI_ERROR (Status)) {
+  //   return Status;
+  // }
 
   if (DirectMsgArg != NULL) {
     ZeroMem (DirectMsgArg, sizeof (DIRECT_MSG_ARGS));
+
+    // MU_CHANGE: Copy the FFA header to the direct message arguments
+    CopyMem (&(DirectMsgArg->Header), &FfaArgs, sizeof (DirectMsgArg->Header));
 
     if (FfaArgs.Arg0 == ARM_FID_FFA_MSG_SEND_DIRECT_RESP) {
       DirectMsgArg->Arg0 = FfaArgs.Arg3;
@@ -782,6 +786,9 @@ ArmFfaLibRun (
   @param [in]      Flags            Message flags
   @param [in, out] ImpDefArgs       Implemented defined arguments and
                                     Implemented defined return values
+                                    // MU_CHANGE: Copy the FFA header to the direct message arguments
+                                    The header registers (x0-x2) will be initialized
+                                    with the values from DestPartId and Flags.
 
   @retval EFI_SUCCESS               Success
   @retval Others                    Error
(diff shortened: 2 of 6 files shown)


MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -769,6 +766,8 @@ ArmFfaLibRun (
       DirectMsgArg->Arg11 = FfaArgs.Arg15;
       DirectMsgArg->Arg12 = FfaArgs.Arg16;
       DirectMsgArg->Arg13 = FfaArgs.Arg17;
+    } else {
```
cfernald: It seems this commit could be squashed with the commit creating the definitions.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
```diff
@@ -769,6 +766,8 @@ ArmFfaLibRun (
       DirectMsgArg->Arg11 = FfaArgs.Arg15;
       DirectMsgArg->Arg12 = FfaArgs.Arg16;
       DirectMsgArg->Arg13 = FfaArgs.Arg17;
+    } else {
+      CopyMem (DirectMsgArg, &FfaArgs, sizeof (DIRECT_MSG_ARGS));
```
cfernald: I wonder if you should just always copy the header before checking the response, and let the args be copied for the specific reponses.

ArmPkg/Drivers/MmCommunicationPei/MmCommunicationPei.c
```diff
@@ -235,7 +235,7 @@ SendFfaMmCommunicate (
 
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the StMM SP since it is UP.
-    Status = ArmFfaLibRun (mStMmPartId, 0x00, NULL);
+    Status = ArmFfaLibRun (GET_SOURCE_PARTITION_ID (CommunicateArgs.Header.x1), 0x00, &CommunicateArgs);
```
cfernald, line 239: nit: seems this could be squashed with the DXE phase commit

SecurityPkg/Tcg/Tcg2Config/Tcg2ConfigFfaPeim.c
```diff
@@ -69,7 +69,7 @@ Tpm2FfaCheckInterfaceVersion (
   Status = ArmFfaLibMsgSendDirectReq2 (TpmPartId, &gTpm2ServiceFfaGuid, &TpmArgs);
   while (Status == EFI_INTERRUPT_PENDING) {
     // We are assuming vCPU0 of the TPM SP since it is UP.
-    Status = ArmFfaLibRun (TpmPartId, 0x00, &TpmArgs);
+    Status = ArmFfaLibRun (GET_SOURCE_PARTITION_ID (TpmArgs.Header.x1), 0x00, &TpmArgs);
```
cfernald, line 73: ditto with the squashing. Implementors choice though

cfernald (approved):

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
kuqin12: Will update.

SecurityPkg/Tcg/Tcg2Config/Tcg2ConfigFfaPeim.c
kuqin12: Agreed with other commit suggestions. I would have done them in 2 commits. But that might be controversial because it will eventually need each targeted reviewer to sign off, if we view it from upstreaming perspective.

cfernald: Yeah, i just mean effectively squash to 3 commits, one for each package. But I dont really care too much, so treat this as merely a suggestion.

cfernald (approved):

apop5 (approved):

SecurityPkg/Library/Tpm2DeviceLibFfa/Tpm2ServiceFfaRaw.c
```diff
@@ -242,8 +243,10 @@ Tpm2ServiceStart (
 
   Status = ArmFfaLibMsgSendDirectReq2 (FfaTpm2PartitionId, &gTpm2ServiceFfaGuid, &FfaDirectReq2Args);
   while (Status == EFI_INTERRUPT_PENDING) {
+    DEBUG ((DEBUG_INFO, "Tpm2ServiceStart: Interrupt pending, waiting...\n"));
```
kuqin12: Remove debug code.

kuqin12: Removed.

MdeModulePkg/Library/ArmFfaLib/ArmFfaCommon.c
kuqin12: Updated.

apop5 (approved):

PR #1832, opened 2026-06-22
[CHERRY-PICK] FatPkg: Clean any volume caches during exit boot services

cfernald:
## Description

The current implementation assumes the the caller will perform the necessary cleanup before exiting boot services. This has been observed to drop some cached file writes that occur even before the application calling exit boot services is launched.

This commit adds a per-volume pre-ExitBootServices event to flush any dirty caches and perform other cleanup the volume to ensure all write data is persisted and consistent. After the flush, caching will be disabled for the volume in the future to ensure that all subsequent access persists.

- [x] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Tested on QEMU with one-off shell test and boot to OS

## Integration Instructions

N/A

### DIFF
--- a/FatPkg/EnhancedFatDxe/Misc.c
+++ b/FatPkg/EnhancedFatDxe/Misc.c
@@ -331,6 +331,14 @@ FatDiskIo (
   //
   Status = EFI_VOLUME_CORRUPTED;
   if (Offset + BufferSize <= Volume->VolumeSize) {
+    if (Volume->CachingDisabled) {
+      //
+      // Caching has been turned off for this volume. Convert
+      // IO mode to raw disk access equivalent.
+      //
+      IoMode = (IO_MODE)RAW_ACCESS (IoMode);
+    }
+
     if (CACHE_ENABLED (IoMode)) {
       //
       // Access cache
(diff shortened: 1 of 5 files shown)


os-d (approved):
nit: We've been doing `git cherry-pick -x` to get the cherry-picked commit hash in the commit message for easy reference.

FatPkg/EnhancedFatDxe/Misc.c
```diff
@@ -449,6 +457,68 @@ FatFreeDirEnt (
...
+  @param Context - The context of the event, which is the FAT_VOLUME for which to flush caches.
+
+**/
+VOID
+EFIAPI
+
+FatOnBeforeExitBootServices (
+  IN EFI_EVENT  Event,
+  IN VOID       *Context
+  )
```
makubacki, line 475: It would be ideal if this empty line was not here but if it's in the cherry-pick source, feel free to ignore.

```suggestion
VOID
EFIAPI
FatOnBeforeExitBootServices (
  IN EFI_EVENT  Event,
  IN VOID       *Context
  )
```

makubacki (approved):

FatPkg/EnhancedFatDxe/Misc.c
cfernald: oh weird. It is in the source, I guess I accidentally added that at some point. I will make a minor PR upstream to fix.

apop5 (approved):

apop5 (approved):

PR #1846, opened 2026-07-10
[Rebase & FF] Adding MM communication buffer validation

kuqin12:
## Description

This change adds a few validation before using the derived buffer size from message length.

However, note that the buffer size is not used directly due to a check after that. Thus the MM communication agent or the core in MM will not consume the value directly. But this would cause the MM handler to potentially receive a huge message in length.

- [ ] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

This is tested on QEMU ARM Virt platform and booted to Windows desktop.

## Integration Instructions

N/A

### DIFF
--- a/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
+++ b/ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
@@ -19,6 +19,7 @@
 #include <Library/UefiBootServicesTableLib.h>
 #include <Library/UefiLib.h>
 #include <Library/UefiRuntimeServicesTableLib.h>
+#include <Library/SafeIntLib.h> // MU_CHANGE: Add SafeIntLib for safe integer operations
 
 #include <Protocol/MmCommunication2.h>
 #include <Protocol/MmCommunication3.h>
@@ -213,6 +214,7 @@ MmCommunicationCommon (
   EFI_MM_COMMUNICATE_HEADER     *CommunicateHeader;
   EFI_MM_COMMUNICATE_HEADER_V3  *CommunicateHeaderV3;
   UINTN                         BufferSize;
+  UINTN                         InputBufferSize;  // MU_CHANGE
   UINTN                         *MessageSize;
   UINTN                         HeaderSize;
   EFI_STATUS                    Status;
@@ -243,14 +245,40 @@ MmCommunicationCommon (
     BufferSize          = CommunicateHeaderV3->BufferSize;
     MessageSize         = &CommunicateHeaderV3->MessageSize;
     HeaderSize          = sizeof (EFI_MM_COMMUNICATE_HEADER_V3);
+
+    // MU_CHANGE Starts: Add a check to ensure that the buffer size in the header is sane
+    if (BufferSize < HeaderSize) {
+      return EFI_INVALID_PARAMETER;
+    }
+
+    if (BufferSize - HeaderSize < *MessageSize) {
+      return EFI_INVALID_PARAMETER;
+    }
+
+    // MU_CHANGE Ends
   } else {
-    BufferSize = CommunicateHeader->MessageLength +
-                 sizeof (CommunicateHeader->HeaderGuid) +
-                 sizeof (CommunicateHeader->MessageLength);
+    // MU_CHANGE: Use SafeIntLib for safe arithmetic operations
+    // BufferSize = CommunicateHeader->MessageLength +
+    //              sizeof (CommunicateHeader->HeaderGuid) +
+    //              sizeof (CommunicateHeader->MessageLength);
+    Status = SafeUintnAdd (
+               CommunicateHeader->MessageLength,
+               OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data),
+               &BufferSize
+               );
+    if (EFI_ERROR (Status)) {
+      return EFI_INVALID_PARAMETER;
+    }
+
     MessageSize = &CommunicateHeader->MessageLength;
-    HeaderSize  = sizeof (CommunicateHeader->HeaderGuid) +
-                  sizeof (CommunicateHeader->MessageLength);
+    // MU_CHANGE: Avoid direct calculation of HeaderSize
+    // HeaderSize  = sizeof (CommunicateHeader->HeaderGuid) +
+    //               sizeof (CommunicateHeader->MessageLength);
+    HeaderSize = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data);
   }
+
+  // MU_CHANGE: Record the input buffer size for later comparison with the returned buffer size
+  InputBufferSize = BufferSize;
 
   // If CommSize is not omitted, perform size inspection before proceeding.
   if (CommSize != NULL) {
@@ -269,6 +297,9 @@ MmCommunicationCommon (
     if (*CommSize < BufferSize) {
       Status = EFI_INVALID_PARAMETER;
     }
+
+    // MU_CHANGE: Record the updated input buffer size for later comparison with the returned buffer size
+    InputBufferSize = *CommSize;
   }
 
   //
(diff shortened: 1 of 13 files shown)


kuqin12:
The platform validation is failed due to override hash failure.

apop5 (approved):

StandaloneMmPkg/Core/StandaloneMmCore.c
```diff
@@ -560,10 +561,34 @@ MmEntryPoint (
...
+        }
       }
 
-      if (BufferSize <= EFI_PAGES_TO_SIZE (mMmCommunicationBuffer->NumberOfPages)) {
+      // MU_CHANGE: Ensure the buffer size does not exceed the allocated MM Communication Buffer size
+      Status = SafeUintnAdd (
+                 EFI_PAGE_SIZE,
+                 mMmCommunicationBuffer->NumberOfPages,
+                 &MaxBufferSize
+                 );
```
makubacki: I believe this was intended.

```suggestion
      Status = SafeUintnMult (
                 EFI_PAGE_SIZE,
                 mMmCommunicationBuffer->NumberOfPages,
                 &MaxBufferSize
                 );
```

StandaloneMmPkg/Core/StandaloneMmCore.c
```diff
@@ -585,8 +610,20 @@ MmEntryPoint (
                         &BufferSize
                         );
 
-        BufferSize = BufferSize + CommHeaderSize;
-        if (BufferSize <= EFI_PAGES_TO_SIZE (mMmCommunicationBuffer->NumberOfPages)) {
+        // MU_CHANGE: Use SafeIntLib to safely add the communication header size back to the buffer size if needed
+        // BufferSize = BufferSize + CommHeaderSize;
+        Status = SafeUintnAdd (
```
makubacki: This overwrites `Status` from the `MmiManage()` call, so the status from the addition is returned instead. You'll probably need to use a second status variable for the addition.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
```diff
@@ -320,9 +333,18 @@ MmCommunicationCommon (
...
       BufferSize          = CommunicateHeaderV3->BufferSize;
     } else {
-      BufferSize = CommunicateHeader->MessageLength +
-                   sizeof (CommunicateHeader->HeaderGuid) +
-                   sizeof (CommunicateHeader->MessageLength);
+      // MU_CHANGE: Use SafeIntLib for safe arithmetic operations
+      // BufferSize = CommunicateHeader->MessageLength +
+      //              sizeof (CommunicateHeader->HeaderGuid) +
+      //              sizeof (CommunicateHeader->MessageLength);
+      Status = SafeUintnAdd (
```
makubacki, line 359: Given `CommunicateHeader->MessageLength` is `UINT64`:

```
typedef struct {
  ///
  /// Allows for disambiguation of the message format.
  ///
  EFI_GUID    HeaderGuid;
  ///
  /// Describes the size of Data (in bytes) and does not include the size of the header.
  ///
  // MU_CHANGE: BZ3398 Make MessageLength the same size in EFI_MM_COMMUNICATE_HEADER for both IA32 and X64.
  UINT64      MessageLength;
  ///
  /// Designates an array of bytes that is MessageLength in size.
  ///
  UINT8       Data[1];
} EFI_MM_COMMUNICATE_HEADER;
```

And `BufferSize` is `UINT64`:

```c
typedef struct {
  ///
  /// Indicator GUID for MM core that the communication buffer is compliant with this v3 header.
  /// Must be gEfiMmCommunicateHeaderV3Guid.
  ///
  EFI_GUID    HeaderGuid;
  ///
  /// Describes the size of the entire buffer (in bytes) available for communication, including this communication header.
  ///
  UINT64      BufferSize;
  ///
  /// Reserved for future use.
  ///
  UINT64      Reserved;
  ///
  /// Allows for disambiguation of the message format.
  ///
  EFI_GUID    MessageGuid;
  ///
  /// Describes the size of MessageData (in bytes) and does not include the size of the header.
  ///
  UINT64      MessageSize;
  ///
  /// Designates an array of bytes that is MessageSize in size.
  ///
  UINT8       MessageData[];
} EFI_MM_COMMUNICATE_HEADER_V3;
```

Was there a reason not to use `SafeUint64Add()` (and in some similar places)?

StandaloneMmPkg/Core/StandaloneMmCore.c
```diff
@@ -574,7 +599,7 @@ MmEntryPoint (
           );
         ZeroMem (
           (UINT8 *)mInternalCommBufferCopy + BufferSize,
-          EFI_PAGES_TO_SIZE (mMmCommunicationBuffer->NumberOfPages) - BufferSize
+          MaxBufferSize - BufferSize // MU_CHANGE: Use MaxBufferSize instead of EFI_PAGES_TO_SIZE
           );
 
         BufferSize -= CommHeaderSize;
```
makubacki: For V3, `BufferSize` appears to be `CommunicateHeader->BufferSize` (otherwise it's `OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data) + LegacyCommunicateHeader->MessageLength`).

In the V3 case, is it guaranteed that `BufferSize` is >= `sizeof (EFI_MM_COMMUNICATE_HEADER_V3` (the value of `CommHeaderSize` in that case)?

StandaloneMmPkg/Core/StandaloneMmCore.c
```diff
@@ -560,10 +561,34 @@ MmEntryPoint (
...
-        BufferSize              = OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data) + LegacyCommunicateHeader->MessageLength;
+        // MU_CHANGE: Use SafeIntLib for safe arithmetic operations
+        Status = SafeUintnAdd (
+                   OFFSET_OF (EFI_MM_COMMUNICATE_HEADER, Data),
+                   LegacyCommunicateHeader->MessageLength,
+                   &BufferSize
+                   );
+        if (EFI_ERROR (Status)) {
+          DEBUG ((DEBUG_ERROR, "Failed to calculate buffer size: %r\n", Status));
+          ASSERT (FALSE);
```
makubacki: nit: This would be a little more helpful.

```suggestion
          ASSERT_EFI_ERROR (Status);
```

StandaloneMmPkg/Core/StandaloneMmCore.c
kuqin12: Agreed. This is fixed.

StandaloneMmPkg/Core/StandaloneMmCore.c
kuqin12: Added a new variable to record the status

StandaloneMmPkg/Core/StandaloneMmCore.c
kuqin12: A few new checks are added to prevent that from happening. Some more checks are added before copying the content back into the caller supplied buffer.

StandaloneMmPkg/Core/StandaloneMmCore.c
kuqin12: Thanks. This is updated.

ArmPkg/Drivers/MmCommunicationDxe/MmCommunication.c
kuqin12: I think this is rooted from the original PI spec defining MessageLength to be UINTN. And then a few other interfaces are cascaded with UINTN as the in/out argument type.

We can update the functions used here, and also update the local variables, but we will likely sprinkle a few (other) pointer castings here and there.

makubacki: The larger issue that comes to mind is truncation on a 32-bit machine where `UINTN` is 32-bit (which I don't think is a likely practical target except in the PEI modules). In any case, I think it should be as consistent as possible for the current types while making these types of validation correctness changes.

kuqin12: Agree with the concern. Part of the issue arise from the common communication routine between v1 and v3. If we separate them, some of the issues will be resolved, where v1 keeps using uintn while v3 uses only u64.

But I think for the case of these DXE phased modules, we have been using 64 bit as far as I am aware, so the UINTN and UINT64 are equivalent. The ARM modules are all UINT64 at this point.

So the only issue is in StandaloneMmIplPei, which does use 2 separate routines for v1 and v3 communicate. I will update the types there to match the native usage.

kuqin12: The STMM IPL is updated.

makubacki (approved):

PR #1855, opened 2026-07-21
MdeModulePkg: Print unknown error code instead of asserting

cfernald:
## Description

DumpUicCmdExecResult is called to dump the result of UIC commands after a failure. Depending on the nature of the failure, not all information may have been properly initialized. This is already handled in callers with existing retry logic, but the assert in the dump command can cause a crash in debug builds for due to hardware race conditions on first attempt.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Testing on internal hardware

## Integration Instructions

N/A


os-d (approved):
Also take to edk2 in parallel?

makubacki (approved):

makubacki:
I don't think a function to dump information in debug prints should have asserted anyway. Code somewhere else should assert/handle the error but not here.

cfernald:
> Also take to edk2 in parallel?

That's the plan

os-d (approved):

PR #1866, opened 2026-07-27
ArmPlatformPkg, MdeModulePkg, SecurityPkg: Update Tpm2StartupLib

Raymond-MS:
## Description

Updated the Tpm2StartupLib to consolidate the functionality from Tcg2Pei. This functionality was updated such that it can be used in both SEC and PEI. Updated PeilessSec to include the updated Tpm2StartupLib functionality. Moved the SyncPcrs functionality to Tcg2Dxe; this is only able to be done when the PCD lib is available due to dynamic PCDs. This gives us a common location for syncing to occur on both PEI and PEI-less platforms. Created a fixed version of the DeviceStateLib. This was required due to the use of another dynamic PCD which is not available in SEC. Moved the PCD in MdeModulePkg.dec such that it can be set as FixedAtBuild.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Verified boot to UEFI shell on both Q35 (PEI path) and VIRT (SEC path).

## Integration Instructions

Add Tpm2StartupLib.inf to your platform's .dsc.


SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -1,78 +1,940 @@
...
+          DigestCount++;
+          *RemainingTpm2HashMask &= ~HashAlgoMask;
+        }
+
+        PreHashInfo = (HASH_INFO *)((UINT8 *)(PreHashInfo + 1) + PreHashInfo->HashSize);
+      }
+
+      WriteUnaligned32 (&DigestList->count, DigestCount);
+      return EFI_SUCCESS;
+    }
```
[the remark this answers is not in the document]
Raymond-MS: Good catch, fixed.

SecurityPkg/Tcg/Tcg2Dxe/Tcg2Dxe.c
```diff
@@ -3224,6 +3226,81 @@ InstallTcg2 (
...
+  if ((Tpm2PcrMask != TpmActivePcrBanks) && FixedPcdGetBool (PcdForceReallocatePcrBanks)) {
+    DEBUG ((DEBUG_INFO, "%a - Reallocating PCR banks from 0x%X to 0x%X.\n", __func__, TpmActivePcrBanks, Tpm2PcrMask));
+    Status = Tpm2PcrAllocateBanks (NULL, (UINT32)TpmHashAlgorithmBitmap, Tpm2PcrMask);
+    if (EFI_ERROR (Status)) {
+      DEBUG ((DEBUG_ERROR, "%a - Failed to reallocate PCRs!\n", __func__));
+      ASSERT_EFI_ERROR (Status);
+    }
+
+    ResetCold ();
+  }
```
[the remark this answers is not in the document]
Raymond-MS: Good catch, fixed.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
```diff
@@ -1080,8 +1103,202 @@ FirmwareVolumeInfoPpiNotifyCallback (
...
+    Status = PeiServicesLocatePpi (
+               &gEfiPeiFirmwareVolumeInfoMeasurementExcludedPpiGuid,
+               Instance,
+               NULL,
+               (VOID **)&MeasurementExcludedFvPpi
+               );
+    if (!EFI_ERROR (Status)) {
+      if (MeasurementExcludedFvPpi->Count <= 0) {
+        DEBUG ((DEBUG_ERROR, "ExcludedFvPpi has invalid count %d", MeasurementExcludedFvPpi->Count));
+        ASSERT (MeasurementExcludedFvPpi->Count > 0);
```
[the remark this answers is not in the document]
Raymond-MS: Sure, this makes sense. Fixed.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -1,78 +1,981 @@
...
+
+  Status = Tpm2GetCapabilitySupportedAndActivePcrs (&HashAlgorithmBitmap, &ActivePcrBanks);
+  if (EFI_ERROR (Status)) {
+    return EFI_DEVICE_ERROR;
+  }
+
+  SupportedEventLogs = EFI_TCG2_EVENT_LOG_FORMAT_TCG_1_2 | EFI_TCG2_EVENT_LOG_FORMAT_TCG_2;
+  RetStatus          = EFI_SUCCESS;
+
+  for (Index = 0; Index < sizeof (mTcg2EventInfo) / sizeof (mTcg2EventInfo[0]); Index++) {
```
cfernald: nit: use `ARRAY_SIZE`

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -1,78 +1,981 @@
...
+    if ((SupportedEventLogs & mTcg2EventInfo[Index].LogFormat) == 0) {
+      continue;
```
cfernald: Should this be an assert, or perhaps deleted? Considering both SupportedEventLogs and mTcg2EventInfo are hard-coded this should never happen unless someone changes the code.

ArmPlatformPkg/PeilessSec/PeilessSec.c
```diff
@@ -206,29 +206,40 @@ SecMain (
...
+              FvHob.FirmwareVolume->Length
+              );
+      FvHob.Raw = GET_NEXT_HOB (FvHob);
+      FvHob.Raw = GetNextHob (EFI_HOB_TYPE_FV, FvHob.Raw);
+    }
+
+    Tpm2StartupPublishMeasuredFvHob ();
+  } else if (Status != EFI_UNSUPPORTED) {
+    DEBUG ((DEBUG_ERROR, "%a: Tpm2StartupInitializeTpm failed: %r\n", __func__, Status));
+    ASSERT_EFI_ERROR (FALSE);
```
cfernald:
```suggestion
    ASSERT_EFI_ERROR (Status);
```

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -1,78 +1,981 @@
...
+  if (EFI_ERROR (Status)) {
+    return EFI_DEVICE_ERROR;
+  }
+
+  if (IsFvMeasurementExcluded (FvBase, FvLength)) {
+    DEBUG ((DEBUG_INFO, "FV excluded from measurement: base=0x%lx len=0x%lx\n", FvBase, FvLength));
+    return EFI_SUCCESS;
+  }
+
+  if (IsFvAlreadyRecorded (FvBase, FvLength)) {
```
cfernald, line 690: Wouldn't you want to check this with the values returned from `ResolveMigratedFvBases`?

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
```diff
@@ -1,78 +1,981 @@
...
+  Count   = 0;
+  GuidHob = GetFirstGuidHob (&mTpm2StartupMeasuredFvHobGuid);
+  while (GuidHob != NULL) {
+    Count++;
+    GuidHob = GetNextGuidHob (&mTpm2StartupMeasuredFvHobGuid, GET_NEXT_HOB (GuidHob));
+  }
+
+  MeasuredHobData = BuildGuidHob (
+                      &gMeasuredFvHobGuid,
+                      sizeof (UINTN) + sizeof (EFI_PLATFORM_FIRMWARE_BLOB) * Count
```
cfernald: Why UINTN, isn't `MeasuredHobData->Num` and UINT32?

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
```diff
@@ -1324,9 +1578,29 @@ PeimEntryMA (
     //
     Status = PeiServicesInstallPpi (&mTpmInitializedPpiList);
     ASSERT_EFI_ERROR (Status);
+ #endif // MU_CHANGE - [END]
+
+    // MU_CHANGE - [BEGIN]
+    BuildExcludedFvHobFromPpi ();
```
cfernald: This used to occur in EndofPeiSignalNotifyCallBack, right? Is it guaranteed the PPIs will be published at the init here?

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
```diff
@@ -1097,16 +1322,19 @@ PeimEntryMP (
   IN      EFI_PEI_SERVICES  **PeiServices
   )
 {
-  EFI_STATUS  Status;
+  EFI_STATUS         Status;
+  EFI_PEI_FV_HANDLE  VolumeHandle; // MU_CHANGE
+  EFI_FV_INFO        VolumeInfo;   // MU_CHANGE
 
-  DEVICE_STATE  CurrentDeviceState; // MU_CHANGE - Measure Firmware Debugger Enabled
+  // DEVICE_STATE  CurrentDeviceState; // MU_CHANGE
```
cfernald: It's a MU change to begin with. I would just delete this line if you no longer need it.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
```diff
@@ -1097,16 +1322,19 @@ PeimEntryMP (
...
 
   //
   // install Tcg Services
   //
   Status = PeiServicesInstallPpi (&mTcgPpiList);
   ASSERT_EFI_ERROR (Status);
 
+ #if 0 // MU_CHANGE - [BEGIN]
   // MU_CHANGE_103691
   // MU_CHANGE [BEGIN] - Add support for measurements extended before Tcg2 stack is available.
```
cfernald: Delete these blocks of MU changes if you are #if 0'ing the section. No point layering our tracked changes.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
cfernald: ditto on deleting MU changes if they are no-op'ed

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
```diff
@@ -1324,9 +1578,29 @@ PeimEntryMA (
...
+    }
+
+    Status = PeiServicesInstallPpi (&mTpmInitializedPpiList);
+    ASSERT_EFI_ERROR (Status);
+    // MU_CHANGE - [END]
   }
 
   if (mImageInMemory) {
+ #if 0 // MU_CHANGE - [BEGIN]
     // MU_CHANGE_23086
```
cfernald: here as well.

cfernald (commented):
Reading through the PR, I do think there is the layering is a bit off for the startup lib. It declares itself as BASE but does take dependence on the phase as we discussed offline with the use of HOB creation. Now the obvious problem is that there is no suitable library (except perhaps TpmMeasurementLib, but that isn't for internal use really) that provides the abstraction for the log writing. Realistically, this means either making StartupLib split into phases or creating yet-another library (e.g. Tcg2EventLogLib) that would have a DXE and a PEI/SEC version.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
Raymond-MS: Good catch, yeah. I am not sure why I added this. It doesn't feel like it adds anything. Removed.

ArmPlatformPkg/PeilessSec/PeilessSec.c
Raymond-MS: Good catch! Fixed!

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
Raymond-MS: I took some time to look into this and I don't believe so. The FvBase and FvLength are passed into the measure function and are used to create the mTpm2StartupMeasuredFvHobGuid HOB. ResolveMigratedFvBases return values (FvOrgBase and FvDataBase) wouldn't be found in IsFvAlreadyRecorded.

SecurityPkg/Library/Tpm2StartupLib/Tpm2StartupLib.c
Raymond-MS: This is actually a very good catch. This was how it was done in PEI before but even UINT32 is incorrect. The better option is to use OFFSET_OF in order to correctly calculate the size. Fixed.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
Raymond-MS: Good catch. Unfortunately, no. I need to fix this. It is possible that not all PPIs will be available at the entry point of Tcg2Pei and thus could be missed.

Raymond-MS: Okay, I worked through this problem. Thank you for catching this. I ended up making the ExcludedFvHob and PrehashedFvHob functions callbacks which get triggered when the PPI is installed for each meaning that anytime an ExcludedFvHob or PrehashedFvHob is needed, the callback will trigger and the HOB will get generated. These were added to the NotifyList and I made sure that this gets installed before we make the call to measure the main FV this way we can gather any currently active ExcludedFvHob or PrehashedFvHob. Afterwards, if any new ExcludedFvHob or PrehashedFvHob is generated, it will get checked when a new FV needs to be hashed via the PPI. This should resolve this issue. Please let me know if this sounds good to you. Thanks!

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
Raymond-MS: Removed.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
Raymond-MS: Removed.

SecurityPkg/Tcg/Tcg2Pei/Tcg2Pei.c
Raymond-MS: Removed. Will do for all instances.

Raymond-MS:
Going to close this PR. After discussions with @spbrogan and @cfernald we decided to go a different route with the startup library. The current implementation has issues where if an intel system wanted to include the library in SEC it would not work. The idea was to have something that worked for PEI-less platforms as well as platforms with PEI. This focused the startup library into a solution that wouldn't always work in both SEC and PEI (not necessarily anyway). The implementation is also too complex and contains measurements of the FVs which we decided should stay within their respective phases. The current idea is to simplify the implementation down to what we think is currently necessary and expand on it in the future if necessary.

PR #1877, opened 2026-08-06
CryptoPkg: Update OneCrypto binary ext_dep to v1.0.2

Flickdm:
## Description

This pull request updates the `OneCrypto_ext_dep.json` file to use the latest version of the OneCrypto binary. The update ensures that the dependency is current and that the integrity check matches the new release.

Dependency update:

* Updated the `source` URL, `version`, and `sha256` checksum in `CryptoPkg/Binaries/OneCrypto_ext_dep.json` to reference OneCrypto version 1.0.2 instead of 1.0.1.

Since we're holding on to the `dual copy` approach - this PR is _not_ a breaking change. If we drop this - then this may be a breaking change. 

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Single copy and dual copy via QemuArmVirtPkg

## Integration Instructions

This introduces two styles of integration. 

Single copy is the preferred mechanism and works when the platform supports DXE and S*MM.

Single Copy FDF
```inf
# DXE FV: loader only. Do not include OneCryptoBinDxe.
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoLoaders/OneCryptoLoaderDxeFromMm.inf

# Standalone MM FV
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoLoaders/OneCryptoLoaderStandaloneMm.inf
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoLoaders/OneCryptoImageProviderStandaloneMm.inf

# The sole binary copy, either directly in the Standalone MM FV:
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoBin/OneCryptoBinStandaloneMm.inf

# Or in a dedicated compressed FV embedded in Standalone MM, as QemuArmVirt does.
```

If your platform cannot communicate over a MM-Communicate channel or only supports DXE. Then the dual copy mechanism can be used. Where on the platform without S*MM you can just include only the DXE changes.

Dual Copy FDF
```inf
# DXE FV
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoLoaders/OneCryptoLoaderDxe.inf
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoBin/OneCryptoBinDxe.inf

# Standalone MM FV
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoLoaders/OneCryptoLoaderStandaloneMm.inf
INF $(ONE_CRYPTO_PATH)/$(TARGET)/AARCH64/OneCryptoBin/OneCryptoBinStandaloneMm.inf
```

### DIFF
--- a/CryptoPkg/Binaries/OneCrypto_ext_dep.json
+++ b/CryptoPkg/Binaries/OneCrypto_ext_dep.json
@@ -3,11 +3,13 @@
   "type": "web",
   "id": "onecrypto-bin",
   "name": "onecrypto-bin",
-  "source": "https://github.com/microsoft/mu_crypto_release/releases/download/v1.0.1-OneCrypto/OneCrypto-Accelerated.zip",
-  "version": "1.0.1",
-  "sha256": "4f7ff3f6483c35f9168c67b28a462353f138115a3cd024551d80525550cd54b1",
+  "source": "https://github.com/microsoft/mu_crypto_release/releases/download/v1.0.2-OneCrypto/OneCrypto-Accelerated.zip",
+  "version": "1.0.2",
+  "sha256": "76e71a85e313417bca3644e69cb12577c51150da7be6535dca38ce698199ab52",
   "compression_type": "zip",
   "internal_path": "/",
-  "flags": ["set_build_var"],
+  "flags": [
+    "set_build_var"
+  ],
   "var_name": "BLD_*_ONE_CRYPTO_PATH"
 }


Javagedes (approved):

Flickdm:
Waiting on success from here: https://github.com/microsoft/mu_tiano_platforms/pull/1460

Flickdm:
In v1.0.3 I will revisit dropping dual copy (this will be a breaking change) - since I am losing my ability to test the binaries with the move to single copy. What will be challenging is that some platforms can't easily move off. Some platforms depend on a dxe only copy currently satisfied by dual copy.

makubacki (approved):

makubacki:
> In v1.0.3 I will revisit dropping dual copy (this will be a breaking change) - since I am losing my ability to test the binaries with the move to single copy. What will be challenging is that some platforms can't easily move off. Some platforms depend on a dxe only copy currently satisfied by dual copy.

Can you update the integration instructions to explain when each approach should be used?

PR #1878, opened 2026-08-06
[dev/PQC] Add Image Secure Boot Verification Result Table

Javagedes:
## Description

Adds the Image Secure Boot Verification Result Table definition and a helper library class for creating and parsing the table.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

CI

## Integration Instructions

N/A

### DIFF
--- a/SecurityPkg/Include/Guid/ImageSecureBootVerificationResultTable.h
+++ b/SecurityPkg/Include/Guid/ImageSecureBootVerificationResultTable.h
@@ -1,0 +1,183 @@
+/** @file
+  Image Secure Boot Verification Result Table (IVRT) layout and structure definitions.
+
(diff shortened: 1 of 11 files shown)


Flickdm (approved):
Plan is to perform a broader review later -- and that this is just so that we have a place to put things as we experiment

Javagedes:
Ultimately, I plan on each commit to be self-contained and reviewable, so I plan on pushing as a PR to ensure CI continues to pass as we add the necessary functionality.

jyao1:
Should this be called in ImageVerificationLib? As such, we can see if the API is good enough from consumer side.

PR #1893, opened 2026-08-26
[Draft] Add RMEM ACPI reporting proposal

chapsiru:
## Description

This draft PR adds a proposed Reserved-Memory Reporting (RMEM) ACPI interface for discussion.

Firmware may reserve physical memory for security services, firmware runtime use, shared communication buffers, graphics, NPUs, crash handling, and other platform functions. Operating systems generally expose only the aggregate hardware-reserved amount or efi descriptor, which makes it difficult to identify the purpose and owner of individual reservations.

This change introduces:

- A Revision 1 RMEM ACPI table containing the base address, size, category, and diagnostic label for each reserved-memory range.
- A versioned GUID HOB for reservations discovered before DXE.
- A DXE registration protocol for dynamic or late-discovered reservations.
- A common DXE publisher that:
  - Imports RMEM HOB records.
  - Accepts protocol registrations until `ReadyToBoot`.
  - Validates ranges, categories, labels, duplicates, and overlaps.
  - Constructs and checksums one RMEM ACPI table.
  - Installs the table through `EFI_ACPI_TABLE_PROTOCOL`.
- Package declarations for the RMEM HOB and registration protocol GUIDs.

This is a discussion proposal and is not ready to merge. Feedback is requested on:

- The ACPI table layout, category taxonomy, and revision policy.
- ACPI signature allocation and the standardization path.
- Package ownership while RMEM is not yet an approved ACPI standard.
- Whether pre-DXE HOB reporting, DXE protocol reporting, or both should be supported.
- The testing and platform-validation requirements.
- etc.

- [x] Impacts functionality?
- [x] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [x] Includes documentation?

The change adds new opt-in functionality and does not modify an existing interface. It has security and privacy implications because it exposes reserved physical-memory addresses, sizes, categories, and labels to operating-system consumers. The table is diagnostic only and does not grant access to the reported memory.

Public interface and data-format documentation is included in the new headers. 

## How This Was Tested

The following Mu non-compiler CI checks passed for `MdePkg` and `MdeModulePkg`:

- GUID uniqueness
- Package dependency validation
- DSC completeness
- License validation
- Character encoding and line-ending checks
- Uncrustify and other static formatting checks
- Editor diagnostics
- Patch whitespace validation

The checks were run with:

```powershell
stuart_ci_build.exe `
  -c CISettings.py `
  -p MdeModulePkg,MdePkg `
  -a X64 `
  -t NOOPT `
  CompilerPlugin=skip `
  HostUnitTestCompilerPlugin=skip

### CODE REVIEW

File: MdePkg/Include/Guid/ReservedMemoryReportingHob.h [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef RESERVED_MEMORY_REPORTING_HOB_H_
+#define RESERVED_MEMORY_REPORTING_HOB_H_
+
+#include <IndustryStandard/ReservedMemoryReportingTable.h>
+
+#define EDKII_RMEM_RECORD_HOB_GUID \
```
spbrogan, line 13: In general, if the code goes into MdePkg it is expected to be industry specification defined so it would be something like EFI_RMEM_RECORD_HOB_GUID.   This would indicate that this hob was defined (most likely) in the PI Specification (owned by UEFI forum).  

So if we don't want to do that then this file would best fit into MdeModulePkg and this would be an Tianocore/Edk2 implementation feature.

MdePkg/Include/Protocol/ReservedMemoryReporting.h
```diff
@@ -0,0 +1,63 @@
+/** @file
```
spbrogan, line 1: some comment here about Edkii vs EFI and MdePkg vs MdeModulePkg

MdePkg/Include/Protocol/ReservedMemoryReporting.h
```diff
@@ -0,0 +1,63 @@
...
+
+  The protocol copies the label before returning. The caller retains ownership
+  of the label buffer.
+
+  @param[in] This      A pointer to the EDKII_RMEM_REGISTRATION_PROTOCOL
+                       instance.
+  @param[in] Base      The physical address of the first byte in the range.
+  @param[in] Size      The size of the range in bytes.
+  @param[in] Category  The purpose category assigned to the range.
+  @param[in] Label     An optional null-terminated ASCII diagnostic label.
```
spbrogan, line 32: should probably call out the define that dictates the allowed max size and be clear if NULL is included in size or not.

MdePkg/MdePkg.dec
```diff
@@ -1163,6 +1166,9 @@
   gEfiPeiMmCommunication3PpiGuid = { 0xe70febf6, 0x13ef, 0x4a21, { 0x89, 0x9e, 0xb2, 0x36, 0xf8, 0x25, 0xff, 0xc9 }}
 
 [Protocols]
+  ## Include/Protocol/ReservedMemoryReporting.h
+  gEdkiiRmemRegistrationProtocolGuid = { 0x0cb661d3, 0x3c81, 0x4074, { 0xa6, 0xa8, 0xc9, 0x75, 0x9c, 0xc5, 0x12, 0x07 }}
+
   ## Include/Protocol/MemoryAccept.h
   gEdkiiMemoryAcceptProtocolGuid = { 0x38c74800, 0x5590, 0x4db4, { 0xa0, 0xf3, 0x67, 0x5d, 0x9b, 0x8e, 0x80, 0x26 }}
```
spbrogan, line 1173: Just FYI - gEdkiiMemoryAcceptProtocolGuid seems like an outlier and did not follow the rules.  I don't know exactly why it is in defined in this package.

spbrogan (commented):
this looks great.  
I added a few comments

MdeModulePkg/Universal/Acpi/RmemAcpiDxe/ReadMe.md
eeshanl, line 1: can we also add a section for how to view the acpi table entries from the os? example python/powershell commands.

MdeModulePkg/Universal/Acpi/RmemAcpiDxe/RmemAcpiDxe.c
```diff
@@ -0,0 +1,342 @@
...
+  IN RMEM_CATEGORY     Category,
+  IN CONST CHAR8       *Label
+  )
+{
+  return (Existing->Base == Base) &&
+         (Existing->Size == Size) &&
+         (Existing->Category == (UINT32)Category) &&
+         (AsciiStrCmp (Existing->Label, Label) == 0);
+}
+
```
eeshanl, line 71: nit: add comments for these functions, ie describe how we reject overlapping regions, etc.

eeshanl (commented):
overall looks good, mainly just need to consolidate everything into one Pkg + some small nits.

MdePkg/Include/Guid/ReservedMemoryReportingHob.h
chapsiru: Hello, @spbrogan

Thanks for the comment. I’m currently gathering feedback from each SiPartners for this change, and I’ll update it before merging.

For ACPI Table standardization, we should get the sign-off from them right?

Once this change is merged, how can we get their confirmation for the upstream?

eeshanl:
Please also change the target merge branch to release/202608

eeshanl:
Also, since we plan to upstream this to edk2, please add #MU_CHANGE start/end tags to any changes in existing files.

MdeModulePkg/Include/Guid/ReservedMemoryReportingHob.h
```diff
@@ -0,0 +1,40 @@
...
+  Copyright (c) Microsoft Corporation.
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#ifndef RESERVED_MEMORY_REPORTING_HOB_H_
+#define RESERVED_MEMORY_REPORTING_HOB_H_
+
+#include <IndustryStandard/ReservedMemoryReportingTable.h>
+
+#define EDKII_RMEM_RECORD_HOB_GUID \
```
chapsiru: Codes are moved to MdeModulePkg, thanks

MdePkg/MdePkg.dec
chapsiru: I addressed it by moving all proposed RMEM interfaces out of MdePkg and into MdeModulePkg

MdeModulePkg/Universal/Acpi/RmemAcpiDxe/ReadMe.md
chapsiru: I've added the powershell script for the example. Thanks

chapsiru:
Rebased the target branch to release/202608

RMEM related source code is moved to MdeModulePkg

Add MU_CHANGE for the upstream

PR #1913, opened 2026-09-11
[PQC] Add CmsSignerInfoNum, X509IsPublicKeySupported and unit tests

Flickdm:
## Description

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

<_Describe the test(s) that were run to verify the changes._>

## Integration Instructions


CryptoPkg/Include/Protocol/OneCrypto.h
```diff
@@ -5677,6 +5703,10 @@ typedef struct _ONE_CRYPTO_PROTOCOL {
   ONE_CRYPTO_X509_GET_TBS_CERT_HASH                  X509GetTbsCertHash;
   ONE_CRYPTO_AUTHENTICODE_VERIFY_EX                  AuthenticodeVerifyEx;
   ONE_CRYPTO_HASH_ALL_BY_GUID                        HashAllByGuid;
+  /// v2.0 CMS ---------------------------------------------------------------
```
Javagedes, line 5706: why 2.0 if you are appending?

Javagedes (approved):

CryptoPkg/Include/Protocol/OneCrypto.h
Flickdm: Figured it was time to just move it forward given all the changes we're making and before this gets merged I expect to remove things and really break things. I"m find changing it since the version in this case really doesnt matter since its all prototyping and the normal rules around the ABI aren't really true

PR #1915, opened 2026-09-14
RustPackageHelper.py: Update __set_members to avoid duplicates

antklein:
## Description

* Updates `RustPackageHelper` to deduplicate Cargo workspace members by path before constructing `RustPackage` objects.
* Adds unit tests for the `__set_members` function.
* Also replaces the direct `type(...) != str` comparison with `isinstance()` when identifying path-based workspace dependencies.

Some local crates are declared both as Cargo workspace members and as path-based workspace dependencies. The previous implementation placed `RustPackage` instances in a set, but `RustPackage` does not define value-based equality or hashing. Separate instances representing the same path were therefore retained as distinct entries.

Collecting `Path` objects in the set provides value-based deduplication and ensures each workspace package is tested once.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

Ran the Rust host-based unit test runner through `stuart_ci_build`.
Verified that each crate appears only once in the generated Cargo coverage command.
All unit tests are passing.

## Integration Instructions

N/A


makubacki (approved):
This would benefit from unit tests, but I don't think it's necessary given there are currently none in the plugin.

antklein:
> This would benefit from unit tests, but I don't think it's necessary given there are currently none in the plugin.

It was simple enough to add unit tests for __set_members. I modeled it after the DebugMacroCheckPlugin test design.

apop5 (approved):

PR #1926, opened 2026-09-17
[REBASE&FF][202511] Add Clang plugin

Javagedes:
## Description

Reverts the MU_BASECORE commit that adds the Clang PDB plugin in favor of the EDK II commit that adds CLANG plugin.

- [ ] Impacts functionality?
- [ ] Impacts security?
- [ ] Breaking change?
- [ ] Includes tests?
- [ ] Includes documentation?

## How This Was Tested

N/A

## Integration Instructions

N/A


os-d (commented):
Looks like this didn't apply cleanly, it pulled some stuff in from edk2 that we didn't have (VS2026) for instance, which is okay, but if we want that, we should cherry-pick those edk2 commits.

In general, if this is just cleanup to bring back to 2511 and not important for the branch, we can leave it as is in 2511 and just do the revert/cp in 2608 since Aaron has already done the rebase work from 2511, we don't need to clean up history here.

If you do keep this, can you do `git cherry-pick -x <hash>` to keep the edk2 hash?

apop5:
given how close we are to 202608, should we consider just dropping this for 202511?

Javagedes:
I will close this unless someone really wants it :)

